PR #70, opened 2022-10-28
Firefox: Added path to snap directories

cobyge:
This PR adds another directory to search in for Firefox files, which is the default snap directory.
Since Ubuntu 21.10, Firefox is installed as a Snap by default


cobyge:
@DissectBot agree

dissect/target/plugins/browsers/firefox.py
```diff
@@ -17,6 +17,7 @@ class FirefoxPlugin(Plugin):
         "AppData/Roaming/Mozilla/Firefox/Profiles",
         "AppData/local/Mozilla/Firefox/Profiles",
         ".mozilla/firefox",
+        "snap/firefox/common/.mozilla",
```
Schamper, line 20:
```suggestion
        "snap/firefox/common/.mozilla/firefox",
```
I believe this is the directory that actually contains the profiles.

Schamper (changes requested):

Schamper (approved):

PR #83, opened 2022-11-17
Refactor filesystem code

pyrco:
Note for reviewers: review each commit on its own, they are intended to go un-squashed on the main branch.

(DIS-1632)

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -946,15 +946,15 @@ class VirtualSymlink(FilesystemEntry):
 class VirtualFilesystem(Filesystem):
     __fstype__ = "virtual"
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
         self.root = VirtualDirectory(self, "/")
 
     @staticmethod
-    def detect(fh):
+    def detect(fh: BinaryIO) -> bool:
         raise TypeError("Detect is not allowed on VirtualFilesystem class")
 
-    def get(self, path, relentry=None):
+    def get(self, path: str, relentry: FilesystemEntry = None) -> FilesystemEntry:
         entry = relentry or self.root
         path = fsutil.normalize(path).strip("/")
         full_path = fsutil.join(entry.path, path)
@@ -998,7 +998,7 @@ class VirtualFilesystem(Filesystem):
 
         return entry
 
-    def makedirs(self, path):
+    def makedirs(self, path: str) -> VirtualDirectory:
         """Create virtual directories into the VFS from the given path."""
         path = fsutil.normalize(path).strip("/")
         directory = self.root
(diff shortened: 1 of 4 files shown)


Schamper:
Can't place a comment on it, but the commit message for "Return self on simple VirtualFile.get()" has a typo: "CirtualDirectory".

pyrco:
> Can't place a comment on it, but the commit message for "Return self on simple VirtualFile.get()" has a typo: "CirtualDirectory".

Check. Fixed it.

dissect/target/filesystem.py
```diff
@@ -948,10 +951,10 @@ def __init__(self, **kwargs):
         self.root = VirtualDirectory(self, "/")
 
     @staticmethod
-    def detect(fh):
+    def detect(fh: BinaryIO):
```
Schamper, line 954: Add type hints to `Virtual*` and return types?

dissect/target/filesystem.py
```diff
@@ -1030,44 +1038,58 @@ def map_dir(self, vfspath, realpath):
...
+                real_file_path = os.path.join(root, file_)
+                directory.add(file_, MappedFile(self, vfs_file_path, real_file_path))
 
-    def map_file(self, vfspath, realpath):
+    def map_file(self, vfspath: str, realpath: str):
         """Map a file from the host machine into the VFS."""
         vfspath = fsutil.normalize(vfspath)
-        self.map_file_entry(vfspath, MappedFile(self, vfspath, realpath))
+        if vfspath[-1] == "/":
+            raise AttributeError(f"Can not map a file onto a directory: {vfspath}")
```
Schamper, line 1054:
```suggestion
            raise AttributeError(f"Can't map a file onto a directory: {vfspath}")
```
Maybe it could be a feature that you can, but it takes the filename from the real path? Obviously that won't be possible for `map_file_fh`.

dissect/target/filesystem.py
```diff
@@ -1030,44 +1038,58 @@ def map_dir(self, vfspath, realpath):
...
+        file_path = vfspath.lstrip("/")
+        self.map_file_entry(vfspath, MappedFile(self, file_path, realpath))
 
-    def map_file_fh(self, vfspath, fh):
+    def map_file_fh(self, vfspath: str, fh: BinaryIO):
         """Map a file handle into the VFS."""
         vfspath = fsutil.normalize(vfspath)
-        self.map_file_entry(vfspath, VirtualFile(self, vfspath, fh))
+        if vfspath[-1] == "/":
+            raise AttributeError(f"Can not map a file onto a directory: {vfspath}")
```
Schamper, line 1062:
```suggestion
            raise AttributeError(f"Can't map a file onto a directory: {vfspath}")
```

tests/test_plugins_apps_remoteaccess_teamviewer.py
```diff
@@ -8,7 +8,7 @@
 def test_teamviewer_plugin_global_log(target_win_users, fs_win):
 
     teamviewer_logfile = absolute_path("data/TestTeamviewer.log")
-    target_logfile_name = "sysvol/Program Files/TeamViewer/TestTeamviewer.log"
+    target_logfile_name = "/sysvol/Program Files/TeamViewer/TestTeamviewer.log"
```
Schamper, line 11: I think I missed it but why does this matter now?

Schamper (changes requested):

tests/test_plugins_apps_remoteaccess_teamviewer.py
pyrco: Before it depended on the whether the `path` parameter given to `RootFilesystem.get()` had a leading `/` or not on whether the `path` attribute of the `RootFilesystemEntry` would have one or not. Now it is always consistently the same: always a leading `/`. Added to that, it means that the code flow in `RootFilesystem.get()` can be more similar to `VirtualFilesystem.get()`.

dissect/target/filesystem.py
pyrco: I decided against type hinting the other `Virtual*` classes as they were barely touched and would be too much scope creep.

dissect/target/filesystem.py
Schamper: Sounds good, but can you make a ticket for it then?

tests/test_plugins_apps_remoteaccess_teamviewer.py
Schamper: What happens if you don't add a leading `/` now?

pyrco: It will be added to the `RootFilesystemEntry`'s `.path` attribute no matter what.

dissect/target/filesystem.py
pyrco: Done

Schamper (approved):

PR #85, opened 2022-11-21
Add Virtuozzo Loader

Zawadidone:
Add Virtuozzo container loader

> VE_ROOT /vz/root/$VEID
> The mount point for container’s root. Must contain the literal string $VEID that will be substituted with the actual container UUID.

## Reference

* [https://docs.virtuozzo.com/virtuozzo_hybrid_server_7_command_line_reference/managing-system/configuration-files.html](https://docs.virtuozzo.com/virtuozzo_hybrid_server_7_command_line_reference/managing-system/configuration-files.html)


Zawadidone:
@Schamper 
> Hm can you give me a quick rundown of how Virtuozzo works and how this is supposed to work? Is each directory under /vz/root/<uuid> a separate container, and should thus be a separate target?

Virtuozzo can be used to run Linux containers which are mounted in the `/vz/root` folder. So the Virtouzzo server itself should be a target which, but also every underling container mounted to that path.

Yes every container should be a separate target is this possible with the current `DirLoader`?

Schamper:
Hm I think in that case, it's best to implement it as a child target plugin. In that case, any type of evidence that contains Virtuozzo containers (VMDK, E01, directories) will be able to list all containers.

You can take a look at the other child target plugins for inspiration: https://github.com/fox-it/dissect.target/tree/main/dissect/target/plugins/child

Once you implement the Virtuozzo child plugin, it should all hook automatically into the current `--child` and `--children` command line arguments.

Zawadidone:
Is it intended that the child target is the actual location in the target?

`/vz/root/669ef446-683b-11ed-9022-0242ac120001` (child) instead of `[...]/targets/linux/vz/root/669ef446-683b-11ed-9022-0242ac120001` (parent+child)

```bash
target-query -t ~/targets/linux/ --children -f hostname -q
<Target [...]/targets/linux> kali
<Target /vz/root/669ef446-683b-11ed-9022-0242ac120001> 669ef446-683b-11ed-9022-0242ac120001
<Target /vz/root/669ef446-683b-11ed-9022-0242ac120002> 669ef446-683b-11ed-9022-0242ac120002
```

Schamper:
Yes, child targets are always nested in the parent target. In your example it looks like you have a directory target, for which you could in theory have a working path to the child target. However, if you were to use a VMDK or EWF target, that obviously won't hold up anymore.

dissect/target/plugins/child/virtuozzo.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    __type__ = "virtuozzo"
+
+    PATH = "/vz/root"
+
+    def __init__(self, target):
+        super().__init__(target)
+        self.vz_containers = [p for p in self.target.fs.iterdir(self.PATH)]
+
```
Schamper, line 31:
```suggestion
```
`self.PATH` may not exist. `__init__` runs before `check_compatible`.

dissect/target/plugins/child/virtuozzo.py
```diff
@@ -0,0 +1,42 @@
...
+        if not self.target.fs.path(self.PATH).exists():
+            raise UnsupportedPluginError("No /vz/root folder found")
+
+    def list_children(self):
+        for container in self.vz_containers:
+            yield ChildTargetRecord(
+                type=self.__type__, 
+                path=str(self.target.fs.path(self.PATH).joinpath(container)), 
+                _target=self.target
+            )
```
Schamper, line 42:
```suggestion
    def list_children(self):
        for container in self.target.fs.path(self.PATH).iterdir():
            yield ChildTargetRecord(
                type=self.__type__, 
                path=str(container), 
                _target=self.target
            )
```
There's no real benefit to preloading the paths in this case, so just iterate over the subdirectories in here. Also if you directly use `fs.path(self.PATH).iterdir()` you can directly do `str()` to get the full path.

Schamper (changes requested):

dissect/target/plugins/child/virtuozzo.py
Zawadidone: 👍

dissect/target/plugins/child/virtuozzo.py
Zawadidone: 👍

Zawadidone:
Thanks for the suggestions.

dissect/target/plugins/child/virtuozzo.py
```diff
@@ -0,0 +1,38 @@
...
+      etc/
+      var/
+      vz/
+          root/
+              <container-uuid>/
+              <container-uuid>/
+
+    Sources:
+        - https://docs.virtuozzo.com/virtuozzo_hybrid_server_7_command_line_reference/managing-system/configuration-files.html
+    """
```
Schamper, line 22:
```suggestion
    """  # noqa: E501
```
This will fix the linting error.

Schamper (changes requested):

Schamper (approved):

PR #90, opened 2022-11-28
Adhere to the correct path separators

pyrco:
Path normalization is now done using the correct (alternative) path separators according to the path's source (filesystem).

TargetPath's flavours are now also set when not constructed through _from_parts().

(DIS-1639)

### DIFF
--- a/dissect/target/filesystems/dir.py
+++ b/dissect/target/filesystems/dir.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 from dissect.target.exceptions import FileNotFoundError, FilesystemError, IsADirectoryError, NotADirectoryError
 from dissect.target.filesystem import Filesystem, FilesystemEntry
@@ -8,9 +9,9 @@ from dissect.target.helpers import fsutil
 class DirectoryFilesystem(Filesystem):
     __fstype__ = "dir"
 
-    def __init__(self, path, case_sensitive=True):
+    def __init__(self, path: Path, alt_separator: str = "", case_sensitive: bool = True) -> None:
         self.base_path = path
-        super().__init__(case_sensitive=case_sensitive)
+        super().__init__(alt_separator=alt_separator, case_sensitive=case_sensitive)
 
     @staticmethod
     def detect(fh):
@@ -54,7 +55,8 @@ class DirectoryFilesystemEntry(FilesystemEntry):
         return self
 
     def get(self, path):
-        return self.fs.get(fsutil.join(self.path, path))
+        path = fsutil.join(self.path, path, alt_separator=self.fs.alt_separator)
+        return self.fs.get(path)
 
     def open(self):
         if self.is_dir():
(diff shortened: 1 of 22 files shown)


dissect/target/filesystems/dir.py
```diff
@@ -54,7 +54,8 @@ def _resolve(self):
         return self
 
     def get(self, path):
-        return self.fs.get(fsutil.join(self.path, path))
+        path = fsutil.join(self.path, path, alt_separator=self.fs.alt_separator)
```
Schamper, line 57: Should the `DirLoader` not set the `alt_separator` of the `DirectoryFilesystem` to `\\` if it's Windows?

dissect/target/helpers/fsutil.py
```diff
@@ -903,29 +929,31 @@ def has_glob_magic(s):
 
 
 def resolve_link(
-    fs: filesystem.Filesystem, entry: filesystem.FilesystemEntry, previous_links: Set[str] = None
-) -> filesystem.FilesystemEntry:
+    fs: filesystem.Filesystem, entry: filesystem.filesystem.FilesystemEntry, previous_links: Set[str] = None
+) -> filesystem.filesystem.FilesystemEntry:
```
Schamper, line 933: This seems like a find/replace error?

tests/test_helpers_fsutil.py
```diff
@@ -1,10 +1,175 @@
 import os
+from unittest.mock import Mock
 
 import pytest
 
+from dissect.target.filesystem import VirtualFilesystem
 from dissect.target.helpers import fsutil
 
 
+@pytest.mark.parametrize(
```
Schamper, line 10: Would it be possible to create a unit test where the target is a regular Windows target, but it has a "Unix-y" filesystem? E.g. an EXTFS partition.

dissect/target/filesystems/dir.py
pyrco: It would make it less surprising for users that they couldn't use `\`es on a filesystem that looks Windows-y. Functionally it wouldn't matter as `DirFilesystem` is an abstraction on top of the actual filesystem.

dissect/target/helpers/fsutil.py
pyrco: Yes indeed

tests/test_helpers_fsutil.py
pyrco: What would that accomplish wrt extra tests done?

At the moment support for this is not great, you won't be able to get to files which are valid on Unix systems, differ only in case or have a `\` in their name, but not on Windows, as there is no extfs-on-windows shim which translates these names.

Note that case sensitivity and alt path separators for all filesystems in a target are forced through its `RootFilesystem`. These properties are set by `OSPlugin.create()` which is called either by a loader or by `Target`'s `_init_os()`

Schamper: Mostly a yes/no if this will work or not, but if I understand you correctly, this won't work right now?

pyrco: It will work but with limitations and unknown failure modes.

Schamper: Can you create a ticket to see if this is possible to properly support?

pyrco: Done

Schamper (approved):

PR #94, opened 2022-11-29
Skip emulation drives in local loader

cecinestpasunepipe:
(DIS-1655)


dissect/target/loaders/local.py
```diff
@@ -207,6 +207,12 @@ def _get_windows_drive_volumes(log):
 
             if disk_num not in disk_map:
                 disk_path = f"\\\\.\\PhysicalDrive{disk_num}"
+
+                # Check if drive can be accessed (skip emulated drives like RAM disks)
+                if not _is_physical_drive(disk_path):
+                    log.debug(f"Skipped drive {disk_num}, not a physical drive (could be emulation or ram disk)")
```
Schamper, line 213:
```suggestion
                    log.debug("Skipped drive %d from %s, not a physical drive (could be emulation or ram disk)", disk_num, drive_path)
```

Schamper (changes requested):

PR #102, opened 2022-12-12
Make the driveletter case insensitive using the TarLoader

Miauwkeru:
(DIS-1705)﻿

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -90,7 +90,6 @@ def target_win(hive_hklm, fs_win):
     mock_target.registry.map_hive("HKEY_LOCAL_MACHINE", hive_hklm)
 
     mock_target.filesystems.add(fs_win)
-    mock_target.fs.mount("c:", fs_win)
     mock_target.apply()
 
     yield mock_target


tests/conftest.py
```diff
@@ -90,7 +90,7 @@ def target_win(hive_hklm, fs_win):
     mock_target.registry.map_hive("HKEY_LOCAL_MACHINE", hive_hklm)
 
     mock_target.filesystems.add(fs_win)
-    mock_target.fs.mount("C:", fs_win)
+    mock_target.fs.mount("c:", fs_win)
```
Schamper, line 93: Why is this change necessary?

tests/test_tools_shell.py
```diff
@@ -11,4 +11,4 @@ def test_target_cli_ls(target_win, capsys, monkeypatch):
     cli.onecmd("ls")
 
     captured = capsys.readouterr()
-    assert captured.out == "\n".join(["C:", "sysvol"]) + "\n"
+    assert captured.out == "\n".join(["c:", "sysvol"]) + "\n"
```
Schamper, line 14: Why is this change necessary?

tests/conftest.py
Schamper: Why is that? That sounds like a more severe bug and this is a band-aid to fix that?

Schamper: Isn't removing this line then the correct fix? The `WindowsPlugin` will mount the filesystem to the correct letter.

Schamper (approved):

PR #110, opened 2022-12-28
Improve unix command history plugin

JSCU-CNI:
This patch makes the bashhistory unix function shell-agnostic.


pyrco:
Hi, thanks for the nice extension of the commandhistory plugin!

Before it can be merged in however, some changes need to be made to the `RecordDescriptorExtention` used for the `_user` parameter of the `CommandHistoryRecord`.

The design/idea behind this extension field is that it is uniform across different OSes and lightweight on information to prevent to much data duplication in the output. As you can imagine the latter is of course a balance, hence the inclusion of e.g. `user_home`. Any other information can be enriched by correlating these records with the output of the users plugin. So this means using a `UnixUserRecordDescriptorExtension` for the `_user` parameter is unfortunately not what should be done.

We discussed internally and agreed that adding a `user_group` field to the `UserRecordDescriptorExtension` class in `helpers/descripter_extensions.py` which contains the id of the main group of the user should not extend the `UserRecordDescriptorExtension`  too much. So please implement it there. For now it is ok to leave the value for `user_group` as `None` on windows based systems.

dissect/target/helpers/descriptor_extensions.py
```diff
@@ -81,3 +81,25 @@ def _fill_default_fields(self, record_kwargs):
         # Reserved keywords are never part of the args of a Record's __init__.
         record_kwargs["_source"] = source
         return record_kwargs
+
+
+class UnixUserRecordDescriptorExtension(RecordDescriptorExtensionBase):
```
pyrco, line 86: See earlier comment

pyrco (changes requested):

JSCU-CNI:
Thank you for your comment @pyrco. I added `user_group` to the generic `UserRecordDescriptorExtension`.

Could you clarify why `WindowsUserRecord` and `UnixUserRecord` exist in `dissect/target/helpers/record.py`? It seems to me that per your rhetoric they should be merged into a single `UserRecordDescriptorExtension`.

dissect/target/plugins/os/unix/commandhistory.py
```diff
@@ -5,34 +5,54 @@
...
+                    continue
+
             for file_ in user_details.home_path.iterdir():
-                if not file_.name.startswith(".bash_history"):
+                if file_.name not in COMMAND_HISTORY_FILES:
                     continue
 
                 try:
-                    for line in file_.open("rt"):
+                    for line in file_.open("rt", errors="ignore"):  # Ignore Non-ASCII characters in bash_history
```
pyrco, line 55:
```suggestion
                    for line in file_.open("rt", errors="ignore"):  # Ignore Non-UTF-8 characters in bash_history
```

Wouldn't it be better to use `errors="replace"` so you know you are potentially missing something?

dissect/target/plugins/os/unix/commandhistory.py
```diff
@@ -5,34 +5,54 @@
...
+        return False
 
-    @export(record=BashHistoryRecord)
+    @export(record=CommandHistoryRecord)
     def bashhistory(self):
-        """Return bash history for all users.
+        """
+        Deprecated, use commandhistory function.
+        """
+        return self.commandhistory()
```
pyrco, line 34:
```suggestion
        warnings.warn("bashhistory is deprecated, use the commandhistory function", DeprecationWarning)
        return self.commandhistory()
```

pyrco (changes requested):

JSCU-CNI:
We added proper support for timestamps in bash history files per the `HISTTIMEFORMAT` setting (https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-HISTTIMEFORMAT). Support for Zsh timestamps is not yet implemented.

pyrco:
I saw it, nice! It is much cleaner this way.

pyrco (approved):

pyrco (changes requested):
_Could you rebase this patch set on the current main branch? There appear to be some conflicting changes. Once rebased, I'll merge it in._

Nevermind, I, merged main cleanly into the branch.

pyrco (approved):

Issue #113, opened 2022-12-31
Velociraptor Loader


Zawadidone:
The Velociraptor loader should load a Velociraptor Collection, without specifying the root filesystem of the OS as a target.

```bash
ls targets/Collection-DESKTOP-I2PURD1_localdomain-2022-12-25T21_41_18_01_00/uploads/mft/%5C%5C.%5CC%3A/
$Boot           $Extend         $LogFile        $MFT            $Recycle.Bin    ProgramData     Users           Windows
```

Specifying the path `uploads/mft/%5C%5C.%5CC%3A`.

```bash
target-query -t targets/Collection-DESKTOP-I2PURD1_localdomain-2022-12-25T21_41_18_01_00/uploads/mft/%5C%5C.%5CC%3A/ -q -f hostname
<Target [...]/targets/Collection-DESKTOP-I2PURD1_localdomain-2022-12-25T21_41_18_01_00/uploads/mft/%5C%5C.%5CC%3A> DESKTOP-I2PURD1
```

Instead the Velociraptor loader should do this with the collection directory as target.

```bash
target-query -t targets/* -q -f hostname
<Target [...]/targets/Collection-DESKTOP-I2PURD1_localdomain-2022-12-25T21_41_18_01_00> DESKTOP-I2PURD1
```

Zawadidone:
Fixed in #114

PR #114, opened 2022-12-31
Add Velociraptor loader

Zawadidone:
Ref #113

### DIFF
--- a/dissect/target/loaders/velociraptor.py
+++ b/dissect/target/loaders/velociraptor.py
@@ -20,10 +20,10 @@ def find_os_directory(path: Path):
     if os_type in ["linux", "osx"]:
         return dirs[0]
 
-    # Look for first folder in 'uploads' e.g. 'uploads/mft/%5C%5C.%5CC%3A', 'uploads/lazy_ntfs/%5C%5C.%5CC%3A'
+    # This suppports usage of the ntfs accessor 'uploads/mft/%5C%5C.%5CC%3A' not the accessors lazy_ntfs or auto
     if not os_type:
-        accessor_path = [p for p in path.joinpath(ROOT_FILESYSTEM).iterdir()][0]
-        windows_path = [p for p in accessor_path.iterdir()][0]
+        # Filter out files that start with '.'
+        windows_path = [p for p in path.joinpath(ROOT_FILESYSTEM + "/mft").iterdir() if not p.name.startswith(".")][0]
         os_type, dirs = find_dirs(windows_path)
         if os_type == "windows":
             return dirs[0]
--- a/dissect/target/loaders/velociraptor.py
+++ b/dissect/target/loaders/velociraptor.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from dissect.target.loaders.dir import DirLoader, find_dirs, map_dirs
+
+if TYPE_CHECKING:
+    from dissect.target import Target
+
+ROOT_FILESYSTEM = "uploads"
+
+
+def find_os_directory(path: Path):
+    # As of Velociraptor version 0.6.7 the structure of the Velociraptor Offline Collector varies by operating system
+    # Generic.Collectors.File (Unix, OS-X) root filesystem is 'uploads/'
+    # Generic.Collectors.File (Windows) and Windows.KapeFiles.Targets (Windows) root filesystem is
+    # 'uploads/<file-accessor>/<drive-name>/'
+    os_type, dirs = find_dirs(path.joinpath(ROOT_FILESYSTEM))
+    if os_type in ["linux", "osx"]:
+        return dirs[0]
+
+    # Look for first folder in 'uploads' e.g. 'uploads/mft/%5C%5C.%5CC%3A', 'uploads/lazy_ntfs/%5C%5C.%5CC%3A'
+    if not os_type:
+        accessor_path = [p for p in path.joinpath(ROOT_FILESYSTEM).iterdir()][0]
+        windows_path = [p for p in accessor_path.iterdir()][0]
+        os_type, dirs = find_dirs(windows_path)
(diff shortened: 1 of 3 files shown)


Schamper:
What if there are multiple drives? I think it would be beneficial if all of the drives are added.

Zawadidone:
Currently Velociraptor only supports one drive.
https://github.com/Velocidex/velociraptor/blob/master/artifacts/definitions/Windows/KapeFiles/Targets.yaml#L32

Not related to this but Velociraptor has support for different file accessors, but my prefered way is using `ntfs`. The other accessors need a different implementation in the loader that should support all three accessors (`ntfs`, `lazy_ntfs` and `auto`. But that is not part of this issue #115.

https://github.com/Velocidex/velociraptor/blob/master/artifacts/definitions/Windows/KapeFiles/Targets.yaml#L2234

Schamper:
Is it expected that Velociraptor will at some point support multiple disks, and that it will work similarly to this? Otherwise it might be good to add support for that ahead of time?

dissect/target/loaders/velociraptor.py
```diff
@@ -0,0 +1,52 @@
...
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from dissect.target.loaders.dir import DirLoader, find_dirs, map_dirs
+
+if TYPE_CHECKING:
+    from dissect.target import Target
+
+ROOT_FILESYSTEM = "uploads"
```
Schamper, line 11:
```suggestion
FILESYSTEMS_ROOT = "uploads"
```
Seems more fitting?

dissect/target/loaders/velociraptor.py
```diff
@@ -0,0 +1,52 @@
...
+    # Generic.Collectors.File (Windows) and Windows.KapeFiles.Targets (Windows) root filesystem is
+    # 'uploads/<file-accessor>/<drive-name>/'
+    os_type, dirs = find_dirs(path.joinpath(ROOT_FILESYSTEM))
+    if os_type in ["linux", "osx"]:
+        return dirs[0]
+
+    # This suppports usage of the ntfs accessor 'uploads/mft/%5C%5C.%5CC%3A' not the accessors lazy_ntfs or auto
+    if not os_type:
+        # Filter out files that start with '.'
+        windows_path = [p for p in path.joinpath(ROOT_FILESYSTEM + "/mft").iterdir() if not p.name.startswith(".")][0]
```
Schamper, line 26:
```suggestion
        windows_path = [p for p in path.joinpath(ROOT_FILESYSTEM).joinpath("/mft").iterdir() if not p.name.startswith(".")][0]
```
Better to individually join the `mft` path.

Schamper: Also, what if the `mft` directory doesn't exist?

dissect/target/loaders/velociraptor.py
```diff
@@ -0,0 +1,52 @@
...
+        os_type, dirs = find_dirs(windows_path)
+        if os_type == "windows":
+            return dirs[0]
+    return False
```
Schamper, line 30:
```suggestion
    return None
```

dissect/target/loaders/velociraptor.py
```diff
@@ -0,0 +1,52 @@
...
+
+from dissect.target.loaders.dir import DirLoader, find_dirs, map_dirs
+
+if TYPE_CHECKING:
+    from dissect.target import Target
+
+ROOT_FILESYSTEM = "uploads"
+
+
+def find_os_directory(path: Path):
```
Schamper, line 14:
```suggestion
def find_os_directory(path: Path) -> Optional[Path]:
``

**Review by Schamper — CHANGES_REQUESTED** [review]

**Comment by Zawadidone** [comment]
> Is it expected that Velociraptor will at some point support multiple disks, and that it will work similarly to this? Otherwise it might be good to add support for that ahead of time?

I don't know I cannot find any issue mentioning this besides https://github.com/Velocidex/velociraptor/issues/1576 which is unrelated. Nonetheless using Velociraptor VQL this could be supported but I have not seen usage of that.

File: dissect/target/loaders/velociraptor.py (hunk 2 above)
Comment (Zawadidone) [reply dissect/target/loaders/velociraptor.py:27]: The function should return `None`

File: dissect/target/loaders/velociraptor.py [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+    # This suppports usage of the ntfs accessor 'uploads/mft/%5C%5C.%5CC%3A' not the accessors lazy_ntfs or auto
+    if not os_type and path.joinpath(FILESYSTEMS_ROOT).joinpath("mft").exists():
+        # Filter out files that start with '.'
+        windows_path = [
+            p for p in path.joinpath(FILESYSTEMS_ROOT).joinpath("mft").iterdir() if not p.name.startswith(".")
+        ][0]
+        os_type, dirs = find_dirs(windows_path)
+        if os_type == "windows":
+            return dirs[0]
+    return None
```
Schamper, line 32:
```suggestion
    fs_root = path.joinpath(FILESYSTEMS_ROOT)
    os_type, dirs = find_dirs(fs_root)
    if os_type in ["linux", "osx"]:
        return dirs[0]

    # This suppports usage of the ntfs accessor 'uploads/mft/%5C%5C.%5CC%3A' not the accessors lazy_ntfs or auto
    mft_root = fs_root.joinpath("mft")
    if not os_type and mft_root.exists():
        # Filter out files that start with '.'
        windows_path = [p for p in mft_root.iterdir() if not p.name.startswith(".")][0]
        os_type, dirs = find_dirs(windows_path)
        if os_type == "windows":
            return dirs[0]

    return None
```
Re-use some path variables to make it a bit easier to read.

Schamper (changes requested):

Schamper (approved):

Zawadidone:
Thanks!

PR #116, opened 2023-01-03
Add locale plugin for unix and windows

JSCU-CNI:
This plugin aims to extract certain locale information from both unix and windows systems.


dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,127 @@
...
+                yield UnixKeyboardRecord(
+                    layout=k["LAYOUT"],
+                    model=k["MODEL"],
+                    variant=k["VARIANT"],
+                    options=k["OPTIONS"],
+                    backspace=k["BACKSPACE"],
+                    _target=self.target,
+                )
+        except FileNotFoundError:
+            pass
```
Poeloe, line 122: I would suggest something like this:

```suggestion
        k = {}
        paths = ["/etc/default/keyboard", "/etc/vconsole.conf"] + list(
            self.target.fs.glob("/etc/X11/xorg.conf.d/*-keyboard.conf")
        )

        for path_ in paths:
            if (path := self.target.fs.path(path_)).exists():
                for line in path.open("rt"):
                    if len(key_value := line.split("=")) == 2:
                        k[key_value[0].replace("XKB", "").replace("KEYMAP", "LAYOUT")] = (
                            key_value[1].replace('"', "").strip()
                        )

                yield UnixKeyboardRecord(
                    layout=k.get("LAYOUT"),
                    model=k.get("MODEL"),
                    variant=k.get("VARIANT"),
                    options=k.get("OPTIONS"),
                    backspace=k.get("BACKSPACE"),
                    _target=self.target,
                )
                # note that you might still want to exit if LAYOUT is not None. Add if appropriate 
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,127 @@
...
+        Get the timezone of the system.
+        """
+
+        # /etc/timezone should contain a simple timezone string
+        # on most unix systems
+        try:
+            for line in self.target.fs.path("/etc/timezone").open("rt"):
+                return line.strip()
+        except FileNotFoundError:
+            pass
```
Poeloe, line 25:
```suggestion
        if (path := self.target.fs.path("/etc/timezone")).exists():
            for line in path.open("rt"):
                return line.strip()
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # /etc/localtime should be a symlink to
+        # eg. /usr/share/zoneinfo/America/New_York
+        # on centos and some other distros
+        zoneinfo = self.target.fs.path("/etc/localtime")
+
+        if zoneinfo.exists():
```
Poeloe, line 32:
```suggestion
        if (zoneinfo := self.target.fs.path("/etc/localtime")).exists():
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,127 @@
...
+        locale_paths = ["/etc/default/locale", "/etc/locale.conf"]
+
+        for locale_path in locale_paths:
+            try:
+                for line in self.target.fs.path(locale_path).open("rt"):
+                    if "LANG=" in line:
+                        return line.replace("LANG=", "").replace('"', "").strip()
+                return
+            except FileNotFoundError:
+                pass
```
Poeloe, line 53:
```suggestion
        for locale_path in locale_paths:
            if (path := self.target.fs.path(local_path)).exists():
                for line in path.open("rt"):
                    if "LANG=" in line:
                        return line.replace("LANG=", "").replace('"', "").strip()
                return
```

Poeloe (changes requested):
Thanks for your contribution!

In general, I would like to remark that since these records are specific for each OS, it would be great to place them in `_os.py` for each OS. So the Unix part can go to `dissect/target/plugins/os/unix/_os.py` and the Windows part can go to `dissect/target/plugins/os/windows/_os.py`.

My code suggestions in this review are for now focused on the Unix part. I will have another look at the Windows part soon.

Thanks again for your contribution, it is greatly appreciated

dissect/target/helpers/record.py
```diff
@@ -136,6 +136,25 @@ def DynamicDescriptor(types):  # noqa
     ],
 )
 
+WindowsKeyboardRecord = TargetRecordDescriptor(
```
Poeloe, line 139: I suggest to keep these records local (so in the file that uses them), instead of in this general location

Poeloe (changes requested):

Schamper:
> It seems to me that the `_os.py` files are a bit large and placing these functions in separate `locale.py` files is the cleaner approach. Note that the files are already split: `dissect/target/plugins/os/unix/locale.py` and `dissect/target/plugins/os/windows/locale.py`.

I agree, the location is good. The only remark I might add is the location of the test file. Perhaps better to just place that in `tests/data` or change it all up to be something like `tests/data/unix/logs` or something.

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,91 @@
...
+        ("string", "options"),
+        ("string", "backspace"),
+    ],
+)
+
+
+class LocalePlugin(Plugin):
+    def __init__(self, target):
+        super().__init__(target)
+
```
Schamper, line 19:
```suggestion
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,91 @@
...
+    def check_compatible(self):
+        pass
+
+    @export(property=True)
+    def timezone(self):
+        """
+        Get the timezone of the system.
+        """
```
Schamper, line 27:
```suggestion
        """Get the timezone of the system."""
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,91 @@
...
+            timezone = "/".join(zoneinfo_path[-2:])
+            return timezone
+
+        return
+
+    @export(property=True)
+    def language(self):
+        """
+        Get the configured locale(s) of the system.
+        """
```
Schamper, line 49:
```suggestion
        """Get the configured locale(s) of the system."""
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,91 @@
...
+                for line in path.open("rt"):
+                    if "LANG=" in line:
+                        return line.replace("LANG=", "").replace('"', "").strip()
+                return
+
+    @export(record=UnixKeyboardRecord)
+    def keyboard(self):
+        """
+        Get the keyboard layout(s) of the system.
+        """
```
Schamper, line 63:
```suggestion
        """Get the keyboard layout(s) of the system."""
```

dissect/target/plugins/os/windows/locale.py
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+    def check_compatible(self):
+        pass
+
+    @export(record=WindowsKeyboardRecord)
+    def keyboard(self):
+        """
+        Yield records of installed keyboards on the system.
+        """
```
Schamper, line 28:
```suggestion
        """Yield records of installed keyboards on the system."""
```

dissect/target/plugins/os/windows/locale.py
```diff
@@ -0,0 +1,61 @@
...
+                        layout=self.LANG_DICT.get(language_id, language_id),
+                        id=language_id,
+                        _target=self.target,
+                    )
+
+    @export(property=True)
+    def language(self):
+        """
+        Get a list of installed languages on the system.
+        """
```
Schamper, line 45:
```suggestion
        """Get a list of installed languages on the system."""
```

dissect/target/plugins/os/windows/locale.py
```diff
@@ -0,0 +1,61 @@
...
+                language = subkey.name
+                if language not in found_languages:
+                    found_languages.append(language)
+        return found_languages
+
+    @export(property=True)
+    def timezone(self):
+        """
+        Get the configured timezone of the system.
+        """
```
Schamper, line 59:
```suggestion
        """Get the configured timezone of the system."""
```

tests/test_plugins_general_locale.py
```diff
@@ -0,0 +1,74 @@
...
+from dissect.target.plugins.os.unix.locale import LocalePlugin as UnixLocalePlugin
+from dissect.target.plugins.os.windows.locale import LocalePlugin as WindowsLocalePlugin
+
+from ._utils import absolute_path
+
+
+def test_locale_plugin_windows(target_win_users, hive_hku, hive_hklm):
+    """
+    Hive structure originates from a Windows Server 2019 installation.
+    """
```
Schamper, line 14:
```suggestion
    # Hive structure originates from a Windows Server 2019 installation.
```
This is more a code comment than a function docstring.

tests/test_plugins_general_locale.py
```diff
@@ -0,0 +1,74 @@
...
+    assert keyboard[0].layout == "nl"
+    assert keyboard[0].id == "00000413"
+    assert keyboard[1].layout == "us"
+    assert keyboard[1].id == "00000409"
+
+
+def test_locale_plugin_unix(target_unix_users, fs_unix):
+    """
+    Locale locations originate from Ubuntu 20.
+    """
```
Schamper, line 59:
```suggestion
    # Locale locations originate from Ubuntu 20.
```

dissect/target/plugins/os/unix/locale.py
```diff
@@ -0,0 +1,83 @@
...
+
+                yield UnixKeyboardRecord(
+                    layout=k.get("LAYOUT"),
+                    model=k.get("MODEL"),
+                    variant=k.get("VARIANT"),
+                    options=k.get("OPTIONS"),
+                    backspace=k.get("BACKSPACE"),
+                    _target=self.target,
+                )
+                # note that you might still want to exit if LAYOUT is not None. Add if appropriate
```
Poeloe, line 78: As this was a comment within my suggested code; is this still necessary? If so, please implement it or else remove the comment ;)

dissect/target/plugins/os/windows/locale.py
```diff
@@ -0,0 +1,56 @@
...
+    ],
+)
+
+
+class LocalePlugin(Plugin):
+    def __init__(self, target):
+        super().__init__(target)
+        self.LANG_DICT = {
+            lang.name: lang.value
+            for k in self.target.registry.key("HKLM\\SYSTEM\\CurrentControlSet\\Control\\Keyboard Layout\\DosKeybCodes")
```
Poeloe, line 18: I actually don't know if this key is always present, but what I would suggest is to first check if the key exists. Else all function within the locale plugin will not be available.

Same holds for the other keys used with respect to checking for their existence ;)

dissect/target/plugins/os/windows/locale.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+    def check_compatible(self):
+        try:
+            self.target.registry.key("HKLM\\SYSTEM\\CurrentControlSet\\Control\\Keyboard Layout\\DosKeybCodes")
+            self.target.registry.key("HKLM\\SYSTEM\\CurrentControlSet\\Control\\TimeZoneInformation")
+            assert len(self.target.registry.key("HKCU\\Keyboard Layout\\Preload").keys) > 0
+            assert len(self.target.registry.key("HKCU\\Control Panel\\International\\User Profile").keys) > 0
+        except (RegistryKeyNotFoundError, RegistryValueNotFoundError, AssertionError):
+            return False
```
Schamper, line 33: This won't actually work because `__init__` runs before `check_compatible`. Personally I'd revert it to the way it was and "assume" that these keys will always exist. We already do this in multiple places, and I don't think we have a very nice solution for it at the time. So if we ever do think of one we'll fix it everywhere at the same time.

You can however change the `check_compatible` to the following:
```suggestion
    def check_compatible(self):
        return self.target.has_function("registry")
```

Schamper (changes requested):

tests/test_plugins_general_locale.py
```diff
@@ -0,0 +1,70 @@
...
+from io import BytesIO
+
+from dissect.target.helpers.regutil import VirtualKey, VirtualValue
+from dissect.target.plugins.os.unix.locale import LocalePlugin as UnixLocalePlugin
+from dissect.target.plugins.os.windows.locale import LocalePlugin as WindowsLocalePlugin
+
+from ._utils import absolute_path
+
+
+def test_locale_plugin_windows(target_win_users, hive_hku, hive_hklm):
```
Schamper, line 11: Can this test be moved into a new file: `tests/test_plugins_os_windows_locale.py`

tests/test_plugins_general_locale.py
```diff
@@ -0,0 +1,70 @@
...
+
+    keyboard = list(target_win_users.keyboard())
+    assert len(keyboard) == 2
+    assert keyboard[0].layout == "nl"
+    assert keyboard[0].id == "00000413"
+    assert keyboard[1].layout == "us"
+    assert keyboard[1].id == "00000409"
+
+
+def test_locale_plugin_unix(target_unix_users, fs_unix):
```
Schamper, line 54: Can this test be moved into a new file: `tests/test_plugins_os_unix_locale.py`

Schamper (changes requested):

Schamper (approved):

PR #117, opened 2023-01-03
Add rudimentary support for zip files

JSCU-CNI:
This zip filesystem adds support for parsing zip files. Its structure is copied from the tar filesystem. Symlink support is missing since no symlinks are supported by cpython's `zipfile` module.


dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+class ZipFilesystem(Filesystem):
+    """This Zip filesystem implementation is WIP and
+    has only been tested for opening java jar archives.
+    Its structure has been copied from the tar filesystem
+    implementation.
+    """
+
+    __fstype__ = "zip"
+
+    def __init__(self, fh, base=None, case_sensitive=True, zipinfo=None, *args, **kwargs):
```
Schamper, line 23:
```suggestion
    def __init__(self, fh: BinaryIO, base: Optional[str]=None, alt_separator: str = "", case_sensitive: bool = True, *args, **kwargs):
```
Remove `zipinfo` (it's not used), add `alt_separator` and add types.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+        fh.seek(0)
+
+        self.zip = zipfile.ZipFile(fh, mode="r")
+        self.base = base.strip("/") if base else ""
+
+        self._fs = VirtualFilesystem(case_sensitive=case_sensitive)
```
Schamper, line 29:
```suggestion
        self._fs = VirtualFilesystem(alt_separator=alt_separator, case_sensitive=case_sensitive)
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+        for member in self.zip.infolist():
+
```
Schamper, line 32:
```suggestion
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+            rel_name = mname[len(self.base) :]
+
+            # NOTE: Normally we would check here if the member is a symlink or not,
+            # however Python does not have symlink support in the zipfile module.
+            # see https://github.com/python/cpython/issues/82102
+
+            file_entry = ZipFilesystemEntry(self, rel_name, member)
+            self._fs.map_file_entry(rel_name, file_entry)
+
+        super().__init__(case_sensitive=case_sensitive, *args, **kwargs)
```
Schamper, line 49:
```suggestion
        super().__init__(alt_separator=alt_separator, case_sensitive=case_sensitive, *args, **kwargs)
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @staticmethod
+    def detect(fh):
+        """Detect a zip file on a given file-like object."""
+        offset = fh.tell()
+        try:
+            fh.seek(0)
+            zipfile.ZipFile(fileobj=fh, mode="r")
+            return True
```
Schamper, line 58:
```suggestion
            return zipfile.is_zipfile(fh)
```
Should be a bit more efficient.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ZipFilesystem(Filesystem):
+    """This Zip filesystem implementation is WIP and
+    has only been tested for opening java jar archives.
+    Its structure has been copied from the tar filesystem
+    implementation.
+    """
```
Schamper, line 19:
```suggestion
    """Filesystem implementation for zip files.
    
    Python does not have symlink support in the zipfile module, so that's not currently supported.
    See https://github.com/python/cpython/issues/82102 for more information.
    """
```

The untested comment is unnecessary in the docstring, but it would be nice to put the symlink warning there so you don't have to repeat it everywhere.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+            mname = member.filename.strip("/")
+            if not mname.startswith(self.base):
+                continue
+
+            rel_name = mname[len(self.base) :]
+
+            # NOTE: Normally we would check here if the member is a symlink or not,
+            # however Python does not have symlink support in the zipfile module.
+            # see https://github.com/python/cpython/issues/82102
```
Schamper, line 44:
```suggestion
            # NOTE: Normally we would check here if the member is a symlink or not
```
The explanation is now in the docstring, so only leave this note here.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+    def get(self, path):
+        """Returns a ZipFilesystemEntry object corresponding to the given path."""
+        return self._fs.get(path)
+
+
+class ZipFilesystemEntry(VirtualFile):
+    def _resolve(self):
+        # NOTE: Normally we would check here if the member is a symlink or not,
+        # however Python does not have symlink support in the zipfile module.
+        # see https://github.com/python/cpython/issues/82102
```
Schamper, line 74:
```suggestion
```
Having this in multiple places is now unnecessary.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+        return self
+
+    def open(self):
+        """Returns file handle (file-like object)."""
+        try:
+            f = self.fs.zip.read(self.entry)
+            return BufferedStream(io.BytesIO(f), size=self.entry.file_size)
```
Schamper, line 81:
```suggestion
            return BufferedStream(self.fs.zip.open(self.entry), size=self.entry.file_size)
```
No need to fully read the file in memory.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+        except Exception as e:
+            print(str(e))
```
Schamper, line 83:
```suggestion
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+            raise FileNotFoundError()
+
+    def is_dir(self):
+        """Return whether this entry is a directory. Resolves symlinks when possible."""
+        try:
+            return self._resolve().entry.isdir()
```
Schamper, line 89:
```suggestion
            return self._resolve().entry.is_dir()
```
`zipfile.ZipInfo` has no `isdir`.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+        except FilesystemError:
+            return False
+
+    def is_file(self):
+        """Return whether this entry is a file. Resolves symlinks when possible."""
+        try:
+            return self._resolve().entry.isfile()
+        except FilesystemError:
+            return False
```
Schamper, line 98:
```suggestion
        try:
            return not self.is_dir()
        except FilesystemError:
            return False
```
`zipfile.ZipInfo` has no `isfile`, so just use the inverse of `is_dir`. Note that an error should still result in `False`.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def is_symlink(self):
+        """Return whether this entry is a link."""
+        # NOTE: Normally we would check here if the member is a symlink or not,
+        # however Python does not have symlink support in the zipfile module.
+        # see https://github.com/python/cpython/issues/82102
```
Schamper, line 104:
```suggestion
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+        return False
+
+    def readlink(self):
+        """Read the link if this entry is a symlink. Returns a string."""
+        return self.entry.linkname
```
Schamper, line 109:
```suggestion
        raise NotImplementedError()
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def readlink_ext(self):
+        """Read the link if this entry is a symlink. Returns a filesystem entry."""
+        # Can't use the one in VirtualFile as it overrides the FilesystemEntry
+        return fsutil.resolve_link(fs=self.fs, entry=self)
```
Schamper, line 114:
```suggestion
        raise NotImplementedError()
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+                0,
+                0,
+                self.entry.uid,
+                self.entry.gid,
+                self.entry.size,
+                0,
+                self.entry.mtime,
+                0,
+            ]
+        )
```
Schamper, line 136: This is not implemented yet, perhaps something like:
```suggestion
        # ['mode', 'addr', 'dev', 'nlink', 'uid', 'gid', 'size', 'atime', 'mtime', 'ctime']
        return fsutil.stat_result(
            [
                (stat.S_IFDIR if self.is_dir() else S_IFREG) | 0o777,
                self.entry.header_offset,
                id(self.fs),
                0,
                0,
                0,
                self.entry.file_size,
                0,
                datetime(*self.entry.date_time, tzinfo=timezone.utc),
                0,
            ]
        )
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+            # however Python does not have symlink support in the zipfile module.
+            # see https://github.com/python/cpython/issues/82102
+
+            file_entry = ZipFilesystemEntry(self, rel_name, member)
+            self._fs.map_file_entry(rel_name, file_entry)
+
+        super().__init__(case_sensitive=case_sensitive, *args, **kwargs)
+
+    @staticmethod
+    def detect(fh):
```
Schamper, line 52:
```suggestion
    def detect(fh: BinaryIO) -> bool:
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+            fh.seek(0)
+            zipfile.ZipFile(fileobj=fh, mode="r")
+            return True
+        except Exception as e:
+            log.warning("Failed to detect zip filesystem", exc_info=e)
+            return False
+        finally:
+            fh.seek(offset)
+
+    def get(self, path):
```
Schamper, line 65:
```suggestion
    def get(self, path: str) -> ZipFilesystemEntry:
```

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,136 @@
...
+    def readlink_ext(self):
+        """Read the link if this entry is a symlink. Returns a filesystem entry."""
+        # Can't use the one in VirtualFile as it overrides the FilesystemEntry
+        return fsutil.resolve_link(fs=self.fs, entry=self)
+
+    def stat(self):
+        """Return the stat information of this entry."""
+        return self._resolve().lstat()
+
+    def lstat(self):
```
Schamper, line 120:
```suggestion
    def lstat(self) -> fsutil.stat_result:
```

Schamper (changes requested):
I locally tried to see if `zipfile.Path` could be used but unfortunately it's a bit too restrictive. 

I didn't put the imports for the type hints, so you'll have to fix that.

Also please add a unit test. You can use the following as a boilerplate or as inspiration:
```python
import io
import zipfile

import pytest

from dissect.target.filesystems.zip import ZipFilesystem


def _create_zip(prefix=""):
    buf = io.BytesIO()
    zf = zipfile.ZipFile(buf, "w")

    zf.writestr(f"{prefix}file_1", "file 1 contents")
    zf.writestr(f"{prefix}file_2", "file 2 contents")
    for i in range(100):
        zf.writestr(f"{prefix}dir/{i}", f"contents {i}")

    zf.close()
    buf.seek(0)
    return buf


@pytest.fixture
def zip_simple():
    yield _create_zip()


@pytest.fixture
def zip_base():
    yield _create_zip("base/")


@pytest.fixture
def zip_relative():
    yield _create_zip("./")


@pytest.mark.parametrize(
    "obj",
    [
        "split_fhs",
        "split_paths",
        "split_symlink",
    ],
)
def test_filesystems_zip(obj, request):
    fh = request.getfixturevalue(obj)

    assert ZipFilesystem.detect(fh)
    fs = ZipFilesystem(fh)
    # rest of the test here
```

Schamper:
Can you also add a unit test using the boilerplate I gave above?

JSCU-CNI:
Added a test as you suggested @Schamper. Couldn't get relative paths to work in this test, could you take a look? It seems that something goes wrong at `VirtualFilesystem.get`.

Schamper:
I was just checking the code out to work on this :) I made some changes. The tests now pass, can you please verify if these changes also work as you expect them to?

JSCU-CNI:
> I was just checking the code out to work on this :) I made some changes. The tests now pass, can you please verify if these changes also work as you expect them to?

Thanks, this works like a charm!

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,153 @@
...
+            return False
+        finally:
+            fh.seek(offset)
+
+    def get(self, path: str) -> ZipFilesystemEntry:
+        """Returns a ZipFilesystemEntry object corresponding to the given path."""
+        return self._fs.get(path)
+
+
+class ZipFilesystemEntry(VirtualDirectory, VirtualFile):
```
pyrco, line 82: I think it would be better and more robust to have a separate `class ZipFilesystemDirectoryEntry(VirtualDirectoryEntry)` which would store the `ZipInfo` `entry` somewhere and override the `stat()` method.

It would also do away with a lot of the overridden methods in this class.

dissect/target/filesystems/zip.py
```diff
@@ -0,0 +1,153 @@
...
+        self.alt_separator = alt_separator
+
+        self._fs = VirtualFilesystem(alt_separator=alt_separator, case_sensitive=case_sensitive)
+
+        for member in self.zip.infolist():
+            mname = member.filename.strip("/")
+            if not mname.startswith(self.base):
+                continue
+
+            rel_name = fsutil.normpath(mname[len(self.base) :])
```
pyrco, line 57:
```suggestion
            rel_name = fsutil.normpath(mname[len(self.base) :], alt_separator=self.alt_separator)
```

pyrco (changes requested):

dissect/target/filesystems/zip.py
Schamper: Fair enough, I was mostly looking for feedback for specifically this decision.

Schamper:
@JSCU-CNI btw I'll take care of these comments since they were my changed.

pyrco (approved):
LGTM

I guess similar changes (esp. the split in file and dir entries) are in order to #144?

Schamper (approved):

PR #122, opened 2023-01-10
Make mft and mft_timeline plugin a bit more robust

Schamper:
(DIS-1757)


Schamper:
Not really feeling it, that'd require passing in a bunch of arguments and carrying some variables already in the current scope, not to mention an additional function call.

PR #124, opened 2023-01-11
New implementation of -l/-f/-h

cecinestpasunepipe:
Features:

- Allow wildcards/fnmatch expression with -l/-f/-h
- If the output is mixed, ignore non-records and emit warning
- If -f/-l matches the root of the plugin tree then also execute based on tree
- Only allow exported functions with -l/-f/-h
- Only execute/show relevant functions with -f/-l if a target is provided
- Show functions supported by ANY of the specified targets
- Show loaders only if no target has been provided
- List generic OS functions under OS if no target has been specified
- Make -l work with multiple functions
- Batch mode: If multiple functions are requested, don't break on intermediate errors
- Allow child plugins to appear in plugin list as well (future)
- Allow quick and exceptionless compatibility checks (is_compatible)
- Allow plugins to "hide" from wildcard (is_findable) (i.e. Browser, Webserver)
- Move Windows _os.codepage to os.windows.generic.codepage
- Fix amcache issue
- Add data class to encapsulate PluginFunction
- Remove redundant code
- Don't execute any properties/functions when inspecting/listing/checking
- Backward compatibility & Tests

Also fixes:
(https://github.com/fox-it/dissect.target/issues/149)
(DIS-1379)
(DIS-1426)
(DIS-1702)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -5,12 +5,15 @@ See dissect/target/plugins/general/example.py for an example plugin.
 from __future__ import annotations
 
 import enum
+import fnmatch
 import importlib
 import importlib.util
+import inspect
 import logging
 import os
 import sys
 import traceback
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Iterator, Optional, Type
 
@@ -198,6 +201,15 @@ class Plugin:
     """Defines the plugin namespace."""
     __record_descriptors__: list[RecordDescriptor] = None
     """Defines a list of :class:`~flow.record.RecordDescriptor` of the exported plugin functions."""
+    __findable__: bool = True
+    """Determines whether this plugin will be revealed when using search patterns.
+
+    Some (meta)-plugins are not very suitable for wild cards on CLI or
+    plugin searches, because they will produce duplicate records or results.
+    For instance a plugin that offers the same functions as subplugins will
+    produce redundant results when used with a wild card
+    (browser.* -> browser.history + browser.*.history).
+    """
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
@@ -218,6 +230,15 @@ class Plugin:
 
     def __init__(self, target: Target):
         self.target = target
+
+    def is_compatible(self) -> bool:
+        """Perform a compatibility check with the target."""
+        try:
+            self.check_compatible()
+        except Exception:
+            return False
+
+        return True
 
     def check_compatible(self) -> None:
         """Perform a compatibility check with the target.
@@ -591,7 +612,7 @@ def get_plugins_by_namespace(namespace: str, osfilter: str = None) -> Iterator[P
     """Get a plugin descriptor by namespace.
 
     Args:
-        func_name: Function name to lookup.
+        namespace: Plugin namespace to match.
         osfilter: The OS module path the plugin should be from.
     """
     for plugin_desc in plugins(osfilter):
@@ -772,27 +793,6 @@ def _modulepath(cls) -> str:
     return cls.__module__.replace(MODULE_PATH, "").lstrip(".")
 
 
-def get_plugin_classes_with_method(method_name: str) -> Iterator[Type[Plugin]]:
-    """Retrieve plugin classess that have a method that matches ``method_name``."""
-    for desc in get_plugins_by_func_name(method_name):
-        try:
-            yield load(desc)
-        except PluginError:
-            pass
-
-    if method_name in get_nonprivate_attribute_names(OSPlugin):
-        yield OSPlugin
-
-
-def get_plugin_classes_by_namespace(namespace: str) -> Iterator[Type[Plugin]]:
-    """Retrieve plugin classess that have ``__namespace__`` defined that matches ``namespace``."""
-    for desc in get_plugins_by_namespace(namespace):
-        try:
-            yield load(desc)
-        except PluginError:
-            pass
-
-
 # Needs to be at the bottom of the module because __init_subclass__ requires everything
 # in the parent class Plugin to be defined and resolved.
 class InternalPlugin(Plugin):
@@ -809,3 +809,133 @@ class InternalPlugin(Plugin):
 
         super().__init_subclass__(**kwargs)
         return cls
+
+
+@dataclass
+class PluginFunction:
+    name: str
+    output_type: str
+    class_object: str
+    method_name: str
+    plugin_desc: dict
+
+
+def plugin_function_index(target: Target) -> tuple[dict[str, Any], set[str]]:
+    """Returns an index-list for plugins.
+
+    This list is used to match CLI expressions against to find the desired plugin.
+    Also returns the roots to determine whether a CLI expression has to be compared
+    to the plugin tree or parsed using legacy rules.
+    """
+    index = {}
+    rootset = set()
+
+    def all_plugins():
+        yield from plugins()
+        yield from os_plugins()
+        yield from child_plugins()  # Doesn't export anything but added for completeness.
+
+    for available in all_plugins():
+        if "get_all_records" in available["exports"]:
+            available["exports"].remove("get_all_records")
+        modulepath = available["module"]
+        if modulepath.endswith("._os"):
+            if not target._os:
+                # if no target available mention general OS instead of specific one
+                available["module"] = "OS"
+            elif target._os.__class__.__name__ != available["class"]:
+                continue
+            rootset.add(modulepath.split(".")[0])
+            for exported in available["exports"]:
+                index[f"{modulepath}.{exported}"] = available
+                index[exported] = available
+        else:
+            rootset.add(modulepath.split(".")[0])
+            for exported in available["exports"]:
+                index[f"{modulepath}.{exported}"] = available
+
+    return index, rootset
+
+
+def find_plugin_functions(target: Target, patterns: str, compatibility: bool = False) -> list[PluginFunction]:
+    """Finds plugins that match the target and the patterns.
+
+    Given a target, a comma separated list of patterns and an optional compatibility flag,
+    this function finds matching plugins, optionally checking compatibility and returns
+    a list of plugin function descriptors (including output types).
+    """
+    result = []
+    functions, rootset = plugin_function_index(target)
+
+    for pattern in patterns.split(","):
+        wildcard = any(char in pattern for char in ["*", "!", "?", "[", "]"])
+        treematch = pattern.split(".")[0] in rootset and pattern != "os"
+
+        if treematch and not wildcard:
+            # Examples:
+            #     -f browsers -> browsers* (the whole package)
+            #     -f apps.webservers.iis -> apps.webservers.iis* (logs etc)
+            # We do not include  a dot because that does not work if the full path is given:
+            #     -f apps.webservers.iis.logs != apps.webservers.iis.logs.* (does not work)
+            pattern += "*"
+
+        if wildcard or treematch:
+            for index_name, func in functions.items():
+                if fnmatch.fnmatch(index_name, pattern):
+                    method_name = index_name.split(".")[-1]
+                    loaded_plugin_object = load(func)
+
+                    # Skip plugins that don't want to be found by wildcards
+                    if not loaded_plugin_object.__findable__:
+                        continue
+
+                    fobject = inspect.getattr_static(loaded_plugin_object, method_name)
+
+                    if compatibility:
+                        try:
+                            if not loaded_plugin_object(target).is_compatible():
+                                continue
+                        except Exception:
+                            continue
+
+                    result.append(
+                        PluginFunction(
+                            name=index_name,
+                            class_object=loaded_plugin_object,
+                            method_name=method_name,
+                            output_type=getattr(fobject, "__output__", "text"),
+                            plugin_desc=func,
+                        )
+                    )
+        else:
+            # otherwise match using ~ classic style
+            if pattern.find(".") > -1:
+                namespace, funcname = pattern.split(".", 1)
+            else:
+                funcname = pattern
+                namespace = None
+
+            plugindesc = None
+            for index_name, func in functions.items():
+                nsmatch = namespace and func["namespace"] == namespace and funcname in func["exports"]
+                fmatch = not namespace and not func["namespace"] and funcname in func["exports"]
+                if nsmatch or fmatch:
+                    plugindesc = func
+
+            if plugindesc:
+                loaded_plugin_object = load(plugindesc)
+                fobject = inspect.getattr_static(loaded_plugin_object, funcname)
+
+                if compatibility and not loaded_plugin_object(target).is_compatible():
+                    continue
+
+                result.append(
+                    PluginFunction(
+                        name=f"{plugindesc['module']}.{pattern}",
+                        class_object=loaded_plugin_object,
+                        method_name=funcname,
+                        output_type=getattr(fobject, "__output__", "text"),
+                        plugin_desc=plugindesc,
+                    )
+                )
+    return result
--- a/dissect/target/plugins/browsers/browser.py
+++ b/dissect/target/plugins/browsers/browser.py
@@ -35,6 +35,8 @@ class BrowserPlugin(Plugin):
     """
 
     __namespace__ = "browser"
+    __findable__ = False
+
     BROWSERS = [
         "chrome",
         "chromium",
--- a/dissect/target/plugins/general/plugins.py
+++ b/dissect/target/plugins/general/plugins.py
@@ -6,11 +6,14 @@ from dissect.target.helpers.docs import INDENT_STEP, get_plugin_overview
 from dissect.target.plugin import Plugin, arg, export
 
 
-def categorize_plugins() -> dict:
+def categorize_plugins(plugins_selection: list[dict] = None) -> dict:
     """Categorize plugins based on the module it's from."""
 
     output_dict = dict()
-    for plugin_dict in get_exported_plugins():
+
+    plugins_selection = plugins_selection or get_exported_plugins()
+
+    for plugin_dict in plugins_selection:
         tmp_dict = dictify_module_recursive(
             list_of_items=plugin_dict["module"].split("."),
             last_value=plugin.load(plugin_dict),
@@ -98,10 +101,10 @@ class PluginListPlugin(Plugin):
     def check_compatible(self):
         return True
 
-    @export(output="none")
+    @export(output="none", cache=False)
     @arg("--docs", dest="print_docs", action="store_true")
-    def plugins(self, print_docs=False):
-        categorized_plugins = dict(sorted(categorize_plugins().items()))
+    def plugins(self, plugins: list[dict] = None, print_docs: bool = False) -> None:
+        categorized_plugins = dict(sorted(categorize_plugins(plugins).items()))
         plugin_descriptions = output_plugin_description_recursive(categorized_plugins, print_docs)
 
         plugins_list = textwrap.indent(
--- a/dissect/target/tools/query.py
+++ b/dissect/target/tools/query.py
@@ -6,13 +6,14 @@ import logging
 import pathlib
 import sys
 from datetime import datetime
+from typing import Callable
 
 from flow.record import RecordPrinter, RecordStreamWriter, RecordWriter
 
 from dissect.target import Target
 from dissect.target.exceptions import PluginNotFoundError, UnsupportedPluginError
 from dissect.target.helpers import cache, hashutil
-from dissect.target.plugin import PLUGINS, Plugin
+from dissect.target.plugin import PLUGINS, OSPlugin, Plugin, find_plugin_functions
 from dissect.target.report import ExecutionReport
 from dissect.target.tools.utils import (
     configure_generic_arguments,
@@ -20,7 +21,6 @@ from dissect.target.tools.utils import (
     generate_argparse_for_bound_method,
     generate_argparse_for_plugin_class,
     generate_argparse_for_unbound_method,
-    get_attr_for_attr_path,
     persist_execution_report,
     process_generic_arguments,
 )
@@ -57,7 +57,15 @@ def main():
     parser.add_argument("-f", "--function", help="function to execute")
     parser.add_argument("--child", help="load a specific child path or index")
     parser.add_argument("--children", action="store_true", help="include children")
-    parser.add_argument("-l", "--list", action="store_true", help="list available plugins and loaders")
+    parser.add_argument(
+        "-l",
+        "--list",
+        action="store",
+        nargs="?",
+        const="*",
+        default=None,
+        help="list (matching) available plugins and loaders",
+    )
     parser.add_argument("-s", "--strings", action="store_true", help="print output as string")
     parser.add_argument("-d", "--delimiter", default=" ", action="store", metavar="','")
     parser.add_argument("-j", "--json", action="store_true", help="output records as json")
@@ -110,41 +118,71 @@ def main():
                 "The --rewrite-cache option will be ignored as --no-cache or --only-read-cache are specified",
             )
 
-    if args.list:
-        t = Target()
-        fparser = generate_argparse_for_bound_method(t.plugins, usage_tmpl=USAGE_FORMAT_TMPL)
-        fargs, rest = fparser.parse_known_args(rest)
-        t.plugins(**vars(fargs))
-
-        fparser = generate_argparse_for_bound_method(t.loaders, usage_tmpl=USAGE_FORMAT_TMPL)
-        fargs, rest = fparser.parse_known_args(rest)
-        t.loaders(**vars(fargs))
-
-        parser.exit()
-
-    if not args.function:
-        parser.error("argument -f/--function is required")
-
-    functions = args.function.split(",")
-
-    # Show help for a function
+    # Show help for a function or in general
     if "-h" in rest or "--help" in rest:
-        first_func = functions[0]
-        obj = get_attr_for_attr_path(first_func)
-
-        if not obj:
-            parser.error(f"Can't find plugin with function `{first_func}`")
-
+        found_functions = find_plugin_functions(Target(), args.function, False)
+        if not len(found_functions):
+            parser.error("function(s) not found, see -l for available plugins")
+        func = found_functions[0]
+        if issubclass(func.class_object, OSPlugin):
+            func.class_object = OSPlugin
+        obj = getattr(func.class_object, func.method_name)
         if isinstance(obj, type) and issubclass(obj, Plugin):
             parser = generate_argparse_for_plugin_class(obj, usage_tmpl=USAGE_FORMAT_TMPL)
+        elif isinstance(obj, Callable) or isinstance(obj, property):
+            parser = generate_argparse_for_unbound_method(getattr(obj, "fget", obj), usage_tmpl=USAGE_FORMAT_TMPL)
         else:
-            parser = generate_argparse_for_unbound_method(obj, usage_tmpl=USAGE_FORMAT_TMPL)
-
+            parser.error(f"can't find plugin with function `{func.method_name}`")
         parser.print_help()
+        parser.exit()
+
+    # Show the list of available plugins for the given optional target and optional
+    # search pattern, only display plugins that can be applied to ANY targets
+    if args.list:
+        collected_plugins = {}
+
+        if args.targets:
+            for target in args.targets:
+                plugin_target = Target.open(target)
+                funcs = find_plugin_functions(plugin_target, args.list, True)
+                for func in funcs:
+                    collected_plugins[func.name] = func.plugin_desc
+        else:
+            funcs = find_plugin_functions(Target(), args.list, False)
+            for func in funcs:
+                collected_plugins[func.name] = func.plugin_desc
+
+        # Display in a user friendly manner
+        target = Target()
+        fparser = generate_argparse_for_bound_method(target.plugins, usage_tmpl=USAGE_FORMAT_TMPL)
+        fargs, rest = fparser.parse_known_args(rest)
+
+        if collected_plugins:
+            target.plugins(list(collected_plugins.values()))
+
+        # No real targets specified, show the available loaders
+        if not args.targets:
+            fparser = generate_argparse_for_bound_method(target.loaders, usage_tmpl=USAGE_FORMAT_TMPL)
+            fargs, rest = fparser.parse_known_args(rest)
+            target.loaders(**vars(fargs))
         parser.exit()
 
     if not args.targets:
         parser.error("too few arguments")
+
+    if not args.function:
+        parser.error("argument -f/--function is required")
+
+    # Verify uniformity of output types, otherwise default to records
+    output_types = set()
+    funcs = find_plugin_functions(Target(), args.function, False)
+    for func in funcs:
+        output_types.add(func.output_type)
+
+    default_output_type = None
+    if len(output_types) > 1:
+        log.warning("Mixed output types detected: %s. Only outputting records.", ",".join(output_types))
+        default_output_type = "record"
 
     if args.report_dir and not args.report_dir.is_dir():
         parser.error(f"--report-dir {args.report_dir} is not a valid directory")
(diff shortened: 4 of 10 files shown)


dissect/target/plugin.py
```diff
@@ -218,6 +218,34 @@ def __init_subclass__(cls, **kwargs):
...
         self.target = target
 
+    def is_compatible(self) -> bool:
+        """Perform a compatibility check with the target.
+
+        Perform a compatibility check with the target
+        without raising any exceptions and optionally in a
+        more perrformant manner
+
+        """
```
Schamper, line 228:
```suggestion
        """Perform a compatibility check with the target."""
```
I personally don't find that the rest of the text adds useful information.

dissect/target/plugin.py
```diff
@@ -325,6 +353,7 @@ def version(self) -> str:
         """
         raise NotImplementedError
 
+    @export(record="Record")
```
Schamper, line 356: ?

dissect/target/plugins/browsers/browser.py
```diff
@@ -42,6 +42,10 @@ class BrowserPlugin(Plugin):
         "edge",
     ]
 
+    @staticmethod
+    def is_findable() -> bool:
+        return False
```
Schamper, line 47: Preferably add this below `__init__`.

dissect/target/plugins/general/plugins.py
```diff
@@ -1,16 +1,23 @@
...
+from typing import Dict, List, Tuple, Type, Union
 
 from dissect.target import plugin
 from dissect.target.helpers.docs import INDENT_STEP, get_plugin_overview
-from dissect.target.plugin import Plugin, arg, export
+from dissect.target.plugin import Plugin, export
 
 
-def categorize_plugins() -> dict:
+def categorize_plugins(plugins=None) -> dict:
```
Schamper, line 12: Type hint?

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
     def check_compatible(self):
         return True
 
-    @export(output="none")
-    @arg("--docs", dest="print_docs", action="store_true")
-    def plugins(self, print_docs=False):
-
-        categorized_plugins = dict(sorted(categorize_plugins().items()))
+    @export(output="none", cache=False)
+    def plugin_function_db(self) -> Tuple[dict, list]:
```
Schamper, line 118:
```suggestion
    def plugin_function_db(self) -> tuple[dict, list]:
```

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+        db = []
+        rootset = set()
+
+        def all_plugins():
```
Schamper, line 122: Does `plugin.plugins()` not give the same result?

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+            yield from plugin.plugins()
+            yield from plugin.os_plugins()
+            yield from plugin.child_plugins()  # Although nothing is exported here...
+
+        for available in all_plugins():
+            # filter other-os plugins
+            if self.target._os_plugin is not None:
+                if (
+                    available["module"].endswith("_os")
+                    and self.target._os_plugin.__module__.find(available["module"]) == -1
```
Schamper, line 132:
```suggestion
                    and available["module"] not in self.target._os_plugin.__module__
```
?

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+                    "plugin": available,
+                }
+                db.append(entry)
+
+        return db, rootset
+
+    @export(output="none", cache=False)
+    def find_plugin_functions(
+        self, patterns: str, children: bool = False, compatibility: bool = False
+    ) -> List[PluginFunction]:
```
Schamper, line 150:
```suggestion
    ) -> list[PluginFunction]:
```

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+                            method_name=funcname,
+                            output_type=getattr(fobject, "__output__", "text"),
+                            plugin_desc=plugindesc,
+                        )
+                    )
+        return result
+
+    @export(output="none", cache=False)
+    def plugins(self, plugins=None):
+
```
Schamper, line 223:
```suggestion
```

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+                            class_object=loaded_plugin_object,
+                            method_name=funcname,
+                            output_type=getattr(fobject, "__output__", "text"),
+                            plugin_desc=plugindesc,
+                        )
+                    )
+        return result
+
+    @export(output="none", cache=False)
+    def plugins(self, plugins=None):
```
Schamper, line 222: What happened to the `--docs` argument?

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+                entry = {
+                    "name": f"{available['module']}.{exported}",
+                    "plugin": available,
+                }
+                db.append(entry)
+
+        return db, rootset
+
+    @export(output="none", cache=False)
+    def find_plugin_functions(
```
Schamper, line 148: I'm a bit confused by this method and the one above it. Do they really need to be `@export`ed? Can't these be helper functions in `plugin.py`?

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
         t = Target()
         fparser = generate_argparse_for_bound_method(t.plugins, usage_tmpl=USAGE_FORMAT_TMPL)
         fargs, rest = fparser.parse_known_args(rest)
-        t.plugins(**vars(fargs))
-
-        fparser = generate_argparse_for_bound_method(t.loaders, usage_tmpl=USAGE_FORMAT_TMPL)
-        fargs, rest = fparser.parse_known_args(rest)
-        t.loaders(**vars(fargs))
-
+        t.plugins(collected_plugins.values())  # , True, **vars(fargs)
```
Schamper, line 176: Left over comments?

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
+        # No real targets specified, user might be interested to see available loaders
+        if len(tasklist) == 1 and not tasklist[0][0]._applied:
```
Schamper, line 178: Does this matter? You use the fake `t` Target here anyway.

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
+    # either for the list or for execution, we collect all the work to be done here
+    tasklist = []
+    for target in targets:
+        if args.child:
+            try:
+                target = target.open_child(args.child)
+            except Exception:
+                continue
+        functions = target.find_plugin_functions(args.function, args.compatible)
+        tasklist.append((target, functions))
```
Schamper, line 154: Not a big fan of this approach. If you execute anything in a directory with 1000+ targets in it, it'll first attempt to load all of those targets before doing anything.

Schamper (changes requested):

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
+        if not total_num_tasks:
+            parser.error("Function(s) not found, see -l for available plugins")
+            parser.exit()
+        target, func_list = tasklist[0]
+        func = func_list[0]
+        obj = getattr(func.class_object, func.method_name)
         if not obj:
-            parser.error(f"Can't find plugin with function `{first_func}`")
-
+            parser.error(f"Can't find plugin with function `{func.method_name}`")
```
pyrco, line 214: Is this even possible wuth `tasklist` being build up using `target.find_plugin_functions()`?

dissect/target/tools/query.py
pyrco: We probably only want this wwhen `--compatible` is specified, so it is explicit that this may be an expensive operation.

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
+                target = target.open_child(args.child)
+            except Exception:
+                continue
+        functions = target.find_plugin_functions(args.function, args.compatible)
+        tasklist.append((target, functions))
+
+    # Show the list of available plugins for the given optional target and optional
+    # search pattern, only display plugins that can be applied to ALL targets
     if args.list:
+        collected_plugins = None
```
pyrco, line 159:
```suggestion
        collected_plugins = {}
```

dissect/target/tools/query.py
```diff
@@ -110,67 +118,127 @@ def main():
...
+        for index, task in enumerate(tasklist):
+            collected = {}
+            target, funcs = task
+            for func in funcs:
+                collected[func.name] = func.plugin_desc
```
pyrco, line 164: Why this extra step with `collected`? Dict values are unique anyway, so just
```suggestion
                collected_plugins[func.name] = func.plugin_desc
```
should work.

pyrco: Ah, I guess because of the constraint that a plugin should be executable on ALL of targets and their children. But I think that is not what we want, it should be ANY target and their children.

dissect/target/plugins/general/plugins.py
```diff
@@ -94,15 +101,128 @@ def get_description_dict(
...
+        result = []
+        functions, rootset = self.plugin_function_db()
+
+        for pattern in patterns.split(","):
+
+            wildcard = any(char in pattern for char in ["*", "!", "?", "[", "]"])
+            treematch = pattern.split(".")[0] in rootset and pattern != "os"
+
+            if treematch and not wildcard:
+                pattern += "*"  # this seems to be what people expect...
```
pyrco, line 160: Shouldn't this be `.*` as the pattern ends in the last tree part, e.g. `foo.bar` -> `foo.bar.*` instead of `foo.bar*`?

pyrco (changes requested):

dissect/target/plugins/general/plugins.py
cecinestpasunepipe: Not (all) osplugins and childplugins

dissect/target/tools/query.py
Schamper: Is there no way to avoid this approach completely? I'd much prefer that.

dissect/target/plugins/general/plugins.py
cecinestpasunepipe: Ok, moved

dissect/target/tools/query.py
cecinestpasunepipe: That is not the issue, I thought it would be nice to only show the loaders if there are no targets. I refactored this code because it was ugly btw. Loaders take up a lot of space in my terminal, should we always display that chunk?

dissect/target/plugin.py
```diff
@@ -218,6 +218,34 @@ def __init_subclass__(cls, **kwargs):
...
+        """
+        try:
+            self.check_compatible()
+        except UnsupportedPluginError:
+            return False
+
+        return True
+
+    @staticmethod
+    def is_findable() -> bool:
```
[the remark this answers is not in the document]
cecinestpasunepipe: Changed, but without _ because in other comments, there was the request to move the logic to helpers, so it now needs to be accessible from the outside.

dissect/target/plugins/general/plugins.py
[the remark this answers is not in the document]
cecinestpasunepipe: plugin_selection then, slightly more descriptive?

dissect/target/tools/query.py
cecinestpasunepipe: Should not be, but otherwise the error message will probably be quite cryptic...

pyrco (changes requested):

dissect/target/plugin.py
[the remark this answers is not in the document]
cecinestpasunepipe: But if it's a public property one can never be sure the value hasn't been altered anywhere else in the code right? Python does not have read-only properties as far as I know.

dissect/target/tools/query.py
cecinestpasunepipe: Due to a programming error the value of obj can still be Falsy (like None)?

dissect/target/tools/query.py
```diff
@@ -110,67 +149,124 @@ def main():
...
         t = Target()
         fparser = generate_argparse_for_bound_method(t.plugins, usage_tmpl=USAGE_FORMAT_TMPL)
         fargs, rest = fparser.parse_known_args(rest)
-        t.plugins(**vars(fargs))
-
-        fparser = generate_argparse_for_bound_method(t.loaders, usage_tmpl=USAGE_FORMAT_TMPL)
-        fargs, rest = fparser.parse_known_args(rest)
-        t.loaders(**vars(fargs))
-
+        t.plugins(list(collected_plugins.values()))
```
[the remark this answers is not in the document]
cecinestpasunepipe: ok

dissect/target/plugin.py
```diff
@@ -796,3 +824,119 @@ def __init_subclass__(cls, **kwargs):
...
+class PluginFunction:
+    name: str
+    output_type: str
+    class_object: str
+    method_name: str
+    plugin_desc: dict
+
+
+def plugin_function_index(target: Target) -> tuple[dict, list]:
+    db = []
```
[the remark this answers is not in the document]
cecinestpasunepipe: It has evolved to that yes, I think that would be better.

dissect/target/plugin.py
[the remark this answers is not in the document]
pyrco: That's not possible, but such is life when using Python. It's ok I guess if someone explicitly changes this value for an instance at some point. Normally you use a single or double `_` to indicate there will be dragons when you do this.
Maybe specifying a `_findable` class variable and returning that in `findable()` will get the best of both worlds, configuring subclasses will be a simple line of code, the variable is indicated to be private and there is a public way of getting the value through the function.

pyrco (changes requested):

dissect/target/plugin.py
```diff
@@ -796,3 +824,113 @@ def __init_subclass__(cls, **kwargs):
...
+            else:
+                funcname = pattern
+                namespace = None
+
+            plugindesc = None
+            for index_name, func in functions.items():
+                nsmatch = namespace and func["namespace"] == namespace and funcname in func["exports"]
+                fmatch = not namespace and not func["namespace"] and funcname in func["exports"]
+                if nsmatch or fmatch:
+                    plugindesc = func
```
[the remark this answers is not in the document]
cecinestpasunepipe: Yes, I suggested this change on purpose since `getattr` might execute plugin code prematurely (before we establish based on static analysis of the return type whether we want to execute this function or not). Also the return value is dynamic which might lead to insecurities in the CLI (in theory you can never predict the effect unless we peparse the underlying Python code).

dissect/target/plugin.py
```diff
@@ -796,3 +824,113 @@ def __init_subclass__(cls, **kwargs):
...
+    result = []
+    functions, rootset = plugin_function_index(target)
+
+    for pattern in patterns.split(","):
+
+        wildcard = any(char in pattern for char in ["*", "!", "?", "[", "]"])
+        treematch = pattern.split(".")[0] in rootset and pattern != "os"
+
+        if treematch and not wildcard:
+            pattern += ".*"  # this seems to be what people expect...
```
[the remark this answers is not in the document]
cecinestpasunepipe: Done, also changed back to only *

dissect/target/tools/query.py
cecinestpasunepipe: Done

dissect/target/plugin.py
```diff
@@ -809,3 +832,118 @@ def __init_subclass__(cls, **kwargs):
...
+@dataclass
+class PluginFunction:
+    name: str
+    output_type: str
+    class_object: str
+    method_name: str
+    plugin_desc: dict
+
+
+def plugin_function_index(target: Target) -> tuple[list[dict[str, Any]], set[str]]:
```
pyrco, line 846:
```suggestion
def plugin_function_index(target: Target) -> tuple[dict[str, Any], set[str]]:
```

pyrco (changes requested):

dissect/target/tools/query.py
```diff
@@ -110,66 +154,125 @@ def main():
...
+    if len(output_types) > 1:
+        log.warning(f"Mixed output types detected: {'/'.join(output_types)}, only outputting records.")
+        default_output_type = "record"
 
+    # Show help for a function
+    if "-h" in rest or "--help" in rest:
+        if not total_num_tasks:
+            parser.error("Function(s) not found, see -l for available plugins")
+            parser.exit()
+        _, func_list = list(tasks())[0]
```
pyrco, line 242: Probably the only unfortunate thing here is that when doing a `target-query -f browser -h <TARGET>` the TARGET will be opened although it is not needed.

cecinestpasunepipe: Fixed

pyrco (approved):

pyrco:
LGTM

dissect/target/tools/query.py
```diff
@@ -57,7 +83,16 @@ def main():
...
+    parser.add_argument(
+        "-l",
+        "--list",
+        action="store",
+        nargs="?",
+        const="*",
+        default=False,
+        help="list (matching) available plugins and loaders",
+    )
+    parser.add_argument("--compatible", action="store_true", help="full plugin compatibility check")
```
Schamper, line 95: Can you sell me on this addition? I still fail to see the added benefit of this flag. I can see it being useful when being used exclusively in combination with `-l`, but not for execution, because it will only add latency for no real benefit.

Schamper (changes requested):
Does this also fix the absence of the OS functions in `-l` btw?

dissect/target/plugin.py
```diff
@@ -198,6 +201,15 @@ class attribute. Namespacing results in your plugin needing to be prefixed
     """Defines the plugin namespace."""
     __record_descriptors__: list[RecordDescriptor] = None
     """Defines a list of :class:`~flow.record.RecordDescriptor` of the exported plugin functions."""
+    __findable__: bool = True
+    """
+    Determins whether this plugin will be revealed when using search patterns.
+    Some (meta)-plugins are not very suitable for wild cards on CLI or
```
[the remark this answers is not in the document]
cecinestpasunepipe: > Does this also fix the absence of the OS functions in `-l` btw?

OS functions are listed as well

dissect/target/tools/query.py
cecinestpasunepipe: - Clean output, no unnecessary notices between the records when using with a wildcard
- Uniform output, i.e. same records for every target when used with a wildcard, this is nice for streaming to csvs or databases (this feature was rejected so this no longer applies)

I believe @pyrco also mentioned that this was a cleaner way of doing things.
To be honest, I thought the 2nd reason was quite nice, but now that no longer applies I do not see the benefit anymore, however just to be sure @pyrco are you okay with removing this flag?

Schamper: I personally don't find the added complexity and latency worth it for reason 1.

pyrco: I had in mind that `--compatible` was implicit once you specified a target with `-l`, but apparently it isn't.
I think it is useful to have `--compatible` in combination with `-l`, I think it is even more useful to have `--compatible` be automatic with `-l` if you also supply a target.

From the looks of it only implementing `--compatible` behavior for the `-l` option seems to be doable with much less complexity. I think weighing this against the "correctness" of not executing things that you know beforehand won't work does slightly tip in favor of a less complex implementation.

pyrco (commented):
See my reply in  the thread

dissect/target/tools/query.py
Schamper: How does that work if you specify multiple targets by the way? Do you just print the compatible plugin list for each specified target?

cecinestpasunepipe: No just the plugins that match any target.

cecinestpasunepipe: Adjusted, --compatible is now implied with list only

Schamper: The seeming complexity in `query.py` stayed though, is that intentional?

cecinestpasunepipe: I will see if I can simplify it further

cecinestpasunepipe:
> Does this also fix the absence of the OS functions in `-l` btw?

I list them under a header called OS, so you do not see them multiple times (if there is no target).
If there is a target you see the OS functions under _os for the specified target.

Without target:

```
    OS:
      architecture - No documentation (output: text)
      dhcp - No documentation (output: text)
      dns - No documentation (output: text)
      domain - No documentation (output: text)
      gateway - No documentation (output: text)
      get_all_records - Return the records of all exported methods. (output: records)
      hostname - No documentation (output: text)
      interface - No documentation (output: text)
      ips - Returns a list of static IP addresses and DHCP lease IP addresses found on the host system. (output: text)
      netmask - No documentation (output: text)
      os - No documentation (output: text)
      users - Recover users from /etc/passwd, /etc/master.passwd or /var/log/syslog session logins. (output: records)
      version - No documentation (output: text)
```

with target (Windows):

```
windows:
        _os:
          architecture - Returns a dict containing the architecture and bitness of the system (output: text)
          codepage - Returns the current active codepage on the system. (output: text)
          get_all_records - Return the records of all exported methods. (output: records)
          hostname - No documentation (output: text)
          ips - No documentation (output: text)
          os - No documentation (output: text)
          users - No documentation (output: records)
          version - Return a string representation of the Windows version of the target. (output: text)
```

Otherwise you cannot know what OS functions are available right?
Also, unfortunately some OSes seem to have functions that other OSes do not have, like codepage (not available for UNIX right now...)

Schamper:
@cecinestpasunepipe yeah we need to promote some functions to become default OS plugin level functions.

Can you check why `get_all_records` is in there? I wouldn't expect that to be in that list.

dissect/target/tools/query.py
```diff
@@ -110,41 +118,78 @@ def main():
...
+            funcs = find_plugin_functions(Target(), args.function, False)
+            for func in funcs:
+                collected_plugins[func.name] = func.plugin_desc
+
+        # Display in a user friendly manner
+        target = Target()
+        fparser = generate_argparse_for_bound_method(target.plugins, usage_tmpl=USAGE_FORMAT_TMPL)
+        fargs, rest = fparser.parse_known_args(rest)
+
+        if len(collected_plugins):
```
Schamper, line 173:
```suggestion
        if collected_plugins:
```

dissect/target/tools/query.py
```diff
@@ -110,41 +118,78 @@ def main():
...
+            target.plugins(list(collected_plugins.values()))
+
+        # No real targets specified, user might be interested to see available loaders
```
Schamper, line 176:
```suggestion
        # No real targets specified, show the available loaders
```

dissect/target/tools/query.py
```diff
@@ -110,41 +118,78 @@ def main():
...
+
+    # Verify uniformity of output types, otherwise default to records
+    output_types = set()
+    funcs = find_plugin_functions(Target(), args.function, False)
+    for func in funcs:
+        output_types.add(func.output_type)
+
+    default_output_type = None
+    if len(output_types) > 1:
+        log.warning("Mixed output types detected: %s, only outputting records.", "/".join(output_types))
```
Schamper, line 191:
```suggestion
        log.warning("Mixed output types detected: %s. Only outputting records.", ",".join(output_types))
```
"," makes more sense in my opinion.

dissect/target/tools/query.py
```diff
@@ -230,15 +280,23 @@ def main():
...
+                        else:
+                            rs.write(record_entries)
+                        count += 1
+                        if args.limit is not None and count >= args.limit:
+                            break_out = True
+                            break
+                except Exception as problem:
+                    # Ignore errors if multiple functions
+                    if len(funcs) > 1:
+                        target.log.error(f"Exception occurred while processing output of {func}", exc_info=problem)
```
Schamper, line 296:
```suggestion
                except Exception as exc:
                    # Ignore errors if multiple functions
                    if len(funcs) > 1:
                        target.log.error("Exception occurred while processing output of %s", func, exc_info=exc)
```
Or `e` is popularly used in the codebase.

Schamper: Also why was this try/except actually added? I don't think anything changed here?

dissect/target/tools/query.py
```diff
@@ -230,15 +280,23 @@ def main():
...
+                        pass
                     else:
-                        rs.write(record_entries)
-                    count += 1
-                    if args.limit is not None and count >= args.limit:
-                        break_out = True
-                        break
+                        raise problem
```
Schamper, line 299:
```suggestion
                        raise exc
```

Schamper (changes requested):

cecinestpasunepipe:
> @cecinestpasunepipe yeah we need to promote some functions to become default OS plugin level functions.
> 
> Can you check why `get_all_records` is in there? I wouldn't expect that to be in that list.

should be fixed now

dissect/target/target.py
```diff
@@ -632,6 +632,14 @@ def has_function(self, function: str) -> bool:
         except PluginError:
             return False
 
+    def os_plugin_class_name(self) -> str:
```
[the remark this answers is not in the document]
cecinestpasunepipe: It is meant to be called from outside the object (plugin_function_index), so making it private does not make sense to me? There needs to be a way to determine whether a target has an os plugin associated with it and if so, which one that is. Otherwise it is difficult to match the correct OS plugin, given the fact that they have no compatible_check() method.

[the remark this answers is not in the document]
Schamper: I usually consider these kinds of things to be "project internal" rather than "object internal". There's really no reason to call that method outside of any of the internal code in `dissect.target`.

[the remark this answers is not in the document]
cecinestpasunepipe: Okay, well if that is the case, am I allowed to access _os_plugin from the outside directly then?

dissect/target/tools/query.py
```diff
@@ -57,7 +57,15 @@ def main():
...
     parser.add_argument("--child", help="load a specific child path or index")
     parser.add_argument("--children", action="store_true", help="include children")
-    parser.add_argument("-l", "--list", action="store_true", help="list available plugins and loaders")
+    parser.add_argument(
+        "-l",
+        "--list",
+        action="store",
+        nargs="?",
+        const="*",
+        default=False,
```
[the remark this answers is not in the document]
cecinestpasunepipe: Fixed, left-over from a previous iteration. No longer needed.

dissect/target/tools/query.py
```diff
@@ -110,41 +118,78 @@ def main():
...
-        first_func = functions[0]
-        obj = get_attr_for_attr_path(first_func)
-
-        if not obj:
-            parser.error(f"Can't find plugin with function `{first_func}`")
-
+        found_functions = find_plugin_functions(Target(), args.function, False)
+        if not len(found_functions):
+            parser.error("Function(s) not found, see -l for available plugins")
+            parser.exit()
```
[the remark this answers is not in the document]
cecinestpasunepipe: OK

dissect/target/tools/query.py
cecinestpasunepipe: If you run multiple functions, through the use of a wild card I assumed you don't want to stop if one of those in the set causes a problem?

dissect/target/tools/query.py
```diff
@@ -110,41 +118,78 @@ def main():
...
     if "-h" in rest or "--help" in rest:
-        first_func = functions[0]
-        obj = get_attr_for_attr_path(first_func)
-
-        if not obj:
-            parser.error(f"Can't find plugin with function `{first_func}`")
-
+        found_functions = find_plugin_functions(Target(), args.function, False)
+        if not len(found_functions):
+            parser.error("Function(s) not found, see -l for available plugins")
```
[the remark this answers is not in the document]
cecinestpasunepipe: Without Nitpicky no Perfection :wink:

dissect/target/target.py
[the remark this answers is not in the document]
Schamper: I don't see why not. This is all internal code of `dissect.target`, it should be allowed to access `_` methods and attributes.

dissect/target/tools/query.py
Schamper: Ahh nevermind, I read the code wrong. I didn't realize that `entry` would be the generator from the plugins themselves. Carry on.

dissect/target/target.py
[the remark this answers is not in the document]
cecinestpasunepipe: Okay, well in a "traditional" sense of OOP, this was not allowed of course, but I am happy with your "project internal" definition of OO encapsulation! (:smiley:)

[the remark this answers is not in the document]
Schamper: Yeah, but then we'd have a `InternalTarget` object and `getInternalState` methods ;) I prefer to bend the rules a little bit here for a bit more readable code.

dissect/target/plugin.py
```diff
@@ -774,7 +795,7 @@ def _modulepath(cls) -> str:
 
 def get_plugin_classes_with_method(method_name: str) -> Iterator[Type[Plugin]]:
```
pyrco, line 796: This function is no longer used (only user was `tools/utils.py`) and can be removed. Same for `get_plugin_classes_by_namespace()`

dissect/target/plugin.py
```diff
@@ -564,34 +585,34 @@ def child_plugins() -> Iterator[PluginDescriptor]:
...
+    yield from get_plugins_by_namespace(index_name, osfilter=osfilter)
 
 
-def get_plugins_by_func_name(func_name: str, osfilter: str = None) -> Iterator[PluginDescriptor]:
+def get_plugins_by_index_name(index_name: str, osfilter: str = None) -> Iterator[PluginDescriptor]:
     """Get a plugin descriptor by function name.
 
     Args:
-        func_name (str): Function name to lookup.
+        index_name (str): Function name to lookup.
```
pyrco, line 603: Does this mean this now has the same form (e.g. `"namespace.plugin"`) as the keys from `index` returned by `plugin_function_index()`?

cecinestpasunepipe: No, those namespaces are limited to the __namespace__ property, so only one level deep. Or maybe I misunderstand the question?

dissect/target/plugin.py
```diff
@@ -564,34 +585,34 @@ def child_plugins() -> Iterator[PluginDescriptor]:
...
+        if not plugin_desc["namespace"] and index_name in plugin_desc["functions"]:
             yield plugin_desc
 
 
 def get_plugins_by_namespace(namespace: str, osfilter: str = None) -> Iterator[PluginDescriptor]:
     """Get a plugin descriptor by namespace.
 
     Args:
-        func_name: Function name to lookup.
+        namespace: Namespace to match.
```
pyrco, line 615:
```suggestion
        namespace: Plugin namespace to match.
```
?

dissect/target/plugin.py
pyrco: Then wouldn't `func_name`/`get_plugins_by_func_name()` not be more descriptive / less confusing? Now it gives the idea it has something to do with `plugin_function_index()`

cecinestpasunepipe: I think those were changes requested to improve consistency (which it does), but you're right, now I look at it again I think it's a bit confusing on the other hand. Those are really separate functions and I don't see a way to reuse them easily. So better not associate them with the` plugin_function_index()`

tests/test_plugins_general_plugins.py
```diff
@@ -60,8 +60,8 @@ def test_plugin_description_in_dict_multiple():
     ]
 
 
-@patch.object(plugin.plugin, "load")
+@patch.object(plugin, "load")
 @patch.object(plugin, "get_exported_plugins")
-def test_categorize_plugins(mocked_export, mocked_load):
+def _test_categorize_plugins(mocked_export, mocked_load):
```
Schamper, line 65: Disabled test?

cecinestpasunepipe: That file should not have been altered, reset.

pyrco:
LGTM

pyrco (approved):

Schamper (approved):

PR #127, opened 2023-01-12
Add nginx, apache and caddy webserver log parsers

JSCU-CNI:
This PR adds support for nginx, apache and caddy webserver log parsing. This PR also moves the existing IIS parser to the `plugins/apps/webservers` namespace.

PR #131, opened 2023-01-16
Improve unix messages plugin

JSCU-CNI:
This PR adds support for the CentOS `/var/log/messages` equivalent of Debian's `/var/log/syslog`.

Support for `/var/log/secure` and `/var/log/auth.log` will be put in a separate PR.


dissect/target/helpers/fsutil.py
```diff
@@ -963,3 +965,23 @@ def resolve_link(
...
+
+
+def decompress_and_readlines(path: TargetPath) -> [str]:
+    if path.suffix == ".gz":
+        return gzip.GzipFile(fileobj=path.open()).read().decode().splitlines()
+    elif path.suffix == ".bz2":
+        bytestream = path.open().read()
+        return bz2.decompress(bytestream).decode().splitlines()
+    else:
+        return path.open("rt").readlines()
```
Schamper, line 987:
```suggestion
def open_decompress(path: TargetPath, mode: str = "rb"):
    if path.suffix == ".gz":
        import gzip
        
        return gzip.open(path.open(), mode)
    elif path.suffix == ".bz2":
        import bz2
        
        return bz2.open(path.open(), mode)
    else:
        return path.open(mode)
```
What about something like this? That way it's only one function and it allows for some flexibility in how to use it, e.g. (imagine the string is a valid `TargetPath`:

```python
bytes_buf = open_decompress("/dir/file.gz").read()
for line in open_decompress("/dir/file.gz", "rt"):
    print(line)
```

Moving the imports inline is because these modules might not always be available (which is the case for our custom internal CPython).

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -1,41 +1,35 @@
...
     "linux/messages",
     [
         ("datetime", "ts"),
-        ("uri", "filepath"),
         ("string", "daemon"),
         ("varint", "pid"),
-        ("string", "msg"),
+        ("string", "message"),
         ("string", "raw"),
+        ("uri", "source"),
```
Schamper, line 16:
```suggestion
        ("path", "source"),
```

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
-                            daemon=daemon,
-                            pid=pid,
-                            msg=msg,
-                            raw=line,
-                            _target=self.target,
-                        )
+
+        log_files: [TargetPath] = list(self.target.fs.path("/var/log/").glob("syslog*")) + list(
+            self.target.fs.path("/var/log/").glob("messages*")
+        )
```
Schamper, line 48: If you use `itertools.chain` you can make this a proper generator.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+
+        for log_file in log_files:
+
+            file_ctime = self.target.fs.get(str(log_file)).stat().st_ctime
+            year_file_created = datetime.fromtimestamp(file_ctime).year
```
Schamper, line 53: Preferably use `ts.from_unix` from `dissect.util`, since that will make sure it's using the UTC timezone and not a local one.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+            last_seen_year = year_file_created
+            last_seen_month = 0
+
+            for line in decompress_and_readlines(log_file):
+
+                # Line can be a byte object on debian and centos
+                if isinstance(line, bytes):
+                    line = line.decode("utf-8").strip()
```
Schamper, line 61: With the current code that should be impossible due to the `.decode()`, and with the suggested code that should also be impossible due to opening it with `rt`.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+                try:
+                    ts = datetime.strptime(RE_TS.search(line).groups()[0], "%b %d %H:%M:%S")
+                    if last_seen_month > ts.month:
+                        last_seen_year += 1
+                    last_seen_month = ts.month
+                    ts = ts.replace(year=last_seen_year)
+                except Exception as e:
+                    self.target.log.warn(f"Could not convert timestamp line in {log_file}: {e}")
+                    # Set ts to epoch 1970-01-01 if we could not convert
+                    ts = datetime.utcfromtimestamp(0)
```
Schamper, line 77: Preferably use `ts.from_unix(0)` to make sure it's a UTC timezone aware timestamp.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+                except Exception:
+                    msg = line
+
+                yield MessagesRecord(
+                    ts=ts,
+                    daemon=daemon,
+                    pid=pid,
+                    message=msg,
+                    raw=line,
+                    source=str(log_file),
```
Schamper, line 104:
```suggestion
                    source=path.from_posix(log_file),
```
And add `from flow.record.fieldtypes import path`.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+                    continue
+
+                try:
+                    ts = datetime.strptime(RE_TS.search(line).groups()[0], "%b %d %H:%M:%S")
+                    if last_seen_month > ts.month:
+                        last_seen_year += 1
+                    last_seen_month = ts.month
+                    ts = ts.replace(year=last_seen_year)
+                except Exception as e:
+                    self.target.log.warn(f"Could not convert timestamp line in {log_file}: {e}")
```
Schamper, line 75:
```suggestion
                    self.target.log.warn("Could not convert timestamp line in %s: %s", log_file, e)
```

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -48,68 +42,65 @@ def messages(self):
...
+
+                if not line or line == "":
+                    continue
+
+                try:
+                    ts = datetime.strptime(RE_TS.search(line).groups()[0], "%b %d %H:%M:%S")
+                    if last_seen_month > ts.month:
+                        last_seen_year += 1
+                    last_seen_month = ts.month
+                    ts = ts.replace(year=last_seen_year)
```
Schamper, line 73: I believe these timestamps are in local timezone? Preferably try to use the new locale plugin to get the correct timezone:
https://docs.python.org/3/library/datetime.html#timezone-objects

You can add the timezone using `ts.replace(year=last_seen_year, tzinfo=<timezone object here>)`

You can look at the Windows `datetime` plugin for inspiration:
https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/windows/datetime.py

Schamper (changes requested):

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -40,34 +42,39 @@ def messages(self):
...
-                    pass
-            if "bz2" in f.name:
-                fh = bz2.decompress(fh)
-                fh = StringIO(fh.decode())
-
-            for line in fh:
-                line = line.strip()
-                if not line:
+
+        log_files: [TargetPath] = list(self.target.fs.path("/var/log/").glob("syslog*")) + list(
```
Schamper, line 46: Can you adjust the docstring to also mention `syslog` btw?

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -1,33 +1,35 @@
...
 RE_MSG = re.compile(r"[^:]+:\d+:\d+[^:]+:\s(.*)$")
 
 
 class MessagesPlugin(Plugin):
     def check_compatible(self):
-        pass
+        return self.target.fs.path("/var/log/syslog").exists() or self.target.fs.path("/var/log/messages").exists()
+
+    @export(record=MessagesRecord)
+    def syslog(self):
```
Schamper, line 31: And add a docstring here. Just something simple such as:
```suggestion
    def syslog(self):
    """Return contents of /var/log/messages and /var/log/syslog.
    
    See ``messages`` for more information.
    """
```

Schamper (changes requested):

dissect/target/helpers/fsutil.py
```diff
@@ -967,21 +965,45 @@ def resolve_link(
...
 
-def decompress_and_read(path: TargetPath) -> bytes:
-    if path.suffix == ".gz":
-        return gzip.GzipFile(fileobj=path.open()).read()
-    elif path.suffix == ".bz2":
-        bytestream = path.open().read()
-        return bz2.decompress(bytestream)
-    else:
-        return path.open().read()
+# def decompress_and_read(path: TargetPath) -> bytes:
```
Schamper: Forgot to remove the commented code I presume.

dissect/target/helpers/fsutil.py
```diff
@@ -967,21 +965,45 @@ def resolve_link(
...
+    """Open and decompress a file. Can handle regular files, gz and bz2 files.
+
+    Does not check if the TargetPath exists.
+
+    Examples:
+    bytes_buf = open_decompress("/dir/file.gz").read()
+
+    for line in open_decompress("/dir/file.gz", "rt"):
+        print(line)
+    """
```
Schamper, line 978:
```suggestion
    """Open and decompress a file. Handles gz, bz2 and regular files.

    Assumes that the ``path`` exists.

    Example:
        bytes_buf = open_decompress(Path("/dir/file.gz")).read()
    
        for line in open_decompress(Path("/dir/file.gz"), "rt"):
            print(line)
    """
```
Reads a bit nicer.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -42,39 +51,33 @@ def messages(self):
         Sources:
             - https://geek-university.com/linux/var-log-messages-file/
         """
-
-        log_files: [TargetPath] = list(self.target.fs.path("/var/log/").glob("syslog*")) + list(
-            self.target.fs.path("/var/log/").glob("messages*")
-        )
+        syslogs = list(self.target.fs.path("/var/log/").glob("syslog*"))
+        messages = list(self.target.fs.path("/var/log/").glob("messages*"))
+        log_files: Generator[TargetPath] = chain(syslogs, messages)
```
Schamper, line 56:
```suggestion
        syslogs = self.target.fs.path("/var/log/").glob("syslog*")
        messages = self.target.fs.path("/var/log/").glob("messages*")
        log_files: Iterator[TargetPath] = chain(syslogs, messages)
```
The idea here is that you chain the generators from `glob`. Right now you completely exhaust the `glob` generators before starting to parse files. By chaining their respective generators you evaluate the glob lazily and can start returning output earlier.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -42,39 +51,33 @@ def messages(self):
...
-                # Line can be a byte object on debian and centos
-                if isinstance(line, bytes):
-                    line = line.decode("utf-8").strip()
-                else:
-                    line = line.strip()
+            for line in open_decompress(log_file, "rt"):
+                line = line.strip()
 
                 if not line or line == "":
                     continue
```
Schamper, line 69:
```suggestion
                if not line:
                    continue
```
This is the same.

Schamper (changes requested):

dissect/target/helpers/fsutil.py
JSCU-CNI: Removed in 6d028f4b6efa6ebd4d6626580db7181ea75c1fe4.

dissect/target/plugins/os/unix/log/messages.py
Schamper: Are you still looking into this, since you left it unresolved? Or will that be a future fix?

JSCU-CNI: Currently implementing that suggestion, see the first example in https://github.com/fox-it/dissect.target/pull/127/commits/cafcdde77d62e865eb5994c53a3913d8a04ac607.

JSCU-CNI: Added in [6ad6518](https://github.com/fox-it/dissect.target/pull/131/commits/6ad6518d6a8b67b5ffaf0bd14c2ab58d612d6b12).

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -40,34 +61,33 @@ def messages(self):
...
-                fh = bz2.decompress(fh)
-                fh = StringIO(fh.decode())
-
-            for line in fh:
+        syslogs = self.target.fs.path("/var/log/").glob("syslog*")
+        messages = self.target.fs.path("/var/log/").glob("messages*")
+        log_files: Iterator[TargetPath] = chain(syslogs, messages)
+
+        for log_file in log_files:
+
```
Schamper, line 69:
```suggestion
        var_log = self.target.fs.path("/var/log")
        for log_file in chain(var_log.glob("syslog*"), var_log.glob("messages*"):
```
Might be a bit cleaner.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -1,37 +1,58 @@
...
+        try:
+            self.target_timezone = ZoneInfo(f"{target.timezone}")
+        except ZoneInfoNotFoundError:
+            self.target.log.warning("Could not determine timezone of target, falling back to UTC.")
+            self.target_timezone = timezone.utc
+
     def check_compatible(self):
-        pass
+        return self.target.fs.path("/var/log/syslog").exists() or self.target.fs.path("/var/log/messages").exists()
```
Schamper, line 43:
```suggestion
        return any(self.target.fs.path("/var/log").glob("syslog*")) or any(self.target.fs.path("/var/log").glob("messages*"))
```
Might be more accurate, and `any` will ensure it'll return as quickly as possible.

tests/test_plugins_os_unix_log_messages.py
```diff
@@ -0,0 +1,35 @@
...
+    fs_unix.map_file_fh("/etc/timezone", BytesIO("Europe/Amsterdam".encode()))
+
+    data_file = absolute_path("data/unix/logs/messages")
+    fs_unix.map_file("var/log/messages", data_file)
+
+    year = datetime.fromtimestamp(stat(absolute_path("data/unix/logs/messages")).st_ctime).year
+
+    target_unix_users.add_plugin(MessagesPlugin)
+
+    results = list(target_unix_users.messages())
```
Schamper, line 24: Can you add a quick `syslog() == messages()` like test?

Schamper (changes requested):

Schamper:
What are your thoughts on something like this? That way this functionality can be shared among all the other log parsing plugins. Windows targets already have the `tzinfo` internal property, so that'd also create uniformity in that regard.

```patch
diff --git a/dissect/target/plugins/os/unix/locale.py b/dissect/target/plugins/os/unix/locale.py
index 5fbc53a..4eab5f5 100644
--- a/dissect/target/plugins/os/unix/locale.py
+++ b/dissect/target/plugins/os/unix/locale.py
@@ -1,5 +1,8 @@
+from datetime import timezone
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
 from dissect.target.helpers.record import TargetRecordDescriptor
-from dissect.target.plugin import Plugin, export
 
 UnixKeyboardRecord = TargetRecordDescriptor(
     "linux/keyboard",
@@ -35,6 +38,14 @@ class LocalePlugin(Plugin):
             timezone = "/".join(zoneinfo_path[-2:])
             return timezone
 
+    @internal(property=True)
+    def tzinfo(self):
+        try:
+            return ZoneInfo(self.target.timezone)
+        except (TypeError, ZoneInfoNotFoundError):
+            self.target.log.warning("Could not determine timezone of target, falling back to UTC.")
+            return timezone.utc
+
     @export(property=True)
     def language(self):
         """Get the configured locale(s) of the system."""
diff --git a/dissect/target/plugins/os/unix/log/messages.py b/dissect/target/plugins/os/unix/log/messages.py
index c3874b5..882958f 100644
--- a/dissect/target/plugins/os/unix/log/messages.py
+++ b/dissect/target/plugins/os/unix/log/messages.py
@@ -29,15 +29,6 @@ RE_MSG = re.compile(r"[^:]+:\d+:\d+[^:]+:\s(.*)$")
 
 
 class MessagesPlugin(Plugin):
-    def __init__(self, target):
-        super().__init__(target)
-
-        try:
-            self.target_timezone = ZoneInfo(f"{target.timezone}")
-        except ZoneInfoNotFoundError:
-            self.target.log.warning("Could not determine timezone of target, falling back to UTC.")
-            self.target_timezone = timezone.utc
-
     def check_compatible(self):
         return any(self.target.fs.path("/var/log").glob("syslog*")) or any(
             self.target.fs.path("/var/log").glob("messages*")
@@ -62,6 +53,8 @@ class MessagesPlugin(Plugin):
         Sources:
             - https://geek-university.com/linux/var-log-messages-file/
         """
+        target_tz = self.target.tzinfo
+
         var_log = self.target.fs.path("/var/log")
         for log_file in chain(var_log.glob("syslog*"), var_log.glob("messages*")):
 
@@ -81,7 +74,7 @@ class MessagesPlugin(Plugin):
                     if last_seen_month > line_ts.month:
                         last_seen_year += 1
                     last_seen_month = line_ts.month
-                    line_ts = line_ts.replace(year=last_seen_year, tzinfo=self.target_timezone)
+                    line_ts = line_ts.replace(year=last_seen_year, tzinfo=target_tz)
                 except Exception as e:
                     self.target.log.warn("Could not convert timestamp line in %s: %s", log_file, e)
                     # Set ts to epoch 1970-01-01 if we could not convert
```

JSCU-CNI:
We added a function called `YearRolloverHelper` to aid in generating correct timestamps for unix log files without year notation.

Usage:
```python
helper = YearRolloverHelper(self.target, log_file, RE_TS, LOG_FORMAT)
for line in open_decompress(log_file, "rt"):
    line_ts = helper.apply_year_rollovers(line)
```

The current setup of `YearRolloverHelper` could be improved, as we're not sure if passing the `RE_TS` and `LOG_FORMAT` to the helper class is the best way to go. This however saves the user of the helper function some repetitive lines of code.

---

*Background*

Unfortunately some unix logs do not store the current year in their log timestamps. We can infer the year of a log line by checking the `st_mtime`* of the parent log file. `st_mtime` will contain an epoch timestamp of when the file was last modified. In order to derive the year of the first log line, we have to iterate over the log file twice**.

The first time we iterate over all log lines to determine the amount of year rollovers inside the file. The second time we iterate over all log lines and can infer the year of the line by subtracting the amount of year rollovers we witnessed from the year of the last modified timestamp (`st_mtime`).

\* `st_mtime` is the only file attribute available in _all_ dissect filesystems. A good second candidate would be `st_crtime` / `st_btime`, but different filesystems have different implementations and names for this attribute. And unfortunately this attribute will never reliably be available for certain loaders (such as gzip, zip or tarball targets).

\*\* In theory we can just look at the first and last log lines. This method would not work with log files which are never rotated, as these can span multiple years. In reality most unix system log files are rotated regularly by default (every day, week or month). However some unix distro's have default logrotate configs with a setting such as `minsize 1M`. It would be a mistake to assume that a file is unlikely to _not_ have multiple year rollovers.

This method will not work reliably on very stale log files, for example files which are only written to a few times a year; where one log line could be from 2021-04-01 and next one from 2022-04-02. However, parsing those log files would be difficult either way, even without inferring the timestamp year.

Resources:
- https://www.geeksforgeeks.org/file-timestamps-mtime-ctime-and-atime-in-linux/
- https://lkml.org/lkml/2010/7/22/249
- https://savannah.gnu.org/forum/forum.php?forum_id=9394
- https://linux.die.net/man/8/logrotate

JSCU-CNI:
Added a test for the year rollover helper in [ff327bd](https://github.com/fox-it/dissect.target/pull/131/commits/ff327bdcf3ec2b292c549d60e7563f72ab59f8e0).

Schamper:
Just pushed an alternative approach for handling the year rollover by iterating over the log files in reverse. Please let me know if this approach works as expected for you.

JSCU-CNI:
Very neat solution, thanks. Can confirm this works for us.

dissect/target/helpers/fsutil.py
```diff
@@ -963,3 +966,96 @@ def resolve_link(
...
+    current_year = from_unix(path.stat().st_mtime).astimezone(tzinfo).year
+    last_seen_month = None
+
+    with open_decompress(path, "rt") as fh:
+        for line in reverse_readlines(fh):
+            line = line.strip()
+            if not line:
+                continue
+
+            relative_ts = datetime.strptime(re.search(re_ts, line).group(0), ts_format)
```
pyrco, line 1056:
```suggestion
            timestamp = re.search(re_ts, line).group(0)
            relative_ts = datetime.strptime(timestamp, ts_format)
```

Makes it easier to visually parse

dissect/target/helpers/fsutil.py
```diff
@@ -963,3 +966,96 @@ def resolve_link(
...
+    Yields tuples of the parsed timestamp and the lines of the file in reverse.
+
+    Args:
+        path: A path to the log file to parse.
+        re_ts: Regex pattern for extracting the timestamp from each line.
+        ts_format: Time format specification for parsing the timestamp.
+        tzinfo: The timezone to use when parsing timestamps.
+    """
+    # Convert the mtime to the local timezone so that we get the correct year
+    current_year = from_unix(path.stat().st_mtime).astimezone(tzinfo).year
```
pyrco, line 1047:
```suggestion
    mtime = path.stat().st_mtime
    current_year = from_unix(mtime).astimezone(tzinfo).year
```

Makes it easier to visually parse

pyrco (changes requested):

Schamper:
The change in `filesystems/tar.py` is temporary until #144 is merged, just to make the test work.

pyrco (approved):

Schamper (approved):

PR #132, opened 2023-01-16
Add installdate function for unix and windows

JSCU-CNI:

PR #136, opened 2023-01-19
Add target-info

JSCU-CNI:
This PR implements the command `target-info`.

`target-info` can show useful basic system properties to the end user, similar to `target-shell`'s `info` command. 

Output formats of `target-info`:
* human readable text (default)
* one-line or pretty json (using `-J` or `-j`)
* `InfoRecord` for `rdump` (using `-r`)

This feature depends on #116, #133 and #132.

Example default output:

```
$ target-info /path/to/targets/example
Disks
- <VmdkContainer size=1337 vs=<DissectVolumeSystem serial=1234>>

Volumes
- <Volume name='part_001' size=1337 fs=<NtfsFilesystem>>
- <Volume name='part_002' size=1337 fs=<NtfsFilesystem>>

Hostname      : WIN-1234
OS family     : windows, WindowsPlugin
OS version    : Windows Server 2019 Standard (NT 6.3) 13337
Arch          : amd64-win64
Domain        : test.local
IPs           : 10.13.37.1
Install date  : 2021-01-01 13:37:00
Last activity : 2023-08-13 21:34:55
Language(s)   : en_US, nl_NL
Timezone      : Europe/Berlin

```


JSCU-CNI:
Note, using `net.ipaddress[]` will currently break `flow.record`. Could be fixed by adding `typedlist` to the imports of `flow.record/flow/record/fieldtypes/net/__init__.py`.

```
AttributeError: module 'flow.record.fieldtypes.net' has no attribute 'typedlist'
```

Schamper:
This will be fixed with https://github.com/fox-it/flow.record/pull/55

dissect/target/plugins/os/windows/generic.py
```diff
@@ -167,37 +167,46 @@ def domain(self):
...
 
-        return None
+        return from_unix(last_seen)
 
     @export(property=True)
     def install_date(self) -> Optional[datetime]:
-        """Returns the install date of the system."""
+        """Returns the install date of the system.
+
+        The value of the registry key is stored as seconds since January 1 1970 inside a 32-bit REG_DWORD.
```
Schamper, line 199:
```suggestion
        The value of the registry key is stored as a Unix epoch timestamp.
```
?

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+    parser.add_argument("-r", "--record", action="store_true", help="print output as record")
+    parser.add_argument("-j", "--json-pretty", action="store_true", help="output records as pretty json")
+    parser.add_argument("-J", "--json", action="store_true", help="output records as one-line json")
+    configure_generic_arguments(parser)
+
+    args, rest = parser.parse_known_args()
+
+    if not args.targets and ("-h" in rest or "--help" in rest):
+        parser.print_help()
+        parser.exit()
```
Schamper, line 59:
```suggestion
    args = parser.parse_args()
```
This tool doesn't have to do anything with additional arguments, so can just use `parse_args()` here.

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+logging.lastResort = None
+logging.raiseExceptions = False
+
+
+def main():
+    """target-info"""
+
+    help_formatter = argparse.ArgumentDefaultsHelpFormatter
+    parser = argparse.ArgumentParser(
+        description="dissect.info",
```
Schamper, line 41:
```suggestion
        description="target-info",
```

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+log = logging.getLogger(__name__)
+logging.lastResort = None
+logging.raiseExceptions = False
+
+
+def main():
+    """target-info"""
+
```
Schamper, line 38:
```suggestion
```

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import argparse
+import json
+import logging
```
Schamper, line 6:
```suggestion
import logging
from pathlib import Path
```

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+
+    help_formatter = argparse.ArgumentDefaultsHelpFormatter
+    parser = argparse.ArgumentParser(
+        description="dissect.info",
+        fromfile_prefix_chars="@",
+        formatter_class=help_formatter,
+        add_help=False,
+    )
+    parser.add_argument("targets", metavar="TARGETS", nargs="*", help="Targets to display info from")
+    parser.add_argument("--from-file", nargs="?", help="file containing targets to load")
```
Schamper, line 47:
```suggestion
    parser.add_argument("--from-file", nargs="?", type=Path, help="file containing targets to load")
```

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if not args.targets and not args.from_file:
+        parser.error("too few arguments")
+
+    if args.from_file:
+        with open(args.from_file, "r") as f:
+            targets = f.read().splitlines()
+            while targets[-1] == "":
+                targets = targets[:-1]
+            args.targets = targets
```
Schamper, line 71:
```suggestion
    if args.from_file:
        if not args.from_file.exists():
            parser.error(f"--from-file {args.from_file} does not exist")

        targets = args.from_file.read_text().splitlines()
        while targets[-1] == "":
            targets = targets[:-1]
        args.targets = targets
```

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+            print(json.dumps(obj_target_info(target)))
+        elif args.json_pretty:
+            print(json.dumps(obj_target_info(target), indent=4))
+        elif args.record:
+            rs = record_output(args.strings, args.json)
+            rs.write(
+                InfoRecord(
+                    os_family=target.os,
+                    os_version=target.version,
+                    ips=list(set(target.ips + target.ips_dhcp)),
```
Schamper, line 85: `ips_dhcp` does not exist on Windows targets and thus will error.

You can use `target.has_function("ips_dhcp")` to check if the target has the supported function.

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def obj_target_info(target):
+    return {
+        "hostname": target.hostname,
+        "os_family": target.os,
+        "os_version": target.version,
+        "architecture": target.architecture,
+        "domain": target.domain,
+        "ips": list(set(target.ips + target.ips_dhcp)),
```
Schamper, line 105: Same here regarding `ips_dhcp`.

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+        "install_date": str(target.install_date),
+        "last_activity": str(target.activity),
+        "language": target.language,
+        "timezone": target.timezone,
+        "disks": [{"type": d.__class__.__name__, "size": d.size} for d in target.disks],
+        "volumes": [{"name": v.name, "size": v.size, "fs": v.fs.__class__.__name__} for v in target.volumes],
```
Schamper, line 111: Filesystems optionally have `fs.__fstype__` that you can use, which will return a slug. E.g. `ntfs`.

dissect/target/tools/info.py
```diff
@@ -0,0 +1,145 @@
...
+        for c in children:
+            print(f"- <Child type='{c.type}' path='{str(c.path)}'>")
+
+    print()
+    print(f"Hostname      : {target.hostname}")
+    print(f"OS family     : {target.os}, {target._os_plugin.__name__}")
+    print(f"OS version    : {target.version}")
+    print(f"Arch          : {target.architecture}")
+    print(f"Domain        : {target.domain}")
+    print(f"IPs           : {', '.join(list(set(target.ips + target.ips_dhcp)))}")
```
Schamper, line 137:
```suggestion
    print(f"IPs           : {', '.join(set(target.ips + target.ips_dhcp))}")
```
Making a list is necessary again for the other methods but not here.

Schamper (changes requested):

dissect/target/tools/info.py
JSCU-CNI: For API consistency one could argue Windows targets should also have a `ips_dhcp` function. Currently `ips` for Windows targets will also return DHCP IPs. 

Do you prefer checking if `ips_dhcp` exists for the current target or adding `ips_dhcp` to `plugins/os/windows/_os.py`?

dissect/target/tools/info.py
JSCU-CNI: Good to know. Unfortunately for partitions/volumes with an unidentified filesystem (eg. `Microsoft reserved partition`), this wouldn't work as the returned class will be `NoneType`.

Schamper: Oops, good point. Then this is indeed the solution with the most straightfoward code.

dissect/target/tools/info.py
Schamper: I agree. Technically everything here that isn't defined in `OSPlugin` is not guaranteed to exist, but we can start formalizing most of those soon.

I'm very torn on `ips_dhcp`. Can you give an indication on the time it takes to parse a decently sized syslog? If it's not very long, maybe I'm going back on my initial decision to separate it out, eat the performance hit now, and properly fix it later with more elaborate network interface code.

JSCU-CNI: Did some small tests with uncompressed logs ranging from 1MB to 10MB. The time it takes to parse a syslog (in seconds) seems directly proportional to the log size (in MBs). Meaning that 1MB of syslog equals to roughly 1 second of parsing time in Dissect.

It's difficult to say how large on average syslogs are on Unix systems, as the size is heavily dependent on the specifics of what the machine is used for. Depending on configuration, `logrotated` is usually configured to rotate logs every day or week, regardless of size, and keep logs for four weeks.

Schamper: Had a short internal discussion, let's eat the performance hit now and fix it properly later with more elaborate network interface code. That means reverting the change in #133 and putting the DHCP logic back into the `ips` function.

JSCU-CNI: Implemented in [4584d9b](https://github.com/fox-it/dissect.target/pull/136/commits/4584d9bdad96a4b66c0c1dbcc142d0d0b4281199) and [2c3216e](https://github.com/fox-it/dissect.target/pull/133/commits/2c3216ea1aec73a55875698ec84162709ff0a994).

dissect/target/tools/info.py
```diff
@@ -0,0 +1,142 @@
...
+    print(f"Hostname      : {target.hostname}")
+    print(f"OS family     : {target.os}, {target._os_plugin.__name__}")
+    print(f"OS version    : {target.version}")
+    print(f"Arch          : {target.architecture}")
+    print(f"Domain        : {target.domain}")
+    print(f"IPs           : {', '.join(target.ips)}")
+    print(f"Install date  : {target.install_date}")
+    print(f"Last activity : {target.activity}")
+    print(f"Language(s)   : {', '.join(target.language)}")
+    print(f"Timezone      : {target.timezone}")
```
Schamper, line 138:
```suggestion
    print(f"Hostname      : {target.hostname}")
    print(f"Domain        : {target.domain}")
    print(f"IPs           : {', '.join(target.ips)}")
    print(f"OS family     : {target.os}, {target._os_plugin.__name__}")
    print(f"OS version    : {target.version}")
    print(f"Architecture  : {target.architecture}")
    print(f"Language(s)   : {', '.join(target.language)}")
    print(f"Timezone      : {target.timezone}")
    print(f"Install date  : {target.install_date}")
    print(f"Last activity : {target.activity}")
```

Schamper (changes requested):

dissect/target/tools/info.py
```diff
@@ -0,0 +1,142 @@
...
+    "target/info",
+    [
+        ("datetime", "last_activity"),
+        ("datetime", "install_date"),
+        ("net.ipaddress[]", "ips"),
+        ("string", "os_family"),
+        ("string", "os_version"),
+        ("string", "architecture"),
+        ("string", "language"),
+        ("string", "timezone"),
```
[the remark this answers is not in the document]
JSCU-CNI: We prefer to have `datetime` fields first, just like regular records have `ts` first for processing reasons.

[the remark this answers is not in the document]
Schamper: That's fine, but can you then apply the same consistent ordering to the other places?

dissect/target/tools/info.py
```diff
@@ -0,0 +1,140 @@
...
+                    os_version=target.version,
+                    architecture=target.architecture,
+                    language=target.language,
+                    timezone=target.timezone,
+                    _target=target,
+                )
+            )
+        else:
+            print_target_info(target)
+
```
Schamper, line 92: Maybe put this code into a try/except with an error, such as:
```python
target.log.error("Exception in retrieving information for target: `%s`", target, exc_info=True)
```

Schamper (changes requested):

dissect/target/tools/info.py
JSCU-CNI: Added in [94af2a6](https://github.com/fox-it/dissect.target/pull/136/commits/94af2a62f1b38dd7c6046bdb27d7a94ccc82aa00).

dissect/target/tools/info.py
```diff
@@ -0,0 +1,142 @@
...
+        "architecture": target.architecture,
+        "domain": target.domain,
+        "ips": target.ips,
+        "install_date": str(target.install_date),
+        "last_activity": str(target.activity),
+        "language": target.language,
+        "timezone": target.timezone,
+        "disks": [{"type": d.__class__.__name__, "size": d.size} for d in target.disks],
+        "volumes": [{"name": v.name, "size": v.size, "fs": v.fs.__class__.__name__} for v in target.volumes],
+        "children": [{"type": c.type, "path": str(c.path)} for c in target.list_children()],
```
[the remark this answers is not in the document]
JSCU-CNI: Added in [8b807dc](https://github.com/fox-it/dissect.target/pull/136/commits/8b807dc0e34a408da4ca48555cc8ad1260598e23).

JSCU-CNI:
Thanks, looks good to me :)

Schamper:
Made a few additional tweaks, in addition to a change in the Unix locale plugin. I had a VM where the language value was surrounded in quotes.

dissect/target/tools/info.py
```diff
@@ -50,8 +50,8 @@ def main():
     parser.add_argument("-d", "--delimiter", default=" ", action="store", metavar="','")
     parser.add_argument("-s", "--strings", action="store_true", help="print output as string")
     parser.add_argument("-r", "--record", action="store_true", help="print output as record")
-    parser.add_argument("-j", "--json-pretty", action="store_true", help="output records as pretty json")
-    parser.add_argument("-J", "--json", action="store_true", help="output records as one-line json")
+    parser.add_argument("-j", "--json", action="store_true", help="output records as pretty json")
+    parser.add_argument("-J", "--jsonlines", action="store_true", help="output records as one-line json")
```
JSCU-CNI, line 54: @Schamper This change breaks `target-info`. At lines 75 and 77 `args.json` needs to be replaced with `args.jsonlines` and `args.json_pretty` needs to be replaced with `args.json`.

Schamper: You are correct, too hastily merged :) Fixed with #186

PR #143, opened 2023-01-25
Improve target-fs command line error handling

pyrco:
(DIS-1708)

### DIFF
--- a/dissect/target/tools/fs.py
+++ b/dissect/target/tools/fs.py
@@ -5,6 +5,7 @@ import argparse
 import logging
 import operator
 import os
+import pathlib
 import shutil
 import sys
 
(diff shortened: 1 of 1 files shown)


dissect/target/tools/fs.py
```diff
@@ -106,16 +107,22 @@ def main():
 
     args = parser.parse_args()
 
+    if args.subcommand is None:
+        parser.error("No subcommand specified")
+
+    if not args.target.is_file():
+        parser.error(f"{args.target} is not a file or does not exist")
```
Poeloe, line 114: As a suggestion to add a little more context:
```suggestion
        parser.error(f"Specified target '{args.target}' is not a file or does not exist")
```

Issue #135, opened 2023-01-19
TarLoader issues


JSCU-CNI:
We've come across two issues regarding the use of tar-files:

1. With the merge of #112, the class `TarFile` is removed. This however [breaks the AsdfLoader](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/asdf.py#L5).
2. In the TarLoader, [directories are skipped](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/tar.py#L37). It seems that as result of this, empty folders are not included in for example `target.walkfs()`, although they might be of value from a forensic point of view. The [current test](https://github.com/fox-it/dissect.target/blob/main/tests/test_loaders_tar.py) does not include the occurrence of empty folders within a tar-file.

Schamper:
I believe both of these issues are currently resolved. Can you confirm and close the issue if this is the case?

JSCU-CNI:
The latter issue has indeed been fixed with #144 and #195. The first issue however still seems to be present. The `AsdfLoader` still tries to import `TarFile` from `dissect.target.loaders.tar` in [`loaders/asdf.py`](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/asdf.py#L6).

PR #144, opened 2023-01-26
Fix tar filesystem

JSCU-CNI:
This PR attempts to fix issues with the tar filesystem and loader (#135 and #117).


Schamper:
As mentioned in #117, can you implement similar changes here?

JSCU-CNI:
Could you be a bit more specific? There's quite a lot we talked about in #117. Added typing and `alt_separator` implementation in [896981d](https://github.com/fox-it/dissect.target/pull/144/commits/896981daadf8889c036651a4ce601ced3f6399d5).

Schamper:
Added some changes and fixes, going to again request a review from a colleague for this one.

dissect/target/filesystems/tar.py
```diff
@@ -1,72 +1,90 @@
...
-    def _resolve(self):
+    def _resolve(self) -> TarFilesystemEntry:
         if self.is_symlink():
             return self.readlink_ext()
         return self
 
-    def open(self):
+    def open(self) -> BinaryIO:
         """Returns file handle (file-like object)."""
+        if self.is_dir():
```
pyrco, line 85: With the use of `TarFilesystemDirectoryEntry` this directory specific handling here and elsewhere is no longer needed right? As directories in a `TarFilesystem` are either mapped as a `VirtualDirectory` (implicitly) or `TarFilesystemDirectoryEntry` explicitly?

pyrco (changes requested):

dissect/target/filesystems/tar.py
Schamper: It does, since `TarFilesystem` supports symlinks and `self.is_dir()` might dereference to a `TarFilesystemDirectoryEntry`.

pyrco (approved):

PR #146, opened 2023-01-28
Add  folder as package

### LINKED ISSUE #145: Wireguard plugin is not loaded
The folder `dissect/target/plugins/apps/vpns` does not contain the empty file `__init__.py`, because of this the [wireguard](https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/apps/vpns/wireguard.py) plugin is not loaded.

```bash
target-query -l | grep wireguard
```

Zawadidone:
Closes #145

```bash
target-query -l
Available plugins:
[...]
      vpns:
        wireguard:
          wireguard.config - Parses interface config files from wireguard installations. (output: records)
[...]
```


Schamper:
Nice find! Interesting that the tests work though.

Schamper (approved):

PR #152, opened 2023-01-30
Defender add `DefenderBehaviorQuarantineRecord` to export

Zawadidone:


Schamper (approved):

Schamper:
Good catch, yes it also looks like the same applies to the `thumbcache` plugin. Can you create a separate PR for that one?

Zawadidone:
Yess I will

PR #155, opened 2023-02-01
Fix incorrect MappedFile behavior

Schamper:
(DIS-1793)


pyrco:
Is it an idea to implement `attr()`/`lattr()` for mapped files through `os.listxattr()`?

pyrco (approved):

Schamper:
> Is it an idea to implement `attr()`/`lattr()` for mapped files through `os.listxattr()`?

Yeah I was also thinking about adding symlink support, but I think both of those are best reserved for other tickets for the time being.

PR #166, opened 2023-02-09
Fix TarFilesystem regression for relative lookups

Schamper:
(DIS-1815)

### DIFF
--- a/dissect/target/filesystems/tar.py
+++ b/dissect/target/filesystems/tar.py
@@ -44,7 +44,6 @@ class TarFilesystem(Filesystem):
         self.base = base or ""
 
         self._fs = VirtualFilesystem(alt_separator=self.alt_separator, case_sensitive=self.case_sensitive)
-        self._fs.tar = self.tar
 
         for member in self.tar.getmembers():
             mname = member.name.strip("/")
--- a/dissect/target/filesystems/tar.py
+++ b/dissect/target/filesystems/tar.py
@@ -54,11 +54,11 @@ class TarFilesystem(Filesystem):
             rel_name = fsutil.normpath(mname[len(self.base) :], alt_separator=self.alt_separator)
 
             entry_cls = TarFilesystemDirectoryEntry if member.isdir() else TarFilesystemEntry
-            file_entry = entry_cls(self._fs, rel_name, member)
+            file_entry = entry_cls(self, rel_name, member)
             self._fs.map_file_entry(rel_name, file_entry)
 
     @staticmethod
-    def detect(fh) -> bool:
+    def detect(fh: BinaryIO) -> bool:
         """Detect a tar file on a given file-like object."""
         offset = fh.tell()
         try:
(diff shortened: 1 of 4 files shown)


dissect/target/filesystems/tar.py
```diff
@@ -156,7 +156,7 @@ def lstat(self) -> fsutil.stat_result:
 
 
 class TarFilesystemDirectoryEntry(VirtualDirectory):
-    def __init__(self, fs: TarFilesystem, path: str, entry: tarfile.TarInfo):
+    def __init__(self, fs: VirtualFilesystem, path: str, entry: tarfile.TarInfo):
```
pyrco, line 159: This change seems incorrect, as these entries are only created by instances of `TarFilesystem` and it will pass `self` into `fs`.

dissect/target/filesystems/zip.py
```diff
@@ -71,9 +71,9 @@ def detect(fh: BinaryIO) -> bool:
         finally:
             fh.seek(offset)
 
-    def get(self, path: str) -> ZipFilesystemEntry:
+    def get(self, path: str, *args, **kwargs) -> ZipFilesystemEntry:
```
pyrco, line 74: I think it is nicer to pass the variables explicitly like:
```suggestion
    def get(self, path: str, relentry: FilesystemEntry = None) -> ZipFilesystemEntry:
```

dissect/target/filesystems/tar.py
```diff
@@ -69,9 +69,9 @@ def detect(fh) -> bool:
         finally:
             fh.seek(offset)
 
-    def get(self, path) -> TarFilesystemEntry:
+    def get(self, path: str, *args, **kwargs) -> TarFilesystemEntry:
```
pyrco, line 72: I think it is nicer to pass the variables explicitly like:
```suggestion
    def get(self, path: str, relentry: FilesystemEntry = None) -> TarFilesystemEntry:
```

pyrco (changes requested):

dissect/target/filesystems/tar.py
Schamper: Yeah sorry this was from a different fix approach.

dissect/target/filesystems/tar.py
```diff
@@ -69,9 +70,9 @@ def detect(fh) -> bool:
         finally:
             fh.seek(offset)
 
-    def get(self, path) -> TarFilesystemEntry:
+    def get(self, path: str, relentry: FilesystemEntry = None) -> TarFilesystemEntry:
         """Returns a TarFilesystemEntry object corresponding to the given path."""
-        return self._fs.get(path)
+        return self._fs.get(path, relentry)
```
pyrco, line 75:
```suggestion
        return self._fs.get(path, relentry=relentry)
```
Otherwise we may get in trouble if the function signature changes.

dissect/target/filesystems/zip.py
```diff
@@ -71,9 +72,9 @@ def detect(fh: BinaryIO) -> bool:
         finally:
             fh.seek(offset)
 
-    def get(self, path: str) -> ZipFilesystemEntry:
+    def get(self, path: str, relentry: FilesystemEntry = None) -> ZipFilesystemEntry:
         """Returns a ZipFilesystemEntry object corresponding to the given path."""
-        return self._fs.get(path)
+        return self._fs.get(path, relentry)
```
pyrco, line 77:
```suggestion
        return self._fs.get(path, relentry=relentry)
```
Otherwise we may get in trouble if the function signature changes.

pyrco (changes requested):

pyrco (approved):

PR #168, opened 2023-02-10
Clean up filesystem detection and add type hints

Schamper:
(DIS-1816)

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -5,7 +5,17 @@ import logging
 import os
 import stat
 from collections import defaultdict
-from typing import Any, BinaryIO, Callable, Iterator, List, Optional, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    BinaryIO,
+    Callable,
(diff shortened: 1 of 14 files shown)


dissect/target/filesystem.py
```diff
@@ -58,26 +71,54 @@ def path(self, *args) -> fsutil.TargetPath:
...
+
+        Args:
+            fh: A file-like object, usually a disk or partition.
+
+        Returns:
+            ``True`` if ``fh`` is supported, ``False`` otherwise.
+        """
+        offset = fh.tell()
+        try:
+            fh.seek(0)
```
pyrco, line 88: I noticed not all filesystem `detect()` implementations originally did this. Doing it anyway won't be a problem?

Schamper: Nope, it never really made sense that it didn't do it, because you could never "open" a filesystem from arbitrary locations anyway. The intended way to do that would be to use a `RelativeStream` or `RangeStream`.

pyrco (approved):

PR #171, opened 2023-02-10
Add open handles plugin

Poeloe:
(DIS-1667)


pyrco:
> Is it an idea to generate an openhandles.csv.gz for test purposes?

I think it is a good idea to have some unit tests indeed.

PR #173, opened 2023-02-14
improve unix audit plugin

JSCU-CNI:
This PR improves /var/log/audit/audit.log parsing on RedHat and CentOS systems.


dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,82 @@
...
     [
         ("datetime", "ts"),
-        ("string", "type"),
-        ("string", "id"),
-        ("string", "msg"),
-        ("string", "auditlog"),
+        ("string", "audit_type"),
+        ("varint", "audit_id"),
+        ("string", "message"),
+        ("uri", "source"),
```
Schamper, line 18:
```suggestion
        ("path", "source"),
```

dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,82 @@
...
+        paths = ["/var/log/audit/audit.log"]
+
+        if (path := self.target.fs.path(default_config)).exists():
+            for line in path.open("rt"):
+                line = line.strip()
+                if not line:
+                    continue
+                if "log_file" in line:
+                    log_file = line.split("=")[-1].strip()
+                    paths.append(log_file)
```
Schamper, line 42:
```suggestion
                if not line or line.startswith("#") or "log_file" not in line:
                    continue

                log_file = line.split("=")[-1].strip()
                if log_file not in paths:
                    paths.append(log_file)
```
Assuming that this config file uses `#` as comments.

dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,82 @@
...
 
-        if path.exists():
-            return True
-        else:
-            return False
+        return paths
+
+    def check_compatible(self) -> bool:
+        return self.target.fs.path("/var/log/audit/").exists()
```
Schamper, line 47:
```suggestion
        return len(self.log_paths) > 0
```

dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,82 @@
...
-                    ts=datetime.datetime.fromtimestamp(float(ts)),
-                    type=_type,
-                    id=int(_id.strip(")")),
-                    msg=msg,
-                    auditlog=None,
-                    _target=self.target,
-                )
+
+        for path in self.log_paths:
+            print(path)
```
Schamper, line 65: Remove prints.

dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,82 @@
...
+            for file in self.target.fs.path(path).parent.glob("*"):
```
Schamper, line 66: Is this glob not a bit too generic? What if it's configured to log to `/var/log/audit.log`, then this would read all log files in `/var/log`.

Schamper (changes requested):

tests/test_plugins_os_unix_log.py
```diff
@@ -97,3 +100,28 @@ def test_atop_plugin(target_unix, fs_unix):
     assert bool(results[0].wasinactive) is False
     assert results[0].container == ""
     assert str(results[0].filepath) == "atop_20221111"
+
+
+def test_audit_plugin(target_unix, fs_unix):
```
Schamper, line 105: Can you also move this test into a separate `test_plugins_os_unix_log_audit.py`?

I actually want the other tests in here also into separate files but we'll fix that later.

dissect/target/plugins/os/unix/log/audit.py
JSCU-CNI: Filtering out comments would mean we would miss out on possible previously used log file locations, eg:

```
# log_file = /previously/used/audit.log
log_file = /foo/bar/audit.log
```

dissect/target/plugins/os/unix/log/audit.py
JSCU-CNI: Improved in [f10e87a](https://github.com/fox-it/dissect.target/pull/173/commits/f10e87adc350cba6a0c720fe32bdc9b1bd16bfd4):

```
self.target.fs.path(path).parent.glob(basename(path) + "*")
```

dissect/target/plugins/os/unix/log/audit.py
Schamper: Fair point, would you suggest the same applies to #127?

JSCU-CNI: Good catch, added in #127.

Schamper:
Made some additional changes.

dissect/target/plugins/os/unix/log/audit.py
```diff
@@ -1,60 +1,89 @@
...
+    def get_log_paths(self) -> list[Path]:
+        log_paths = []
+
+        log_paths.extend(self.target.fs.path("/var/log/audit").glob("audit.log*"))
 
-        if path.exists():
-            return True
-        else:
-            return False
+        default_config = "/etc/audit/auditd.conf"
```
Schamper, line 39: Can you add a unit test for parsing the config btw?

Schamper (changes requested):

dissect/target/plugins/os/unix/log/audit.py
JSCU-CNI: Added in [ea42c88](https://github.com/fox-it/dissect.target/pull/173/commits/ea42c884e249edbb360188c4ee0f13e0da58b82b).

Schamper (approved):

PR #180, opened 2023-02-21
Print errors in target-shell when running with -q

pyrco:
(DIS-1646)

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -501,31 +501,26 @@ class TargetCli(TargetCmd):
             if len(contents) > 1:
                 print(f"total {len(contents)}", file=stdout)
             for target_path, name in contents:
-                try:
-                    self.print_extensive_file_stat(stdout=stdout, target_path=target_path, name=name)
-                except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


dissect/target/tools/shell.py
```diff
@@ -1069,6 +1067,12 @@ def main():
 
     process_generic_arguments(args)
 
+    # For the shell tool we want -q to log slightly more then just CRITICAL
+    # messages.
+    dissect_logger = logging.getLogger("dissect")
+    if args.quiet:
+        dissect_logger.setLevel(level=logging.ERROR)
```
Schamper, line 1074:
```suggestion
    if args.quiet:
        logging.getLogger("dissect").setLevel(level=logging.ERROR)
```
I believe this could just be one line in the if block?

Schamper, line 1074:
```suggestion
    if args.quiet:
        logging.getLogger("dissect").setLevel(level=logging.ERROR)
```
I believe this could just be one line in the if block?

Schamper (approved):

Schamper (approved):

PR #181, opened 2023-02-21
Fix displaying Windows alternate data streams in target-shell

pyrco:
(DIS-1844)

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -1222,22 +1222,15 @@ class EntryList(list):
     there's a virtual filesystem entry as well as a real one.
     """
 
-    def __init__(self, value: Any, subattribute: Optional[str] = None):
+    def __init__(self, value: Any):
         if not isinstance(value, list):
             value = [value]
-        self._sub = subattribute
         super().__init__(value)
 
     def __getattr__(self, attr: str) -> Any:
-        for e in self:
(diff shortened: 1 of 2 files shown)


pyrco:
@Schamper I couldn't find a more elegant way to determine whether the file has Windows attributes. Maybe you know of a way?

Schamper:
Hmm yeah normally you'd do something like `entry.fs.__fstype__ == "ntfs"`. That should normally work on `RootFilesystemEntry` objects using `rootfs_entry.entries.fs.__fstype__`, however I believe that `EntryList` (https://github.com/fox-it/dissect.target/blob/main/dissect/target/filesystem.py#L1218) is buggy atm which is why it doesn't work.

I would prefer a approach that doesn't require importing from `dissect.ntfs`, because this creates a hard requirement on `dissect.ntfs` for `target-shell`.

pyrco:
Looks like the `_sub` property of `EntryList` was the culprit. I removed it completely as there was no immediate use for it.

Schamper:
Ahh yes I remember now, that sub attribute was initially meant to be a shortcut to the underlying `FilesystemEntry.entry` attribute, i.e. the implementation specific entry object. However I don't think it ever actually ended up being used as such.

Schamper (approved):

PR #183, opened 2023-02-21
Add support for extended zsh history files

JSCU-CNI:
This PR adds support for extended zsh_history files with timestamps.


dissect/target/plugins/os/unix/history.py
```diff
@@ -48,41 +53,80 @@ def commandhistory(self):
...
-                        yield CommandHistoryRecord(
-                            ts=cmd_ts,
-                            command=line,
-                            source=str(file_),
-                            _target=self.target,
-                            _user=user_details.user,
-                        )
-                except Exception:
-                    self.target.log.exception("Failed to parse command history: %s", file_)
+                # TODO: Starting with Python 3.10 we can use pattern matching (PEP 634)
```
pyrco, line 60: We generally tend not to leave `TODO`'s in the code, we either do them or create a ticket for them. In this case I think it can be replaaced with `NOTE`.

```suggestion
                # NOTE: Starting with Python 3.10 we can use pattern matching (PEP 634)
```

pyrco (changes requested):
Thanks for the nice addition! Just one small thing.

pyrco (approved):

Issue #184, opened 2023-02-22
Velociraptor Loader Volume Shadow Copies


Zawadidone:
When using the Velociraptor artefact KapeFiles.Targets with [VSSAnalysis](https://github.com/Velocidex/velociraptor/blob/80aeffb8edff94be9ad0fae6a7dc3341fecf5e20/artifacts/definitions/Windows/KapeFiles/Targets.yaml#L33), the collection contains additional volumes as shown below:

```bash
ls Collection-HOSTNAME-TIMESTAMP/uploads/mft/

%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1  
%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy2  
%5C%5C.%5CC%3A
```

Because of this the wrong folder could be used in the [Velociraptor Loader](https://github.com/fox-it/dissect.target/blob/55aa54992cf5d69a70ec6a9cdadc59b411b8dc59/dissect/target/loaders/velociraptor.py#L27).

The current fix is to only map the C volume. But it would be nice if data in the VSS folders could be parsed additionally.

## Reference

[https://github.com/Velocidex/velociraptor/blob/80aeffb8edff94be9ad0fae6a7dc3341fecf5e20/artifacts/definitions/Windows/Collectors/VSS.yaml](https://github.com/Velocidex/velociraptor/blob/80aeffb8edff94be9ad0fae6a7dc3341fecf5e20/artifacts/definitions/Windows/Collectors/VSS.yaml)

Zawadidone:
I wil create a PR for the current fix.

Is it supported by Dissect Target to parse data from folders that don't have a [Windows folder layout](https://github.com/fox-it/dissect.target/blob/55aa54992cf5d69a70ec6a9cdadc59b411b8dc59/dissect/target/loaders/dir.py#L83)?

```bash
%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1/:
'$Extend' [...]

%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy139/:
'$Boot'  '$Extend'  '$LogFile'  '$MFT' [...]
```

Schamper:
[...]

PR #185, opened 2023-02-22
Fix velociraptor loader c folder

Zawadidone:
#184 hardcode the `C.` folder


Zawadidone:
Are the made adjustments okay?

Schamper:
I've been thinking about alternative solutions to this with the discussion of #184 in mind and I think I found a decent alternative.

I can make a code suggestion to your branch if you'd like, but basically:
- Rename `map_dirs` in `dir.py` to `find_and_map_dirs`
- Move the loop (from `for path in dirs:` onwards) to a new function `map_dirs(target: Target, dirs: list[Path], alt_separator: str, case_sensitive: bool, **kwargs)`
- Use this new `map_dirs` to map all the volumes correctly in the velociraptor loader, including the volume shadow copies

I think this should work with your intended behavior, as well as making those functions a bit more flexible.

Let me know if you want me to make these changes or if you want to make them yourself.

Zawadidone:
Yes can you make this change.

Schamper:
I pushed my proposed changes, let me know if this approach works for you.

Zawadidone:
Thanks, yes this approach works and nice way of using a list of directories instead of the hardcoded string `C.`.

PR #188, opened 2023-02-27
DIS-1590 Targetd Loader/Plugin

cecinestpasunepipe:
Loader for targetd-targets (PoC)


dissect/target/tools/utils.py
```diff
@@ -124,6 +125,17 @@ def generate_argparse_for_plugin(
     return generate_argparse_for_plugin_class(plugin_instance.__class__, usage_tmpl=usage_tmpl)
 
 
+def plugin_factory(target: Target, plugin_class: type, funcname: str) -> tuple[Plugin, str]:
+    if TargetdLoader.instance:
```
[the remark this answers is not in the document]
cecinestpasunepipe: Isn't this simpler?

dissect/target/loaders/targetd.py
```diff
@@ -0,0 +1,135 @@
...
+            self.client.module_fromlist = ["command_runner"]
+            self.client.command = self._plugin_func
+            self.peers = peers
+            self.client.start()
+
+        while not self.output:
+            time.sleep(1)
+        return self.output
+
+    def get_command(self, func):
```
pyrco, line 114:
```suggestion
    def _get_function(self, func):
```
?
As it isn't really the idea to use this standalone and you don't really get the command?

dissect/target/loaders/targetd.py
```diff
@@ -0,0 +1,135 @@
...
+        """For target API"""
+        self._plugin_func = func
```
pyrco, line 116: It feels a bit icky that this state separate form the `plugin_bridge` that is returned.
Would it be an idea to curry the `self.plugin_bridge` instead using `functools.partial()`?

pyrco (changes requested):

dissect/target/loaders/targetd.py
pyrco: Hmm, that is a shame. I noticed there is also a `@functools.wraps`/`functools.update_wrapper()`, which turns a wrapper function into something that looks like the original. Maybe that can help?

pyrco: Probably even simpler:

```
curried_plugin_bridge = functools.update_wrapper(functools.partial(self.plugin_bridge, plugin_func=func), self.plugin_bridge)
```

pyrco (commented):
See comments

dissect/target/loaders/targetd.py
cecinestpasunepipe: Interesting, giving this a try now...

cecinestpasunepipe: Okay this seems to work fine, but I still don't see how we can integrate this with the other 'static' part used for target-query https://github.com/fox-it/dissect.target/pull/188/files#diff-906948d654bf75fd862595a8e8df77d945b385ad7e2a898c8214e53b983a381dR128 ?

cecinestpasunepipe: Fixed

pyrco (approved):

PR #194, opened 2023-03-04
Fix namespace non findable plugins

Zawadidone:
When using the following command to list functions that output records, the output contains the namespace plugins `remoteaccess` and `example`. Which from a use case to execute all plugins with `xargs` is not intended due to duplicate records of `remoteaccess` and executing the `example` plugin.

```bash
 target-query -l | grep "output: records"  | grep -E "remoteaccess|example"
          anydesk.remoteaccess - Return the content of the AnyDesk logs. (output: records)
          remoteaccess.remoteaccess - Return Remote Access records from all Remote Access Tools. (output: records)
          teamviewer.remoteaccess - Return the content of the TeamViewer logs. (output: records)
        example_record - Example plugin that generates records. (output: records)
        example_user_registry_record - Example plugin that generates records with registry key and user information. (output: records)
```

This PR makes these plugins non findable like in the [browser plugin](https://github.com/fox-it/dissect.target/blob/dda548a6c08f8c69d02e62459aad14b96e0474f2/dissect/target/plugins/browsers/browser.py#L38).

```bash
 target-query -l | grep "output: records"  | grep -E "remoteaccess|example"
          anydesk.remoteaccess - Return the content of the AnyDesk logs. (output: records)
          teamviewer.remoteaccess - Return the content of the TeamViewer logs. (output: records)
```

If this is intended is there a way that a plugin that groups function is by default non findable?


cecinestpasunepipe:
Looks good, nice improvement. Currently there is no mechanism to automatically mark groups as "unfindable" - not sure if this should always be the case as well.

Schamper:
> Looks good, nice improvement. Currently there is no mechanism to automatically mark groups as "unfindable" - not sure if this should always be the case as well.

There's an (internal atm) ticket to create a boilerplate namespace class, which should address this.

Schamper (approved):

PR #195, opened 2023-03-07
Fix tar loader for targets with '.' folder

JSCU-CNI:
Tarballs can have folders named `.` (eg. `tar -czf .`). Currently this breaks the `walkfs` plugin. 

Stripping leading dots from `entry.path` in the tar loader's volume file entry mapper fixes this.


Schamper:
Proposed an alternative fix, more in line with how `TarFilesystem` currently handles this.

Schamper (approved):

PR #196, opened 2023-03-08
Improve target-shell

JSCU-CNI:
This PR adds two features:

- add support for compressed files: use `fsutil.open_decompress` function for `cat` and `less` commands.
- enforce regular shell behaviour by defining an empty `emptyline` function in `TargetCmd`

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -716,7 +716,7 @@ class TargetCli(TargetCmd):
             if not path:
                 continue
 
-            fh = path.open()
+            fh = fsutil.open_decompress(path)
             shutil.copyfileobj(fh, stdout)
             stdout.flush()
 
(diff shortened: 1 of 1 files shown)


pyrco:
Thanks for the useful improvements! After some internal discussion we think it would be better to implement the `cat`/`less` seamless decompression of files in separate `zcat`/`zless` functions.

The idea is that we would like to still be able to cat the raw bytes of a file with a `.gz` or `.bz2` extension to other processes' standard input. This would also open up the possibility to eventually decompress files irrespective of file extension by determining the compression algorithm based on the file magic.

pyrco (changes requested):
Thanks for the useful improvements! After some internal discussion we think it would be better to implement the `cat`/`less` seamless decompression of files in separate `zcat`/`zless` functions.

The idea is that we would like to still be able to cat the raw bytes of a file with a `.gz` or `.bz2` extension to other processes' standard input. This would also open up the possibility to eventually decompress files irrespective of file extension by determining the compression algorithm based on the file magic.

JSCU-CNI:
That seems reasonable. Implemented `zcat` and `zless` in [86acb9e](https://github.com/fox-it/dissect.target/pull/196/commits/86acb9ecf98afd26b624ef6320b9a38362073e29).

PR #200, opened 2023-03-08
Add McAfee SQLite log parser (DIS-1523)

cecinestpasunepipe:
Adds support for SQLite3 history file from McAfee

(DIS-1523)


tests/test_plugins_apps_av_mcafee.py
```diff
@@ -0,0 +1,16 @@
+from dissect.target import Target
+from dissect.target.plugins.apps.av.mcafee import McAfeeLogRecord
+
+from ._utils import absolute_path
+
+
+def test_mcafee_plugin_log():
+    for record in Target.open(absolute_path("data/apps/mcafee/mcafee.tar.gz")).mcafee.logs():
```
Schamper, line 8:
```suggestion
    for record in Target.open(absolute_path("data/apps/mcafee/mcafee.tar.gz")).mcafee.logs():
```
Can you move the test file to the `data/apps/av/mcafee` directory?

cecinestpasunepipe: Moved to av-dir.

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,116 @@
...
+            yield from self.target.fs.path(path).glob(self.LOG_FILE_PATTERN)
+
+    def _clean_message(self, message: str) -> str:
+        return re.sub("<[^!][^>]*>", "", (" ".join(re.findall(r"<!\[CDATA\[(.*?)\]\]>", message, flags=re.M))))
+
+    @export(record=McAfeeMscLogRecord)
+    def msc(self) -> Iterator[McAfeeMscLogRecord]:
+        """Return msc log history records from McAfee.
+
+        Yields McAfeeLogRecord with the following fields:
```
Schamper, line 57: Can you update the fields here too?

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,116 @@
...
+    def check_compatible(self) -> bool:
+        if not self.get_log_files():
+            raise UnsupportedPluginError("No McAfee Log files found")
+
+    def get_log_files(self) -> Iterator[str]:
+        for path in self.DIRS:
+            yield from self.target.fs.path(path).glob(self.LOG_FILE_PATTERN)
+
+    def _clean_message(self, message: str) -> str:
+        return re.sub("<[^!][^>]*>", "", (" ".join(re.findall(r"<!\[CDATA\[(.*?)\]\]>", message, flags=re.M))))
```
Schamper, line 51: Can you compile these patterns?

tests/test_plugins_apps_av_mcafee.py
```diff
@@ -0,0 +1,16 @@
+from dissect.target import Target
+from dissect.target.plugins.apps.av.mcafee import McAfeeMscLogRecord
+
+from ._utils import absolute_path
+
+
+def test_mcafee_plugin_log():
+    for record in Target.open(absolute_path("data/apps/av/mcafee/mcafee.tar.gz")).mcafee.msc():
```
Schamper, line 8: I haven't checked the content of the archive, but would it not be cleaner to only use the database file as test files?

Schamper (changes requested):

tests/test_plugins_apps_av_mcafee.py
cecinestpasunepipe: I recommend this approach because:

1. You can use the same test file from the CLI with the TAR-loader (for demo purposes or manual testing)
2. This also tests the path detection mechanisms

dissect/target/plugins/apps/av/mcafee.py
cecinestpasunepipe: ok

tests/test_plugins_apps_av_mcafee.py
cecinestpasunepipe: Discussed with team, for now this seems to be a better approach indeed. However an additional benefit of this version was that you could use the same test file for a manual test or demo on the CLI.

dissect/target/plugins/apps/av/mcafee.py
cecinestpasunepipe: Sure, done.

Schamper (changes requested):

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,119 @@
...
+                            protocol = "TCP"
+                            continue
+                        if log_field.startswith(self.MARKER_SUSPICIOUS_UDP_CONNECTION):
+                            port = int(log_field[len_udpstr:])
+                            protocol = "UDP"
+                            continue
+                        if key == self.TEMPLATE_ID_INFECTION and entry.details_info.find(self.MARKER_INFECTION) > -1:
+                            threat = log_field
+                    yield McAfeeMscLogRecord(
+                        ts=entry.date,
```
Schamper, line 111: Can you check if this has a timezone attached? If not, maybe best to fix this directly in `dissect.sql`?

Schamper (changes requested):

dissect/target/plugins/apps/av/mcafee.py
cecinestpasunepipe: No timezone attached, stored in SQLite as a unix timestamp, so also no timezone needed. I think they just use the timezone of the host upon display.

image

Schamper: Unix timestamps are always UTC, so a `tz=timezone.utc` needs to be added to create aware datetime objects. Also see https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

cecinestpasunepipe: I see that flow.record converts this to utc if passed as int so that should give us a uniform datetime record even if the timezones on the hosts may vary, this is quite optimal right?

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,119 @@
...
+    MARKER_SUSPICIOUS_TCP_CONNECTION = "TCP port "
+    MARKER_SUSPICIOUS_UDP_CONNECTION = "UDP port "
+    TABLE_LOG = "log"
+    TABLE_FIELD = "field"
+
+    def check_compatible(self) -> bool:
+        if not self.get_log_files():
+            raise UnsupportedPluginError("No McAfee Log files found")
+
+    def get_log_files(self) -> Iterator[str]:
```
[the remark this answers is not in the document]
cecinestpasunepipe: Ok, done.

dissect/target/plugins/apps/av/mcafee.py
Schamper: `flow.record` does not currently create aware datetime objects, this is a known issue and being considered (it has a significant performance hit).

My question was if this returned a datetime object or an integer. I didn't remember if SQLite has a dedicated datetime type but it turns out it does not. So since this returns an integer, you must add a `ts.from_unix(entry.date)` to this line.

cecinestpasunepipe: Ah I see because the flow.record constructor uses utcfromtimestamp which leaves a naive dt object

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,120 @@
...
+
+from dissect.sql import SQLite3
+from dissect.util.ts import from_unix
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import TargetRecordDescriptor
+from dissect.target.plugin import Plugin, export
+
+McAfeeMscLogRecord = TargetRecordDescriptor(
+    "application/av/mcafee/msc",
```
pyrco, line 15: Would it be useful to also split the records in firewall and virus specific versions? As most of the useful data is mutually exclusive anyway?

cecinestpasunepipe: Agreed, this now evolves into separate records.

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,121 @@
...
+                    fields[field.fkey][field.field_id] = field.data
+                log_table = database.table(self.TABLE_LOG)
+                for entry in log_table.rows():
+                    fkey = entry.fkey
+                    log_fields = fields.get(fkey, {})
+                    ip = None
+                    protocol = None
+                    port = None
+                    threat = None
+                    for key, log_field in log_fields.items():
```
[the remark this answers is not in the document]
pyrco: This one seems to not have been applied?

dissect/target/plugins/apps/av/mcafee.py
```diff
@@ -0,0 +1,138 @@
...
+                            continue
+                        except ValueError:
+                            pass
+                        if log_field.startswith(
+                            (self.MARKER_SUSPICIOUS_TCP_CONNECTION, self.MARKER_SUSPICIOUS_UDP_CONNECTION)
+                        ):
+                            port = int(log_field[len_marker:])
+                            protocol = log_field[:3]
+                            continue
+                        if key == self.TEMPLATE_ID_INFECTION and entry.details_info.find(self.MARKER_INFECTION) > -1:
```
pyrco, line 118:
```suggestion

                        if key == self.TEMPLATE_ID_INFECTION and entry.details_info.find(self.MARKER_INFECTION) > -1:
```

pyrco (changes requested):

pyrco (approved):

PR #203, opened 2023-03-10
Fix external references in documentation

martinvanhensbergen:
Website links with a '(' in the URL were not parsed correctly by Sphinx, resulting in broken links. In addition, the word 'References:' is rendered as all caps by Sphinx, whereas in other places 'Sources:' or 'Resources:' are used.

This change set formats the URLs in Sphinx format so that they will render properly. It also changes 'References:' to 'Sources:'.

### DIFF
--- a/dissect/target/plugins/apps/containers/docker.py
+++ b/dissect/target/plugins/apps/containers/docker.py
@@ -39,10 +39,10 @@ DOCKER_NS_REGEX = re.compile(r"\.(?P<nanoseconds>\d{7,})(?P<postfix>Z|\+\d{2}:\d
 
 class DockerPlugin(Plugin):
     """
-    Resources:
-    - https://didactic-security.com/resources/docker-forensics.pdf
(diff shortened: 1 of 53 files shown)


Schamper:
Isn't "References" the correct intended name for the section? Sphinx source seems to suggest so: https://github.com/sphinx-doc/sphinx/blob/aeeadb6f1e9052d8c6a63318510e38d1e3b3a606/sphinx/ext/napoleon/docstring.py#L216

The capitalisation is how our theme chooses to style section headers.

martinvanhensbergen:
I have updated my PR with the following changes:

- As per your suggestion, I have changed all occurrences of 'Sources' or 'Resources' in the meaning of 'References' to the latter term.
- I have changed the manner with which the URL references are fixed; instead of incorporating Sphinx markup, I have escaped the problematic characters; this also stops Sphinx markup to occur when asking help on the commandline using '-h'

Let me know how this approach suits.

pyrco (approved):

PR #205, opened 2023-03-16
Log parser for Trend Micro Worry-Free Suite

cecinestpasunepipe:
Plugin and parsers for Trend Micro Worry-Free Suite:

- Infection logs parser
- Firewall logs parser

(DIS-1524)


dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,125 @@
...
+    uint16    port;
+    uint32    timestamp;
+    char      _pad[8];
+    char      local_ip[65];
+    char      remote_ip[65];
+    char      path[520];
+    wchar     description[128];
+    char      _[10];
+};
+"""
```
Schamper, line 50:
```suggestion
pfwlog_def = """
struct firewall_entry {
    char      _[1];
    char      direction;
    uint16    port;
    uint32    timestamp;
    char      _pad[8];
    char      local_ip[65];
    char      remote_ip[65];
    char      path[520];
    wchar     description[128];
    char      _[10];
};
"""
c_pfwlog = cstruct.cstruct()
c_pfwlog.load(pfwlog_def)
```
Maybe do something more like this, instead of in the class init. This is more in line with how we use cstruct in other places, plus it saves a parse and compilation time on plugin initialization.

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,125 @@
...
+            message (string): Message as reported in the user interface
+            keywords (string): Unparsed fields that might be visible in user interface.
+            fkey (string): Foreign key for reference for further investigation.
+        """
+        with self.target.fs.path(self.LOG_FILE_INFECTIONS).open("rt") as f:
+            linenum = 0
+            while line := f.readline():
+                cells = line.split("<;>")
+                yield TrendMicroWFLogRecord(
+                    ts=datetime.datetime.fromtimestamp(int(cells[9])),
```
Schamper, line 87: Please use `from_unix`, imported from `dissect.util.ts`. This ensures that a timezone is added.

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,125 @@
...
+            path (string): Path to object that initiated/received connection
+            description (string): Description of the detected threat
+        """
+        for firewall_log in self.target.fs.glob_ext(self.LOG_FILE_FIREWALL):
+            print(firewall_log)
+            with firewall_log.open() as log:
+                try:
+                    while entry := self.pfwlog_parser.read("firewall_entry", log):
+                        yield TrendMicroWFFirewallRecord(
+                            ts=datetime.datetime.fromtimestamp(entry.timestamp),
```
Schamper, line 116: Same here.

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,125 @@
...
+            ts (datetime): timestamp.
+            local_ip (net.ipadress): Local IPv4/IPv6.
+            remote_ip (net.ipaddress): Remote IPv4/IPv6.
+            port (uint16): Port of suspicious connection.
+            direction (string): Direction of the traffic
+            path (string): Path to object that initiated/received connection
+            description (string): Description of the detected threat
+        """
+        for firewall_log in self.target.fs.glob_ext(self.LOG_FILE_FIREWALL):
+            print(firewall_log)
```
Schamper, line 111: Debug log.

Schamper (dismissed):

dissect/target/plugins/apps/av/trendmicro.py
cecinestpasunepipe: removed

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+        """Return Trend Micro Worry-free log history records.
+
+        Yields TrendMicroWFLogRecord with the following fields:
+            hostname (string): The target hostname.
+            domain (string): The target domain.
+            ts (datetime): timestamp.
+            threat (string): Description of the detected threat (if available).
+            message (string): Message as reported in the user interface
+            keywords (string): Unparsed fields that might be visible in user interface.
+            fkey (string): Foreign key for reference for further investigation.
```
pyrco, line 80: Maybe just call it `lineno` so it is more clear to an analyst what can be done with it?

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+                try:
+                    while entry := c_pfwlog.firewall_entry(log):
+                        yield TrendMicroWFFirewallRecord(
+                            ts=from_unix(entry.timestamp),
+                            local_ip=entry.local_ip.strip(b"\x00").decode("utf-8"),
+                            remote_ip=entry.remote_ip.strip(b"\x00").decode("utf-8"),
+                            port=entry.port,
+                            direction=("out" if entry.direction == b"\x01" else "in"),
+                            path=entry.path.strip(b"\x00").decode(),
+                            description=entry.description.encode().strip(b"\x00").decode(),
```
pyrco, line 119:
```suggestion
                            description=entry.description.strip("\x00"),
```

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+            with firewall_log.open() as log:
+                try:
+                    while entry := c_pfwlog.firewall_entry(log):
+                        yield TrendMicroWFFirewallRecord(
+                            ts=from_unix(entry.timestamp),
+                            local_ip=entry.local_ip.strip(b"\x00").decode("utf-8"),
+                            remote_ip=entry.remote_ip.strip(b"\x00").decode("utf-8"),
+                            port=entry.port,
+                            direction=("out" if entry.direction == b"\x01" else "in"),
+                            path=entry.path.strip(b"\x00").decode(),
```
pyrco, line 118:
```suggestion
                            path=path.from_windows(entry.path.strip(b"\x00").decode()),
```

The result you are currently getting in your testcase is not what you think it is ;-) It is a path where the `\`'s are part of the filename and not path separators.

Also don't forget to add a 
```
from flow.record.fieldtypes import path
```
in the imports.

Would be interesting to see how filenames with e.g. Chinese characters would be handled. Will it use the 8+3 file name "encoding" also in that instance?

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+            path (string): Path to object that initiated/received connection
+            description (string): Description of the detected threat
+        """
+        for firewall_log in self.target.fs.glob_ext(self.LOG_FILE_FIREWALL):
+            with firewall_log.open() as log:
+                try:
+                    while entry := c_pfwlog.firewall_entry(log):
+                        yield TrendMicroWFFirewallRecord(
+                            ts=from_unix(entry.timestamp),
+                            local_ip=entry.local_ip.strip(b"\x00").decode("utf-8"),
```
pyrco, line 114: `utf-8` is the default encoding used in `decode()`, so either leave it out in all `decode()`'s here and below or add it everywhere.

pyrco (changes requested):

tests/test_plugins_apps_av_trendmicro.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_trendmicro_plugin_worryfree_log(target_win, fs_win):
+    log_file = absolute_path("data/apps/av/trendmicro/pccnt35.log")
+    fs_win.map_file("Program Files (x86)/Trend Micro/Security Agent/Misc/pccnt35.log", log_file)
+    target_win.add_plugin(TrendMicroPlugin)
+    records = list(target_win.trendmicro.wflogs())
+    assert len(records) == 1
+    assert records[0].ts == dt("2023-03-10T15:06:19Z")
+    assert records[0].threat == "Eicar_test_file"
```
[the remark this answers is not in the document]
cecinestpasunepipe: They were not really important but I just wanted to create a uniform struct for all AVs, because fkey also does not apply I think this needs to be further researched now. It is clear to me that the McAfee template was too idiosyncratic to use as a template.

dissect/target/plugins/apps/av/trendmicro.py
cecinestpasunepipe: Worse, it will just store ??? (3F) internally. Only supports cp1252.

tests/test_plugins_apps_av_trendmicro.py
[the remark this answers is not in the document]
pyrco: But I think the path & filename are relevant, which are now also removed?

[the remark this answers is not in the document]
pyrco: But I think the path & filename are relevant, which are now also removed?

pyrco (changes requested):

tests/test_plugins_apps_av_trendmicro.py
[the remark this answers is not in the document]
cecinestpasunepipe: These can still be useful, agreed.

pyrco (approved):

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+from dissect.util.ts import from_unix
+from flow.record.fieldtypes import path
+
+from dissect.target import Target
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import TargetRecordDescriptor
+from dissect.target.plugin import Plugin, export
+
+TrendMicroWFLogRecord = TargetRecordDescriptor(
+    "application/av/trendmicro/wflog",
```
pyrco, line 13:
```suggestion
    "application/av/trendmicro/wf/log",
```
Same as for McAffee?

dissect/target/plugins/apps/av/trendmicro.py
```diff
@@ -0,0 +1,122 @@
...
+        ("datetime", "ts"),
+        ("string", "threat"),
+        ("path", "path"),
+        ("varint", "lineno"),
+    ],
+)
+
+
+TrendMicroWFFirewallRecord = TargetRecordDescriptor(
+    "application/av/trendmicro/wffirewall",
```
pyrco, line 24:
```suggestion
    "application/av/trendmicro/wf/firewall",
```

idem?

pyrco (changes requested):

pyrco (approved):

pyrco (approved):

PR #206, opened 2023-03-17
Fix path file accessor

Zawadidone:
Add UNIX to file accessor operating systems


dissect/target/loaders/velociraptor.py
```diff
@@ -14,17 +14,21 @@
...
     fs_root = path.joinpath(FILESYSTEMS_ROOT)
-    os_type, dirs = find_dirs(fs_root)
-    if os_type in [OperatingSystem.LINUX, OperatingSystem.OSX]:
-        return os_type, [dirs[0]]
+
+    # Linux, Unix and OS-X
+    file_root = fs_root.joinpath("file")
+    if file_root.exist():
+        os_type, dirs = find_dirs(file_root)
+        if os_type in [OperatingSystem.LINUX, OperatingSystem.UNIX ,OperatingSystem.OSX]:
```
Schamper, line 26: `find_dirs` can currently only return `OperatingSystem.LINUX`, `OperatingSystem.OSX` or `OperatingSystem.WINDOWS`. There shouldn't be a need to check against `OperatingSystem.UNIX`.

Schamper (approved):

Issue #215, opened 2023-04-03
Specifying a directory as a target does not load target files that should be picked up by the RawLoader


Schamper:
Given the following directory structure:

```
targets/
├── IE11-Win81-VMWare-disk1.vmdk
├── MSEDGEWIN10_20220708124036.tar
├── SCHARDT.001
└── SCHARDT.raw
```

Performing `target-query targets/` will only load `MSEDGEWIN10_20220708124036.tar` because there's a specific loader for `.tar` files. All the other targets are skipped. These other targets will only load when doing `target-query targets/*`. This is not the expected behavior. All files in this directory should be loaded as targets when passing the directory path.

This is also documented incorrectly in the documentation, see https://github.com/fox-it/dissect-docs/pull/13.

Zawadidone:
Thanks for fixing this issue!

PR #218, opened 2023-04-05
Proposal processing directory target without wildcard

cecinestpasunepipe:
- Load all targets in dir instead of just one
- If it's a loadable dir, ignore non-targets
- Allows for multiple loadable dirs
- Suppress unnecessary errors
- For detailed description of behaviors see scenarios in unit tests (https://github.com/fox-it/dissect.target/pull/218/files#diff-bc9ab330a035ca1d92114c9b2cd7896e93e3def74e0ab9a6c22fa5bbeebcbb0dR23)

(DIS-1919)

Fixes https://github.com/fox-it/dissect.target/issues/215

### DIFF
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -122,7 +122,9 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
     LOADERS_BY_SCHEME[module_name] = loader
 
 
-def find_loader(item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None) -> Optional[Loader]:
+def find_loader(
+    item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None, fallbacks: list = [DirLoader]
+) -> Optional[Loader]:
     """Finds a :class:`Loader` class for the specific ``item``.
 
     This searches for a specific :class:`Loader` classs that is able to load a target pointed to by ``item``.
@@ -134,6 +136,7 @@ def find_loader(item: Path, parsed_path: Optional[urllib.parse.ParseResult] = No
 
     Args:
         item: The target path to load.
+        fallbacks: Fallback loaders to try
 
     Returns:
         A :class:`Loader` class for the specific target if one exists.
@@ -142,7 +145,7 @@ def find_loader(item: Path, parsed_path: Optional[urllib.parse.ParseResult] = No
         if loader := LOADERS_BY_SCHEME.get(parsed_path.scheme):
             return loader
 
-    for loader in LOADERS + [DirLoader]:
+    for loader in LOADERS + fallbacks:
         try:
             if loader.detect(item):
                 return loader
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -123,7 +123,7 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
 
 
 def find_loader(
-    item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None, fallbacks: list = [DirLoader]
+    item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None, fallbacks: list[Loader] = [DirLoader]
 ) -> Optional[Loader]:
     """Finds a :class:`Loader` class for the specific ``item``.
 
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -136,7 +136,7 @@ def find_loader(
 
     Args:
         item: The target path to load.
-        fallbacks: Fallback loaders to try
+        fallbacks: Fallback loaders to try.
 
     Returns:
         A :class:`Loader` class for the specific target if one exists.
--- a/dissect/target/loaders/cb.py
+++ b/dissect/target/loaders/cb.py
@@ -43,10 +43,6 @@ class CbLoader(Loader):
     def detect(path):
         return urlparse(str(path)).scheme == "cb"
 
-    @staticmethod
-    def find_all(path):
-        yield path
-
     def map(self, target):
         for drive in self.session.session_data["drives"]:
             cbfs = CbFilesystem(self.cb, self.sensor, self.session, drive)
--- a/dissect/target/loaders/raw.py
+++ b/dissect/target/loaders/raw.py
@@ -1,7 +1,8 @@
 from pathlib import Path
 
-from dissect.target import Target, container
+from dissect.target import container
 from dissect.target.loader import Loader
+from dissect.target.target import Target
 
 
 class RawLoader(Loader):
@@ -9,5 +10,5 @@ class RawLoader(Loader):
     def detect(path: Path) -> bool:
         return not path.is_dir()
 
-    def map(self, target: Target):
+    def map(self, target: Target) -> None:
         target.disks.add(container.open(self.path))
--- a/dissect/target/loaders/raw.py
+++ b/dissect/target/loaders/raw.py
@@ -6,7 +6,7 @@ from dissect.target.loader import Loader
 
 class RawLoader(Loader):
     @staticmethod
-    def detect(path: Path):
+    def detect(path: Path) -> bool:
         return not path.is_dir()
 
     def map(self, target):
--- a/dissect/target/loaders/raw.py
+++ b/dissect/target/loaders/raw.py
@@ -1,11 +1,13 @@
+from pathlib import Path
+
 from dissect.target import container
 from dissect.target.loader import Loader
 
 
 class RawLoader(Loader):
     @staticmethod
-    def detect(path):
-        return True
+    def detect(path: Path):
+        return not path.is_dir()
 
     def map(self, target):
         target.disks.add(container.open(self.path))
--- a/dissect/target/loaders/raw.py
+++ b/dissect/target/loaders/raw.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 
-from dissect.target import container
+from dissect.target import Target, container
 from dissect.target.loader import Loader
 
 
@@ -9,5 +9,5 @@ class RawLoader(Loader):
     def detect(path: Path) -> bool:
         return not path.is_dir()
 
-    def map(self, target):
+    def map(self, target: Target):
         target.disks.add(container.open(self.path))
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -266,12 +266,9 @@ class Target:
 
             # Search for targets one directory deep
             for entry in _find(path):
-                loader_cls = loader.find_loader(entry, parsed_path=parsed_path)
+                loader_cls = loader.find_loader(entry, parsed_path=parsed_path, fallbacks=[DirLoader, RawLoader])
                 if not loader_cls:
-                    if not entry.is_dir():
-                        loader_cls = RawLoader
-                    else:
-                        continue
+                    continue
 
                 getlogger(entry).debug("Attempting to use loader: %s for: %s", loader_cls, entry)
                 for sub_entry in loader_cls.find_all(entry):
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -270,7 +270,7 @@ class Target:
                 if not loader_cls:
                     continue
 
-                getlogger(entry).debug("Attempting to use loader: %s for: %s", loader_cls, entry)
+                getlogger(entry).debug("Attempting to use loader: %s", loader_cls)
                 for sub_entry in loader_cls.find_all(entry):
                     try:
                         ldr = loader_cls(sub_entry, parsed_path=parsed_path)
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -284,8 +284,8 @@ class Target:
                         at_least_one_loaded = True
                         yield target
 
-                    except Exception as error:
-                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=error)
+                    except Exception as e:
+                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=e)
 
                     if include_children:
                         try:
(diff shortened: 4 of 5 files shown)


dissect/target/target.py
```diff
@@ -256,70 +257,43 @@ def _find(find_path: Path):
...
+                getlogger(entry).debug("Attempting to use loader: %s for: %s", loader_cls, entry)
                 for sub_entry in loader_cls.find_all(entry):
                     try:
                         ldr = loader_cls(sub_entry, parsed_path=parsed_path)
                     except Exception as e:
-                        getlogger(sub_entry).error("Failed to initiate loader", exc_info=e)
-                        continue
+                        if not path.is_dir():
+                            getlogger(sub_entry).error("Failed to initiate loader", exc_info=e)
+                            continue
```
Schamper, line 277: Should we not still `continue` here?

dissect/target/target.py
```diff
@@ -256,70 +257,43 @@ def _find(find_path: Path):
...
                     if include_children:
                         try:
                             yield from target.open_children()
                         except Exception as e:
                             getlogger(sub_entry).error("Failed to load child target from %s", target, exc_info=e)
 
-                # Found a compatible loader for the top level path, no need to search a level deeper
-                # Going deeper could cause unwanted behaviour
-                if loaded and entry is path:
-                    break
```
Schamper, line 302: Is this not still a desired `break` condition?

dissect/target/target.py
```diff
@@ -256,70 +257,43 @@ def _find(find_path: Path):
...
-                    getlogger(entry).error("Failed to find loader for target", exc_info=e)
-
-                if include_children:
-                    try:
-                        yield from target.open_children()
-                    except Exception as e:
-                        getlogger(sub_entry).error("Failed to load child target from %s", target, exc_info=e)
-
-        if not loaded:
-            raise TargetError(f"Failed to find any loader for targets: {paths}")
```
Schamper, line 322: Is it desired to lose this exception?

dissect/target/target.py
```diff
@@ -256,39 +257,34 @@ def _find(find_path: Path):
...
                         # Attempt to load the target using this loader
                         target = cls._load(sub_entry, ldr)
-                        loaded = True
+                        at_least_one_loaded = True
                         yield target
-                    except Exception as e:
-                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=e)
-                        continue
+                        if entry is path:
+                            break
```
Schamper, line 285: I'm not sure if we should break in this loop for this condition (`loader_cls.find_all(entry)`), because the loader may still yield one or more sub entries, even if `entry is path`.

cecinestpasunepipe: If entry is path, the entry seems to be loaded by the dirloader, no other entries can be processed then I assume. To clarify things (at least for myself) I am writing some unit tests for this, that might also make it easier to talk about the different scenarios that can happen. As far as I can see these are not yet tested...?

dissect/target/target.py
cecinestpasunepipe: Researching...

dissect/target/target.py
cecinestpasunepipe: Changed, see unit tests for an overview of all possible scenarios.

dissect/target/loaders/cb.py
```diff
@@ -43,10 +43,6 @@ def __init__(self, path, **kwargs):
     def detect(path):
         return urlparse(str(path)).scheme == "cb"
 
-    @staticmethod
-    def find_all(path):
-        yield path
-
```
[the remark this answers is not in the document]
Schamper: Yeah - it served as a reminder to implement it for that specific loader :) can be removed for now.

tests/test_target.py
```diff
@@ -1,11 +1,121 @@
...
+            [
+                "/dir/etc",
+                "/dir/var",
+                "/dir/raw.img",
+                "/dir/info.txt",
+            ],
+            "/dir",
+            "[DirLoader('/dir')]",
+            0,
+        ),
```
Schamper, line 81: Can you add a scenario for Windows directory targets? So `/dir/c:/windows/system32`. Because here both `/dir` and `/dir/c:` are valid `DirLoader` targets because it supports Windows-like directories with and without directory letters.
```

Comment (cecinestpasunepipe) [reply tests/test_target.py:81]: Added win-scenarios

File: dissect/target/target.py [hunk 7]
```diff
@@ -252,28 +254,26 @@ def open_all(cls, paths: list[Union[str, Path]], include_children: bool = False)
...
-
                 if not loader_cls:
-                    continue
-
-                getlogger(entry).debug("Found compatible loader: %s", loader_cls)
-                found_loader = True
+                    if not entry.is_dir():
+                        loader_cls = RawLoader
+                    else:
+                        continue
```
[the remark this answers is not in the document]
cecinestpasunepipe: Okay, that sounds like a nice idea, I have given it a try...

dissect/target/target.py
```diff
@@ -285,10 +282,14 @@ def _find(find_path: Path):
...
                         target = cls._load(sub_entry, ldr)
                         loaded = True
+                        at_least_one_loaded = True
                         yield target
-                    except Exception as e:
-                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=e)
-                        continue
+
+                        # If DirLoader, ignore further entries. Otherwise search children/nested
+                        if isinstance(ldr, DirLoader):
```
Schamper, line 289: Instinctively this break feels like it's one indent too far. I know that `DirLoader` currently has no custom `find_all()` implemented, but if it would, would this not break after the first entry?

dissect/target/target.py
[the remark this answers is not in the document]
cecinestpasunepipe: I figured this is a nice solution, what do you think?

dissect/target/target.py
cecinestpasunepipe: Yes, this is intended. It is not desirable for the DirLoader to generate subentries as the dir as a whole is loaded, this is almost by definition I would say. I think that generating subentries for the Dirloader would make things quite complex because then the DirLoader either encompasses the whole dir or it loads all subdirs and this may interfere with other loaders and files to be ignored. I think it would be better to simplify at this point. But maybe we do not understand each other correctly or I am missing something. Feel free to correct me.

Schamper: It's half that it feels like the code would be more correct with it being one indent level less (but still have the same effect, in the current state of the code base), and half that it may have a side-effect in a future loader scenario.

Currently, `DirLoader`'s `find_all()` implementation only yields itself. This is correct. However, say a hypothetical situation and format exists that would look like this:

```
/root/metadata.bin
/root/data/host1/<dir structure>
/root/data/host2/<dir structure>
/root/data/host3/<dir structure>
```

In this hypothetical situation, `metadata.bin` would contain important information that is necessary to load data from `host1`, `host2` and `host3`. For example, a decryption key. A possible loader for this could subclass from `DirLoader` and could implement the `detect()` method to look for the existence of `metadata.bin`, so `/root` is a valid target for this loader. However, it actually contains multiple "targets", so its `find_all()` implementation would iterate from the `/root/data` directory.

With the break being where it is now, only `/root/data/host1` would get returned.

Of course there are many other different questions and problems with this scenario (why subclass from `DirLoader`? why not create a `detect()` that looks for a `metadata.bin` in its parent directories?) but it does highlight why I believe it should be one indent less.

Then again, of course, I could also be overlooking something why one indent less would break :smile:

cecinestpasunepipe: Ok, convinced, thanks for your explanation, this seems to make sense.

dissect/target/loader.py
```diff
@@ -122,7 +122,9 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
     LOADERS_BY_SCHEME[module_name] = loader
 
 
-def find_loader(item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None) -> Optional[Loader]:
+def find_loader(
+    item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None, fallbacks: list = [DirLoader]
```
Schamper, line 126:
```suggestion
    item: Path, parsed_path: Optional[urllib.parse.ParseResult] = None, fallbacks: list[Loader] = [DirLoader]
```

dissect/target/loader.py
```diff
@@ -134,6 +136,7 @@ def find_loader(item: Path, parsed_path: Optional[urllib.parse.ParseResult] = No
 
     Args:
         item: The target path to load.
+        fallbacks: Fallback loaders to try
```
Schamper, line 139:
```suggestion
        fallbacks: Fallback loaders to try.
```

dissect/target/target.py
```diff
@@ -252,28 +254,23 @@ def open_all(cls, paths: list[Union[str, Path]], include_children: bool = False)
...
-                loader_cls = loader.find_loader(entry, parsed_path=parsed_path)
-
+                loader_cls = loader.find_loader(entry, parsed_path=parsed_path, fallbacks=[DirLoader, RawLoader])
                 if not loader_cls:
                     continue
 
-                getlogger(entry).debug("Found compatible loader: %s", loader_cls)
-                found_loader = True
-
+                getlogger(entry).debug("Attempting to use loader: %s for: %s", loader_cls, entry)
```
Schamper, line 273: Question: doesn't `getlogger(entry)` already make the log lines easily findable for that entry (path)?

Schamper (changes requested):

Schamper (changes requested):

dissect/target/target.py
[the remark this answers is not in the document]
cecinestpasunepipe: Should it not be a continue then, you only want to skip children right? One indent to the left only moves it out of the try btw.

dissect/target/loaders/raw.py
```diff
@@ -1,11 +1,13 @@
...
 
 class RawLoader(Loader):
     @staticmethod
-    def detect(path):
-        return True
+    def detect(path: Path) -> bool:
+        return not path.is_dir()
 
-    def map(self, target):
+    def map(self, target: Target):
```
Schamper, line 12:
```suggestion
    def map(self, target: Target) -> None:
```

dissect/target/target.py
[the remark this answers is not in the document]
Schamper: No, "children" here are nested targets, such as nested virtual machine files. That is very much still a valid use case for directory targets.

Schamper (changes requested):

dissect/target/target.py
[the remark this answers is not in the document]
cecinestpasunepipe: I do not understand this, the Dirloader expects a directory reflecting a target, it should not open children because this interferes with the Dirloader concept itself. For instance, a subdir /system may either contain a new Dirloadable target or be part of the current Dirtarget. So I don't think this is a good idea. Maybe I should make a test implementation so we can test this scenario, I always have trouble implementing hypothetical functionality, hard to wrap my head around, so I'll craft a little example to gain better understanding.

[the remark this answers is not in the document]
Schamper: Those are not the same thing - "children" in this context are those detected by the `ChildPlugin`'s, e.g. such as this one: https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/child/hyperv.py. Not "child directories".

A directory target could contain these `.vmcx` and `.vhdx` files, so child targets are still relevant to directory targets.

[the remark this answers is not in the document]
cecinestpasunepipe: O, I see. I knew they were VMs but I did not know it was really a hard limitation, I thought people could write their own child plugins as well and they could be anything conceptually. I would consider renaming this to VMs (and vm-plugins) then, because child may lead to a more open interpretation (as we have just witnessed - even while I knew the "current" implementations were only VMs).

Given these facts, I suggest we remove the dirloader check completely. Indenting it to the left will break stuff because it will also break on a failed ldr (no check for loaded flag). And... if we fix that we end up with duplicate code with the line below that already breaks if the top-level is loaded.

For completeness and understanding I added a 5th scenario to demonstrate that the hypothetical Dir-based loader will work as expected given the current update. Feel free to request any additional scenarios or test cases. If I overlook something else feel free to point it out to me. Sorry for not the misunderstanding on my side.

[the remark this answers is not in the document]
Schamper: VM's are just an example though, they don't necessarily have to be VM's. For example, they can also be Docker containers or [Virtuozzo](https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/child/virtuozzo.py) containers. People can indeed write their own child plugins and it can conceptually be anything, but it's anything _within_ a target. If a `DirLoader` results in a valid target with a valid Windows or Linux root filesystem, anything that can be in there that can be seen as a nested target (VM, container, whatever) can be loaded by a child plugin.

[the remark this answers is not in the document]
cecinestpasunepipe: Ok fair enough.

dissect/target/target.py
```diff
@@ -285,10 +281,11 @@ def _find(find_path: Path):
...
                         target = cls._load(sub_entry, ldr)
                         loaded = True
+                        at_least_one_loaded = True
                         yield target
-                    except Exception as e:
-                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=e)
-                        continue
+
+                    except Exception as error:
+                        getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=error)
```
Schamper, line 288: You could argue that the renaming of `e` to `error` doesn't add much value here, and this change could be reduced to just the removal of the `continue` statement. It's also (currently) less consistent with similar code constructs.

cecinestpasunepipe: Yes, agreed, it does not add much value. Changed.

Schamper (approved):

Schamper:
Code looks good, just the PR title/commit title will need to be changed.

PR #219, opened 2023-04-06
Fix ASDF loader

Schamper:
(DIS-1947)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,4 @@ tests/data/plugins/browsers/firefox/places.sqlite filter=lfs diff=lfs merge=lfs 
 tests/data/plugins/browsers/chrome/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/edge/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/chromium/History.sqlite filter=lfs diff=lfs merge=lfs -text
+tests/data/loaders/asdf/metadata.asdf filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -5,3 +5,4 @@ tests/data/plugins/browsers/firefox/places.sqlite filter=lfs diff=lfs merge=lfs
 tests/data/plugins/browsers/chrome/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/edge/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/chromium/History.sqlite filter=lfs diff=lfs merge=lfs -text
+tests/data/loaders/asdf/metadata.asdf filter=lfs diff=lfs merge=lfs -text
```
cecinestpasunepipe, line 8: Is it necessary to add an 11K file to LFS? I thought this was meant for large files?

dissect/target/loaders/asdf.py
```diff
@@ -23,11 +23,4 @@ def map(self, target):
...
 
-        overlay = target.fs.add_layer()
-        for member in self.asdf.metadata.members():
-            if member.isdir():
-                continue
-
-            path = fsutil.join(self.METADATA_PREFIX, member.name)
-            entry = TarFile(overlay, path, member.name, self.asdf.metadata.tar)
-            overlay.map_file_entry(entry.path, entry)
+        target.fs.mount(self.METADATA_PREFIX, TarFilesystem(self.asdf.open(IDX_METADATA)))
```
cecinestpasunepipe, line 26: I cannot comment on lines that have not changed, but since we are updating the ADSF-loader maybe it is a good idea to also add the types?

cecinestpasunepipe (changes requested):
Seems to work fine. Some minor suggestions/notes.

.gitattributes
Schamper: Maybe good to talk about what exactly the boundary is in a standup, because I'm not sure anymore either :)

cecinestpasunepipe (approved):

PR #220, opened 2023-04-10
Add support for Parallels PVM/PVS/HDD/HDS

Schamper:
Depends on https://github.com/fox-it/dissect.hypervisor/pull/16


dissect/target/containers/hdd.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @staticmethod
+    def detect_fh(fh: BinaryIO, original: Union[list, BinaryIO]) -> bool:
+        return False
+
+    @staticmethod
+    def detect_path(path: Path, original: Union[list, BinaryIO]) -> bool:
+        return path.suffix.lower() == ".hdd"
+
+    def read(self, length: int) -> bytes:
```
cecinestpasunepipe, line 27: Maybe this issue has been raised way before my involvement but... I have looked at all the containers and the read/seek/tell/close methods seem to be much alike. Isn't it an idea to move these to the parent? That would reduce the amount of code.

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #150, opened 2023-01-30
Improve locale for unix and windows

JSCU-CNI:
Some minor improvements and refactors for timezone and locale handling on linux and windows targets.

* adds normalize_timezone helper to translate windows timezones to iana tzdb format
* adds normalize_language function to transform unix and windows locales to ISO-3166 format.
* replaces all occurences of datetime.utcfromtimestamp with ts.from_unix
* improves tests to reflect changes

PR #223, opened 2023-04-11
Add setuptools package-data to pyproject.toml

JSCU-CNI:
This patch explicitly adds `helpers/data/*.xml` to setuptools' package_data. This was [previously implemented](https://github.com/fox-it/dissect.target/pull/150/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R54) in `setup.py` but seems to have been implicitly ported to the new `pyproject.toml`.

Certain build processes can cause unexpected behavior if package data is not explicitly added to `pyproject.toml` files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,4 +89,7 @@ license-files = ["LICENSE", "COPYRIGHT"]
 [tool.setuptools.packages.find]
 include = ["dissect.*"]
 
+[tool.setuptools.package-data]
+"dissect.target" = ["helpers/data/*.xml"]
+
 [tool.setuptools_scm]


pyrco:
The default setuptools behaviour with the PEP-621 setup, which is the current supported way of building these packages, is to include all package data by default. Can you elaborate a bit in which circumstances this may fail?

As all other dissect projects which have extra package data also depend on the same mechanism, we're do not really want to revert to the old behavior of explicitly tracking all these files manually.

pyrco (commented):
The default setuptools behaviour with the PEP-621 setup, which is the current supported way of building these packages, is to include all package data by default. Can you elaborate a bit in which circumstances this may fail?

As all other dissect projects which have extra package data also depend on the same mechanism, we're do not really want to revert to the old behavior of explicitly tracking all these files manually.

JSCU-CNI:
Thank you for elaborating. This seems to work when explicitly upgrading the `build-system`'s `setuptools` environment. Perhaps the minimum required setuptools version should be updated, I haven't looked into this in detail.

On a side note, inexplicitly including package data has the (unintended?) effect of including `tests/data` in your dist release tar files (~60MB). Explicitly stating which package data should be included has the benefit of not including any unwanted files by accident.

PR #226, opened 2023-04-12
Add LNK test

Zawadidone:


dissect/target/plugins/os/windows/lnk.py
```diff
@@ -12,20 +13,20 @@
 LnkRecord = TargetRecordDescriptor(
     "windows/filesystem/lnk",
     [
-        ("uri", "lnk_path"),
+        ("path", "lnk_path"),
         ("string", "lnk_name"),
         ("datetime", "lnk_mtime"),
         ("datetime", "lnk_atime"),
         ("datetime", "lnk_ctime"),
         ("string", "lnk_relativepath"),
```
Zawadidone, line 21: The field `lnk_relativepath` has the field type `string` but the following values are parsed when using the plugin?

* `..\..\..\..\..\Program Files\Wireshark\Wireshark.exe`
* `..\..\..\..\..\Program Files\Microsoft Office\root\Office16\MSACCESS.EXE`
* `.\pestudio.exe`

Can the field type `path` be used for this field? Or is the usage of the fieldtype `string` best practice because the values are not full Windows paths like `C:\Program Files\Wireshark\Wireshark.exe`.

tests/test_plugins_os_windows_lnk.py
```diff
@@ -0,0 +1,36 @@
...
+    target_win.add_plugin(LnkPlugin)
+
+    records = list(target_win.lnk(None))
+
+    assert len(records) == 1
+
+    assert isinstance(records[0], type(LnkRecord()))
+    assert str(records[0].lnk_path) == "sysvol\\users\\pestudio.lnk"
+    assert records[0].lnk_name is None
+    assert records[0].lnk_mtime == datetime("2023-04-12T08:50:02.008187+00:00")
```
Zawadidone, line 21: Which one is preferred, using the function `datetime` or converting the value to a string?

* `assert records[0].lnk_mtime == datetime("2023-04-12T08:50:02.008187+00:00")`
* `assert str(records[0].lnk_mtime) == "2023-04-12 08:50:02.008187+00:00"`

dissect/target/plugins/os/windows/lnk.py
```diff
@@ -72,14 +73,14 @@ def lnk(self, path: Optional[str]) -> Iterator[LnkRecord]:
...
+        codepage = self.target.codepage or "ascii"
 
-        for entry in self.lnk_entries(path):
+        for entry in self.lnk_entries(directory):
             lnk_file = Lnk(entry.open())
             lnk_net_name = lnk_device_name = None
 
             if lnk_file.link_header:
-                lnk_path = str(entry)
+                lnk_path = path.from_windows(entry)
```
Schamper, line 83:
```suggestion
                lnk_path = entry
```
Entry is a `(Target)Path` object so there should be no need to do this. Speaking of, can you change the return type of `lnk_entries` to `Iterator[Path]` instead?

dissect/target/plugins/os/windows/lnk.py
Zawadidone: Okay

dissect/target/plugins/os/windows/lnk.py
```diff
@@ -154,7 +163,7 @@ def lnk(self, path: Optional[str]) -> Iterator[LnkRecord]:
                     target_ctime=target_ctime,
                 )
 
-    def lnk_entries(self, path: Optional[str] = None) -> Iterator[TargetPath]:
+    def lnk_entries(self, path: Optional[str] = None) -> Iterator[Path]:
```
pyrco, line 166:
```suggestion
    def lnk_entries(self, path: Optional[str] = None) -> Iterator[TargetPath]:
```

One small thing, the return type should remain `TargetPath`. Input typing can be broader if the function can deal with any generic `Path` (sub)class, but return types should be specific so you know what (sub)class to exepect.

pyrco (changes requested):

pyrco (approved):

PR #231, opened 2023-04-18
Add parser for extensions for chromium browsers

Wouter-Jansen:
DIS-1954

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ pip install dissect.target
 
 This module is also automatically installed if you install the `dissect` package.
 
+If you wish to use the YARA plugin (`target-query -f yara`), you can install `dissect.target[yara]` to automatically 
+install the `yara-python` dependency.
+
 ## Tools inside this project
 
 ### target-query
--- a/dissect/target/plugins/filesystem/yara.py
+++ b/dissect/target/plugins/filesystem/yara.py
@@ -1,6 +1,10 @@
 from pathlib import Path
 
-import yara
+try:
+    import yara
+except ImportError:
+    raise ImportError("Please install 'yara-python' to use 'target-query -f yara'.")
+
 from flow.record.fieldtypes import uri
 
 from dissect.target.exceptions import FileNotFoundError
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ full = [
     "fusepy",
     "pycryptodome",
     "pyyaml",
-    "yara-python",
     # dissect.target's caching uses flow.record functionlity which depends on the
     # zstandard module being available. However flow.record does not define
     # zstandard as a dependency, nor does it allow zstandard to be installed
@@ -62,6 +61,12 @@ full = [
     # Until such time that this dependency can be installed through
     # flow.record, we define it as a dependency of dissect.target.
     "zstandard",
+]
+yara = [
+    # Grab the dependencies for dissect.target
+    "dissect.target[full]",
+    # Enable using the yara rules functionality
+    "yara-python"
 ]
 
 [project.scripts]
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ minversion = 4.2.4
 requires = virtualenv>=20.16.6
 
 [testenv]
-extras = full
+extras = yara
 deps =
     pytest
     pytest-cov


dissect/target/plugins/browsers/browser.py
```diff
@@ -95,6 +104,62 @@ def _func(self, func_name: str):
             except Exception:
                 self.target.log.exception("Failed to execute browser plugin: %s.%s", plugin_name._name, func_name)
 
+    def check_compatible(self) -> bool:
+        """Perform a compatibility check with the target.
+        This function checks if any of the supported browser plugins
+        can be used. Otherwise it should raise an ``UnsupportedPluginError``.
+        Raises:
+            UnsupportedPluginError: If the plugin could not be loaded.
+        """
```
Schamper, line 113:
```suggestion
```
I don't think `check_compatible` requires a docstring here.

dissect/target/plugins/browsers/browser.py
```diff
@@ -95,6 +104,62 @@ def _func(self, func_name: str):
...
+        if not len(self._plugins):
+            raise UnsupportedPluginError("No compatible browser plugins found")
+
+    @export(record=BrowserDownloadRecord)
+    def downloads(self):
+        """Return browser download records from all browsers installed and supported.
+
+        Yields:
+            BrowserDownloadRecord with the following fieds:
```
Schamper, line 122:
```suggestion
        Yields BrowserDownloadRecord with the following fieds:
```
This is more in line with the other docstrings.

dissect/target/plugins/browsers/browser.py
```diff
@@ -95,6 +104,62 @@ def _func(self, func_name: str):
...
+
+        Browser extensions for Chrome, Chromium, and Edge (Chromium).
+
+        Yields BrowserExtensionRecord with the following fields:
+            hostname (string): The target hostname.
+            domain (string): The target domain.
+            ts_install (datetime): Extension install timestamp.
+            ts_update (datetime): Extension update timestamp.
+            browser (string): The browser from which the records are generated.
+            id (string) = Extension unique identifier.
```
Schamper, line 149:
```suggestion
            id (string): Extension unique identifier.
```

dissect/target/plugins/browsers/chrome.py
```diff
@@ -17,25 +18,33 @@ class ChromePlugin(ChromiumMixin, Plugin):
         # Windows
         "AppData/Local/Google/Chrome/User Data/Default",
         "AppData/Local/Google/Chrome/continuousUpdates/User Data/Default",
-        "Local Settings/Application Data/Google/Chrome/User Data/Default",
```
Schamper, line 20: Why was this removed?

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,86 @@ class ChromiumMixin:
...
+        for user_details in self.target.user_details.all_with_home():
+            for d in hist_paths:
+                cur_dir: TargetPath = user_details.home_path.joinpath(d)
+                if not cur_dir.exists():
+                    continue
+                users_dirs.append((user_details.user, cur_dir))
+        return users_dirs
+
+    def _iter_db(self, filename: str) -> Iterator[SQLite3]:
+        """Generate a connection to a sqlite history database file.
```
Schamper, line 57: There's no "connection", it's a parsed database. So "an opened SQLite database file" would be more accurate I think.

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,86 @@ class ChromiumMixin:
...
+            except FileNotFoundError:
+                self.target.log.warning("Could not find %s file: %s", filename, json_file)
+
+    def check_compatible(self) -> None:
+        """Perform a compatibility check with the target.
+        This function checks if any of the supported browser directories
+        exists. Otherwise it should raise an ``UnsupportedPluginError``.
+        Raises:
+            UnsupportedPluginError: If the plugin could not be loaded.
+        """
```
Schamper, line 105:
```suggestion
```

dissect/target/plugins/browsers/chromium.py
```diff
@@ -154,54 +166,155 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
...
-            FileNotFoundError: If the history file could not be found.
-            SQLError: If the history file could not be opened.
+            - Iterator[BrowserExtensionRecord]: A generator that yields `BrowserExtensionRecord`
+                with the following fields:
+                    hostname (string): The target hostname.
+                    domain (string): The target domain.
+                    ts_install (datetime): Extension install timestamp.
+                    ts_update (datetime): Extension update timestamp.
+                    browser (string): The browser from which the records are generated.
+                    id (string) = Extension unique identifier.
```
Schamper, line 183:
```suggestion
                    id (string): Extension unique identifier.
```

Schamper (changes requested):

dissect/target/plugins/browsers/chromium.py
```diff
@@ -154,54 +166,155 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
...
+                            name=name,
+                            short_name=short_name,
+                            default_title=default_title,
+                            description=description,
+                            version=ext_version,
+                            ext_path=ext_path,
+                            from_webstore=extensions.get(extension_id).get("from_webstore"),
+                            permissions=ext_permissions,
+                            manifest_version=manifest_version,
+                            source=str(json_file),
```
pyrco, line 248:
```suggestion
                            source=json_file.as_posix(),
```
flow.record's `path` will interpret a string  according to the path rules of the system it is currently running on. `str(TargetPath)` will give back the path in a representation according to the path rules of the system that is analyzed. So when analyzing a windows target while running onder Linux, it may give back `C:\some\path`, which flow record will then see as a single file (as `\` is not a path separator in Linux).

Paths in posix representation (with a '/'), will always be correctly represented. You will miss the information that it was originally a windows path though. If you want to also capture that, you'd have to do the same as for `ext_path`:

```suggestion
                           if json_file and self.target.os == "windows":
                                json_file = path.from_windows(json_file)
                            elif json_file:
                                json_file = path.from_posix(json_file)
                                
                             ...
                             
                            source=json_file,
```

dissect/target/plugins/browsers/chromium.py
```diff
@@ -154,54 +166,155 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
...
+                        description=None,
+                        rev_host=None,
+                        visit_type=None,
+                        visit_count=url.visit_count,
+                        hidden=url.hidden,
+                        typed=None,
+                        session=None,
+                        from_visit=row.from_visit or None,
+                        from_url=try_idna(from_url.url) if from_url else None,
+                        source=str(db_file),
```
pyrco, line 312: See the comment for source in `BrowserExtensionRecord`

pyrco (changes requested):

dissect/target/plugins/browsers/chrome.py
Schamper: But now it won't work on systems that don't have `AppData` but do have `Local Settings`. I know that's incredibly unlikely (XP running Chrome probably?), but IMO a better solution would be to perhaps use `.resolve()` to get a full absolute path, and filter out duplicate paths in the iteration process.

dissect/target/plugins/browsers/chromium.py
```diff
@@ -147,61 +154,162 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
...
+                            name=name,
+                            short_name=short_name,
+                            default_title=default_title,
+                            description=description,
+                            version=ext_version,
+                            ext_path=ext_path,
+                            from_webstore=extensions.get(extension_id).get("from_webstore"),
+                            permissions=ext_permissions,
+                            manifest_version=manifest_version,
+                            source=json_file.as_posix(),
```
pyrco, line 243: On second thoughts, I think for now we should go the long way here, so treat this similar to `ext_path`, e.g.:

```suggestion
                            if json_file and self.target.os == "windows":
                                json_file = path.from_windows(json_file)
                            elif json_file:
                                json_file = path.from_posix(json_file)

                            ...

                            source=json_file
```

This should actually be fixed in flow.record so we don't need to do this dance for every TargetPath type path (for stringy typed path you won't be able to circumvent explicitly casting unfortunately). I'm working on some updates in flow.record, so once they are in place we can remove the workarounds here.

As an additional note, I forgot that in the current context `source=json_file` would have resulted in similar behavior as `source=json_file.as_posix(). Just FYI.

pyrco (changes requested):

pyrco (approved):

dissect/target/plugins/browsers/browser.py
```diff
@@ -95,6 +104,55 @@ def _func(self, func_name: str):
...
 
+    def check_compatible(self) -> bool:
+        if not len(self._plugins):
+            raise UnsupportedPluginError("No compatible browser plugins found")
+
+    @export(record=BrowserDownloadRecord)
+    def downloads(self):
+        """Return browser download records from all browsers installed and supported.
+
+        Yields BrowserDownloadRecord with the following fieds:
```
Schamper, line 115:
```suggestion
        Yields BrowserDownloadRecord with the following fields:
```

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,81 @@ class ChromiumMixin:
...
+                yield user, db_file, sqlite3.SQLite3(db_file.open())
+            except FileNotFoundError:
+                self.target.log.warning("Could not find %s file: %s", filename, db_file)
+            except SQLError as e:
+                self.target.log.warning("Could not open %s file: %s", filename, db_file, exc_info=e)
 
-                for row in db.table("visits").rows():
-                    visits[row.id] = row
-                    url = urls[row.url]
+    def _iter_json(self, filename: str) -> Iterator[tuple[str, TargetPath, dict]]:
```
Schamper, line 78:
```suggestion
    def _iter_json(self, filename: str) -> Iterator[tuple[str, Path, dict]]:
```

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,81 @@ class ChromiumMixin:
...
-        """Return browser history records from supported Chromium-based browsers.
+    def _build_userdirs(self, hist_paths: list[str]) -> list[tuple]:
+        """Join the selected browser dirs with the user home path.
 
         Args:
-            browser_name: The name of the browser as a string.
+            hist_paths: A list with browser paths as strings.
+
+        Returns:
+            list with tuples containing user and TargetPath object.
```
Schamper, line 45:
```suggestion
            List of tuples containing user and history file path objects.
```

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,81 @@ class ChromiumMixin:
...
-                        from_url=try_idna(from_url.url) if from_url else None,
-                        source=str(db_file),
-                        _target=self.target,
-                        _user=user,
-                    )
-            except SQLError as e:
-                self.target.log.warning("Error processing history file: %s", db_file, exc_info=e)
+        Yields:
+            Tuple[str, TargetPath, dict]: A tuple containing the name of the
+            user, the path to the JSON file, and the parsed JSON data.
```
Schamper, line 88:
```suggestion
            A tuple containing the name of the user, the path to the JSON file, and the parsed JSON data.
```

Schamper (dismissed):

dissect/target/plugins/browsers/chromium.py
```diff
@@ -23,76 +25,81 @@ class ChromiumMixin:
...
+    BrowserExtensionRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/chromium/extension", GENERIC_EXTENSION_RECORD_FIELDS
+    )
+    BrowserHistoryRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/chromium/history", GENERIC_HISTORY_RECORD_FIELDS
+    )
 
-    def history(self, browser_name: str = None) -> Iterator[BrowserHistoryRecord]:
-        """Return browser history records from supported Chromium-based browsers.
+    def _build_userdirs(self, hist_paths: list[str]) -> list[tuple]:
```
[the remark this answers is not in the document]
pyrco: The second part of the tuple will always be a `TargetPath` as it is extracted from a `Target`'s fs.

dissect/target/plugins/browsers/chromium.py
Schamper: Yes, but everywhere else in `dissect.target` we typehint `Path` instead of `TargetPath`, just to provide a more generic type hint. Similar to how you typehint ` fh` as `BinaryIO` instead of `AlignedStream`. There's not really any specific reason to typehint as specific as `TargetPath` IMO.

pyrco: For input parameters, if you don't use the `TargetPath` specific bits I agree. For output types, I don't. Here it does matter that you don't get a `Path` but a `TargetPath` as it will have different properties and behave slightly different from `Path`. Now for instance I need to trace back the code to make sure I actually have the `._fs` property available when I get back the path.

Granted in this case it is not needed because the consumer of this function does not use the `TargetPath` specifics,. But as a function you'll never know who your consumers will be.

dissect/target/plugins/browsers/chromium.py
```diff
@@ -14,85 +16,91 @@
...
+                yield user, db_file, sqlite3.SQLite3(db_file.open())
+            except FileNotFoundError:
+                self.target.log.warning("Could not find %s file: %s", filename, db_file)
+            except SQLError as e:
+                self.target.log.warning("Could not open %s file: %s", filename, db_file, exc_info=e)
 
-                for row in db.table("visits").rows():
-                    visits[row.id] = row
-                    url = urls[row.url]
+    def _iter_json(self, filename: str) -> Iterator[tuple[str, Path, dict]]:
```
pyrco, line 80:
```suggestion
    def _iter_json(self, filename: str) -> Iterator[tuple[str, TargetPath, dict]]:
```

pyrco (changes requested):

dissect/target/plugins/browsers/chromium.py
Schamper: Fair argument, and I agree. Let's change the return value to `TargetPath` in that case and try to maintain that in other places too.

pyrco (approved):

PR #234, opened 2023-04-21
Add evt/evtx-loader

cecinestpasunepipe:
(DIS-523)


dissect/target/loaders/log.py
```diff
@@ -0,0 +1,31 @@
...
+        "EvtxPlugin": "sysvol/windows/system32/winevt/logs",
+        "EvtPlugin": "sysvol/windows/system32/config",
+    }
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        return False
+
+    def appendix(self, plugin_method: Callable) -> dict[str, str]:
+        return {"logs_dir": self.LOGS_DIRS[str(plugin_method.__self__.__class__.__name__)]}
```
Schamper, line 20: How does this work? Why not detect a directory with `.evt` or `.evtx`, map them to the default directories, and call it a day? Can you provide some comments or explanation on why this route was taken?

cecinestpasunepipe: That will not work since get_logs() will try to obtain paths from the registry in both evt.pyt and evtx.py. The only way to circumvent this is to pass a specific logs_dir, so that is what I am doing on behalf of the user. But maybe I am overlooking something, feel free to correct me if I am wrong.

cecinestpasunepipe: No, I see no reference to a default directory, instead it checks whether EVENTLOG_REGISTRY_KEY is set:

https://github.com/fox-it/dissect.target/blob/8c90fbd3b917ad9d36f01b98f6dc586f8e881da8/dissect/target/plugins/os/windows/log/evt.py#L54

Is this a bug?

Schamper: 1 line up: https://github.com/fox-it/dissect.target/blob/8c90fbd3b917ad9d36f01b98f6dc586f8e881da8/dissect/target/plugins/os/windows/log/evt.py#L52

cecinestpasunepipe: Yes, but then at line 54 it queries the registry anyway, there is no condition that checks whether that last part can be skipped so is this correct behaviour?

Schamper: Ah, yes, that’s indeed a bug then. By all means go ahead and fix that!

dissect/target/loaders/log.py
```diff
@@ -0,0 +1,27 @@
...
+
+class LogLoader(Loader):
+    LOGS_DIRS = {
+        "EvtxPlugin": "sysvol/windows/system32/winevt/logs",
+        "EvtPlugin": "sysvol/windows/system32/config",
+    }
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        return False
```
Schamper, line 16: This could check for file or directory containing `*.evtx` or `*.evt` files perhaps.

Schamper: Sounds reasonable. We can always reevaluate in the future if we see a need for it.

Regarding `rglob`, I don't think `rglob` should be allowed here, only a `glob` of the current path (which must then of course be a directory, or a single file matching the glob pattern).

dissect/target/loaders/log.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _map_entry(self, entry: Path) -> str:
```
Schamper, line 18: I'm not sure this warrants a separate function. The individual functions are small enough that I feel like it makes it a bit harder to follow right now.

dissect/target/loaders/log.py
```diff
@@ -0,0 +1,27 @@
...
+        logs_dir_key = entry.suffix[1:].capitalize()
```
Schamper, line 19: This is of course not guaranteed for possible future logs, perhaps an extension -> plugin map would make more sense? I think that would make this logic look a bit nicer too.

Schamper (changes requested):

dissect/target/loaders/log.py
cecinestpasunepipe: Good idea.

dissect/target/loaders/log.py
```diff
@@ -7,21 +7,22 @@
...
         self.target = target
         vfs = VirtualFilesystem()
-        for entry in self.path.parent.rglob(self.path.name):
-            vfs.map_file(str(self._map_entry(entry)), str(entry))
+        for entry in self.path.parent.glob(self.path.name):
+            ext = entry.suffix.lower()
+            mapping = self.LOGS_DIRS.get(ext, None)
+            if mapping is None:
+                continue
+            mapping = str(Path(mapping).joinpath(entry.name)).lower()
```
Schamper, line 26: Is the `lower()` necessary?

dissect/target/loaders/log.py
```diff
@@ -7,21 +7,22 @@
...
-
     def map(self, target: Target) -> None:
         self.target = target
         vfs = VirtualFilesystem()
-        for entry in self.path.parent.rglob(self.path.name):
-            vfs.map_file(str(self._map_entry(entry)), str(entry))
+        for entry in self.path.parent.glob(self.path.name):
+            ext = entry.suffix.lower()
+            mapping = self.LOGS_DIRS.get(ext, None)
+            if mapping is None:
```
Schamper, line 24:
```suggestion
            if (mapping := self.LOGS_DIRS.get(ext, None)) is None:
```
Or not a fan of walrus operator? :P

Schamper (changes requested):

dissect/target/loaders/log.py
cecinestpasunepipe: Yes, sometimes those files are named .Evt :-(

dissect/target/loaders/log.py
cecinestpasunepipe: Absolutely, big fan! Did not know that people like to call it that. Have to look real close to see it though...

dissect/target/loaders/log.py
Schamper: Would it not be better to make the `VirtualFilesystem` be `case_sensitive=False` in this case? Since we're (at least in the current scenario) exclusively emulating a Windows filesystem.

cecinestpasunepipe: Unfortunately that does not seem to work because the plugin internally uses glob() which even on a case insensitive fs will take casing into account. Overriding glob for case insensitive filesystems may work but may break things.

Schamper: That sounds like a bug in glob that should be fixed. It should be case-insensitive on case-insensitive filesystems and case-sensitive on case-sensitive filesystems.

cecinestpasunepipe: I don't think that is quite a trivial thing to do (https://stackoverflow.com/questions/8151300/ignore-case-in-glob-on-linux) and it might break stuff that accidentally depends on this behaviour. Doesn't this warrant a separate issue?

Schamper: But this is about the case-sensitiveness of `VirtualFilesystem` right? The `fs.glob()` should already be case-insensitive: https://github.com/fox-it/dissect.target/blob/main/dissect/target/helpers/fsutil.py#L919-L923 and the `TargetPath.glob()` should also already be case-insensitive: https://github.com/fox-it/dissect.target/blob/main/dissect/target/helpers/fsutil.py#L287

cecinestpasunepipe: Right, I see that this is a simple error in my code after all. Thanks for your guidance.

dissect/target/loader.py
```diff
@@ -194,6 +194,7 @@ def open(item: Union[str, Path], *args, **kwargs):
 register("tanium", "TaniumLoader")
 register("itunes", "ITunesLoader")
 register("target", "TargetLoader")
+register("log", "LogLoader")
```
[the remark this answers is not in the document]
cecinestpasunepipe: EvtLogLoader only covers the current functionality (evt + evtx), this loader can easily be extended to parse other logs as well (systemd journal files, etl etc). I first named this loader "FileLoader" but after careful inspection I decided I could narrow it down to LogLoader because logs are most likely to exist "outside" a target (registry hives are often integrated with the target, so they are less likely to exist on their own).

Schamper (approved):
Will you also be adding documentation on usage to dissect-docs?

PR #242, opened 2023-04-25
Add Journal Systemd plugin

### LINKED ISSUE #161: Systemd Journal parser
Systemd Journal parser

## References

[https://wiki.archlinux.org/title/Systemd/Journal](https://wiki.archlinux.org/title/Systemd/Journal)
[https://github.com/log2timeline/plaso/blob/main/plaso/parsers/systemd_journal.py](https://github.com/log2timeline/plaso/blob/main/plaso/parsers/systemd_journal.py)

Zawadidone:
Closes #161


.gitattributes
```diff
@@ -5,3 +5,4 @@ tests/data/plugins/browsers/firefox/places.sqlite filter=lfs diff=lfs merge=lfs
 tests/data/plugins/browsers/chrome/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/edge/History.sqlite filter=lfs diff=lfs merge=lfs -text
 tests/data/plugins/browsers/chromium/History.sqlite filter=lfs diff=lfs merge=lfs -text
+tests/data/plugins/os/unix/log/journal filter=lfs diff=lfs merge=lfs -text
```
cecinestpasunepipe, line 8: I see that we are not really giving a good example here but would it be possible to make a journal file for testing purposes? Since you seem to only use 1 record 8 MB seems a bit much. Also, are you sure no personal data is in the binary blob? I assume it comes from your working station? If you can make a test-only systemd journal it also does not have to go to LFS (probably).

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,388 @@
...
+                        elif data_object.flags == c_journal.ObjectFlag.OBJECT_COMPRESSED_LZ4:
+                            data = lz4.decompress(data[8:])
+                        elif data_object.flags == c_journal.ObjectFlag.OBJECT_COMPRESSED_ZSTD:
+                            data = zstandard.decompress(data)
+
+                        key, value = self.decode_value(data)
+                        event[key] = value
+
+                except Exception:
+                    continue
```
cecinestpasunepipe, line 294: Just a question...

Not really a systemd-journal expert but I see this Exception here is to catch invalid offsets, as in... (sample):

```
<EntryItem object_offset=0x3905a0, hash=0x64fb679e5714f4d>
3736992
<EntryItem object_offset=0x390640, hash=0x6fbd0df70f2eae8c>
3737152
<EntryItem object_offset=0x3906e8, hash=0x180e145321813b2b>
3737320
<EntryItem object_offset=0x6, hash=0x38>
6
<EntryItem object_offset=0x391a58, hash=0x390780>
3742296
<EntryItem object_offset=0x390cf0, hash=0x390f28>
3738864
<EntryItem object_offset=0x391648, hash=0x1>
3741256
<EntryItem object_offset=0x67, hash=0x4a45e1aeed5a0fb2>
103
<EntryItem object_offset=0x0, hash=0x0>
0
<EntryItem object_offset=0x390cf0, hash=0x0>
3738864
<EntryItem object_offset=0x1, hash=0x5f454352554f535f>
1
<EntryItem object_offset=0x494e4f544f4e4f4d, hash=0x5453454d49545f43>
5282246636478549837
[Errno 22] Invalid argument
```

The last one seems a negative number at first glance.. Is this normal? I could not find anything about this in the systemd docs, it just says all offsets are 64bit unsigned integers in LE format (https://www.freedesktop.org/wiki/Software/systemd/journal-files/).

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,388 @@
...
+                offset = entry_array_object.next_entry_array_offset
+
+    def decode_value(self, value: bytes) -> tuple[str, str]:
+        value = value.decode().strip()
+
+        # Remove underscores _, __
+        if value[0:1] == "__":
+            value = value[2:]
+        elif value[0] == "_":
+            value = value[1:]
```
pyrco, line 250:
```suggestion
       value = value.lstrip("_")
```
Would this be an option or is it possible to have say 3 `_`'s and the 3rd shouldn't be stripped?

pyrco (changes requested):

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,388 @@
...
+                if not journal.signature == self.JOURNAL_SIGNATURE:
+                    self.target.log.warning("The Journal log file %s has an invalid magic header", f.path)
+                    continue
+
+                for entry in journal:
+                    yield JournalRecord(
+                        ts=entry.get("ts"),
+                        message=entry.get("message"),
+                        message_id=entry.get("message_id"),
+                        priority=int(entry.get("priority")) if entry.get("priority") else None,
```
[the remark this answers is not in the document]
Zawadidone: Yes is this something that should be part of `dissect/target/tools/utils.py`? Because I assume other plugins could also use that function?

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: That is not case, the suggested function solves the current two cases.

dissect/target/plugins/os/unix/log/journal.py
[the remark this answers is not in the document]
pyrco: For now I'd say just put it here. Once it is used elsewhere we can extract it out and put it in utils.py.

dissect/target/plugins/os/unix/log/journal.py
cecinestpasunepipe: Might have something to do with HEADER_INCOMPATIBLE_COMPACT?

cecinestpasunepipe: After all entries is a union... https://systemd.io/JOURNAL_FILE_FORMAT/

cecinestpasunepipe: Some quick research...

```
struct EntryItemCompact {
    le32_t object_offset;
    le32_t nope1;
    le64_t nope2;
};

struct EntryObject {
    ....
    EntryItemCompact items[size / 64];
};
```

seems to be working for compact entries for me (no more exceptions).

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,385 @@
...
+    def journal(self) -> Iterator[JournalRecord]:
+        """Return the content of Systemd Journal log files.
+
+        References:
+            - https://wiki.archlinux.org/title/Systemd/Journal
+            - https://github.com/systemd/systemd/blob/9203abf79f1d05fdef9b039e7addf9fc5a27752d/man/systemd.journal-fields.xml
+        """  # noqa: E501
+
+        for _path in self.journal_paths:
+            for f in self.target.fs.listdir_ext(_path):
```
cecinestpasunepipe, line 321: This gives an error on items like:

/run/systemd/journal/flushed

(at least on my system)


maybe try:

`filter(lambda x: self.target.fs.is_dir(x), self.journal_paths)` ?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: For me also no exceptions on two different system, I left the `Exception` in the code in the case the format is updated which could result in errors.

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -377,8 +381,8 @@ def journal(self) -> Iterator[JournalRecord]:
                         kernel_device=entry.get("kernel_device"),
                         kernel_subsystem=entry.get("kernel_subsystem"),
                         udev_sysname=entry.get("udev_sysname"),
-                        udev_devnode=path.from_posix(entry.get("udev_devnode")) if entry.get("udev_devnode") else None,
-                        udev_devlink=path.from_posix(entry.get("udev_devlink")) if entry.get("udev_devlink") else None,
+                        udev_devnode=get_optional(entry.get("udev_devnode"), path.from_posix),
+                        udev_devlink=get_optional(entry.get("udev_devlink"), path.from_posix),
                         journal_hostname=entry.get("hostname"),
                         filepath=path.from_posix(f.path),
```
pyrco, line 387:
```suggestion
                        filepath=f.path,
```
This is now allowed with the latest stable `flow.record`

pyrco (changes requested):

dissect/target/plugins/os/unix/log/journal.py
cecinestpasunepipe: Can you add logging to this exception? It would be nice for everyone to see if anything goes wrong.

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,389 @@
...
+    ObjectType  type;
+    uint8_t     flags;
+    uint8_t     reserved[6];
+    le64_t      size;
+    le64_t      seqnum;
+    le64_t      realtime;
+    le64_t      monotonic;
+    sd_id128_t  boot_id;
+    le64_t      xor_hash;
+    EntryItem   items[size / 64];
```
cecinestpasunepipe, line 188: Should this not be 32 or 64 depending on HEADER_INCOMPATIBLE_COMPACT?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
cecinestpasunepipe: Thi issue does not seem to be resolved.

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: Is this only not allowed for the field `filepath` or also for `udev_devnode` and `udev_devlink`. If so could a `DeprecationWarning` be introduced if this function is used?

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,393 @@
...
+                            data = lz4.decompress(data[8:])
+                        elif data_object.flags == c_journal.ObjectFlag.OBJECT_COMPRESSED_ZSTD:
+                            data = zstandard.decompress(data)
+
+                        key, value = self.decode_value(data)
+                        event[key] = value
+
+                except Exception as e:
+                    self.target.log.warning(
+                        "The data object in Journal file %s could not be parsed", self.fh.name, exc_info=e
```
cecinestpasunepipe, line 298: Thanks, this is very nice. Unfortunately self.fh.name won't always work, since fh can be a RunlistStream (for instance if you open a .vbox). So maybe see if fh has a name attr first?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
cecinestpasunepipe: I tried to reproduce the error for you but I now get an EOError on /run/systemd/journal/streams/8:356847. Point is, there seem to be all kinds of objects in the journal folders, maybe we need to test if it's a file first? Maybe even look for the .journal extension?

dissect/target/plugins/os/unix/log/journal.py
Schamper: Any `path` flow.record type can now also correctly be instantiated directly from a compatible `Path`-like object (of which `TargetPath` is one). The `from_posix`/`from_windows` functions are still valid when needing to instantiate from e.g. a string.

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: I have added a filter for known extensions

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: Is adding additional c-struct the preferred way of doing this?

cecinestpasunepipe: Yes, I would use double c-structures. That looks like the easiest way to go.

I think something like:

```
 if self.header.incompatible_flags == c_journal.IncompatibleFlag.HEADER_INCOMPATIBLE_COMPACT:
   entry_array_object = c_journal.EntryArrayObject_Compact(self.fh)
else:
   entry_array_object = c_journal.EntryArrayObject(self.fh)
```

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,457 @@
...
+
+        return key, value
+
+    def __iter__(self) -> Iterator[Instance]:
+        "Iterate over the entry objects to read payloads."
+
+        for offset in self.entry_object_offsets():
+            self.fh.seek(offset)
+
+            if self.header.incompatible_flags == c_journal.IncompatibleFlag.HEADER_INCOMPATIBLE_COMPACT:
```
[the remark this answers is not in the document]
cecinestpasunepipe: Maybe better use a cstruct Flag?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,457 @@
...
+    ARCHIVED  =  2,
+    UNKNOWN
+};
+
+union sd_id128_t {
+    uint8_t   bytes[16];
+    uint64_t  qwords[2];
+};
+
+enum IncompatibleFlag : le32_t {
```
cecinestpasunepipe, line 92:
```suggestion
flag IncompatibleFlag : le32_t {
```

dissect/target/plugins/os/unix/log/journal.py
Zawadidone: I have added that path to the TODO

dissect/target/plugins/os/unix/log/journal.py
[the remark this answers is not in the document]
cecinestpasunepipe: This suggestion does not seem to be implemented but the comment has been set to resolved without any explanation? I tried to run it again, but without this change it keeps giving errors. Do I miss something or did the change not come through?

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,457 @@
...
+            for item in entry.items:
+                try:
+                    self.fh.seek(item.object_offset)
+
+                    object = c_journal.ObjectHeader(self.fh)
+
+                    if object.type == c_journal.ObjectType.OBJECT_DATA:
+                        self.fh.seek(item.object_offset)
+
+                        if self.header.incompatible_flags == c_journal.IncompatibleFlag.HEADER_INCOMPATIBLE_COMPACT:
```
cecinestpasunepipe, line 340:
```suggestion
                        if self.header.incompatible_flags & c_journal.IncompatibleFlag.HEADER_INCOMPATIBLE_COMPACT:
```

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/journal.py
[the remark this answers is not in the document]
Zawadidone: I missed this one

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -0,0 +1,457 @@
...
+
+                for entry_object_offset in entry_array_object.entry_object_offsets:
+                    # Check if the offset is not zero and points to nothing
+                    if entry_object_offset:
+                        yield entry_object_offset
+
+                offset = entry_array_object.next_entry_array_offset
+
+    def decode_value(self, value: bytes) -> tuple[str, str]:
+        value = value.decode().strip()
```
cecinestpasunepipe, line 307:
```suggestion
        value = value.decode(encoding="utf-8", errors="surrogateescape").strip()
```
If we do this change, certain errors can be prevented a we can parse more data:

image

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):
LGTM

pyrco (approved):

PR #245, opened 2023-04-26
Fix LVM2 volume names containing hyphens

Schamper:
(DIS-1982)


dissect/target/plugins/os/unix/_os.py
```diff
@@ -315,7 +315,7 @@ def parse_fstab(
         if dev.startswith(("/dev/mapper", "/dev/gpt")):
             volume_name = dev.rsplit("/")[-1]
         elif dev.startswith("/dev/") and dev.count("/") == 3:
-            volume_name = "-".join(dev.rsplit("/")[-2:])
+            volume_name = "-".join(part.replace("-", "--") for part in dev.rsplit("/")[-2:])
```
cecinestpasunepipe, line 318: I think this warrants a comment, i.e. what we discussed.

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #246, opened 2023-05-02
Expand adpolicy plugin to include missing GPO info

cecinestpasunepipe:
(DIS-1745)


dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -62,6 +63,29 @@ def check_compatible(self):
...
+                    yield ADPolicyRecord(
+                        last_modification_time=task_file_stat.st_mtime,
+                        last_access_time=task_file_stat.st_atime,
+                        creation_time=task_file_stat.st_ctime,
+                        guid=task.attrib.get("uid"),
+                        key="XML",
+                        value=properties.attrib.get("appName"),
+                        size=len(xml),
+                        data=xml,
+                        path=str(task_file),
```
Schamper, line 83:
```suggestion
                        path=task_file,
```

dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -62,6 +63,29 @@ def check_compatible(self):
         if len([d for d in self.dirs if d.exists()]) <= 0:
             raise UnsupportedPluginError("No AD policy directories found")
 
+    def _xmltasks(self, policy_dir):
+        for task_file in policy_dir.rglob("ScheduledTasks.xml"):
+            try:
+                task_file_stat = task_file.stat()
+                xml = task_file.open("rt").read()
```
Schamper, line 70:
```suggestion
                xml = task_file.read_text()
```

Schamper (changes requested):

tests/data/adpolicy/{CC63F200-7309-4ba0-B154-A71CD118DBCC}/USER/Preferences/ScheduledTasks/ScheduledTasks.xml
Schamper, line 1: Can you also move this file (and possibly the other ones) to `tests/data/plugins/os/windows/adpolicy`?

dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -62,6 +63,29 @@ def check_compatible(self):
...
+                    properties = task.find("Properties")
+                    yield ADPolicyRecord(
+                        last_modification_time=task_file_stat.st_mtime,
+                        last_access_time=task_file_stat.st_atime,
+                        creation_time=task_file_stat.st_ctime,
+                        guid=task.attrib.get("uid"),
+                        key="XML",
+                        value=properties.attrib.get("appName"),
+                        size=len(xml),
+                        data=xml,
```
pyrco, line 82: Full XML data? If there are multiple tasks in an XML file I think it would be more convenient to have the XML of just that task.

dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -62,6 +63,29 @@ def check_compatible(self):
...
+                xml = task_file.open("rt").read()
+                tree = ElementTree.fromstring(xml)
+                for task in tree.findall("Task"):
+                    properties = task.find("Properties")
+                    yield ADPolicyRecord(
+                        last_modification_time=task_file_stat.st_mtime,
+                        last_access_time=task_file_stat.st_atime,
+                        creation_time=task_file_stat.st_ctime,
+                        guid=task.attrib.get("uid"),
+                        key="XML",
```
pyrco, line 79: Would the Task's `clsid` be a good candidate for the `key`? Or is that too far off the mark @Schamper?

cecinestpasunepipe: From a database perspective I think the best solution is to use a simple code marker like "XML". I don't think that clsid are that simple to filter with a query. But this is just my assumption.

dissect/target/plugins/os/windows/adpolicy.py
cecinestpasunepipe: This won't work, get a decoding error.

dissect/target/plugins/os/windows/adpolicy.py
cecinestpasunepipe: Fixed. Good point.

tests/data/adpolicy/{CC63F200-7309-4ba0-B154-A71CD118DBCC}/USER/Preferences/ScheduledTasks/ScheduledTasks.xml
cecinestpasunepipe: Solved. Makes sense. Clean structure.

dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -62,6 +63,29 @@ def check_compatible(self):
...
+                        value=properties.attrib.get("appName"),
+                        size=len(xml),
```
pyrco, line 81: Shouldn't the size then not also be the size of the task data below?

pyrco (changes requested):

dissect/target/plugins/os/windows/adpolicy.py
cecinestpasunepipe: Good point. Oversight.

dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -78,7 +78,7 @@ def _xmltasks(self, policy_dir):
                         guid=task.attrib.get("uid"),
                         key="XML",
                         value=properties.attrib.get("appName"),
-                        size=len(xml),
+                        size=len(task),
                         data=ElementTree.tostring(task),
```
pyrco:
```suggestion
                        task_data = ElementTree.tostring(task)
                        ...
                        size=len(task_data),
                        data=ElementTree.tostring(task_data),
```

Ithink this would be a bit more in line with the .pol file records.

pyrco (changes requested):

pyrco (approved):

pyrco (approved):

pyrco (approved):

dissect/target/plugins/os/windows/adpolicy.py
Schamper: The `path` field should be changed to be a `path` type instead of `uri` what it is now.

cecinestpasunepipe: Ok done. Thought it was uri for reasons.

PR #252, opened 2023-05-15
Fix services plugin failures on linux targets

Miauwkeru:
`find_plugin_functions` legacy lookup didn't take into account that in some cases multiple plugins have the same name.
e.g, for `services` we have:
  - windows.services
  - unix.services

This caused it to return the plugin it detected last, which was the windows `services` plugin in this case.
Now, multiple plugins with the same name get returned, and if it executes correctly, the other plugins with
the same name get skipped.

(DIS-1885)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -831,7 +831,9 @@ def plugin_function_index(target: Target) -> tuple[dict[str, Any], set[str]]:
     rootset = set()
 
     def all_plugins():
-        yield from plugins()
+        # Filter out plugins based on the target os
+        os_type = type(target._os) if target._os else None
+        yield from plugins(os_type)
         yield from os_plugins()
         yield from child_plugins()  # Doesn't export anything but added for completeness.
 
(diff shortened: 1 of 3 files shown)


dissect/target/plugin.py
```diff
@@ -920,27 +920,28 @@ def find_plugin_functions(target: Target, patterns: str, compatibility: bool = F
...
                         class_object=loaded_plugin_object,
                         method_name=funcname,
                         output_type=getattr(fobject, "__output__", "text"),
-                        plugin_desc=plugindesc,
+                        plugin_desc=description,
                     )
                 )
-    return result
+
+    return list(set(result))
```
pyrco, line 947: How would plugins end up multiple times in result? Or did e.g. `windows.services` and `unix.services` get the exact same `PluginFunction` representation?

pyrco (changes requested):
I have the feeling there is a bigger underlying issue with our plugin lookup/execution.

It seems we use just the final plugin instance/function name for uniqueness leading to the issues we are seeing. And secondly it's a bit surprising that plugins beneath another os'es tree get executed at all.

This solution may still fail because of that if for any reason another os'es plugin does not fail somehow.

Do you see a smart way of not going into the plugin tree of other os'es somehow without piling up even more code?

pyrco (approved):

PR #255, opened 2023-05-16
Handle keyboard interrupts and os errors cleanly for dissect tools

Miauwkeru:
Add sigpipe handling

(DIS-1865)

### DIFF
--- a/dissect/target/helpers/lazy.py
+++ b/dissect/target/helpers/lazy.py
@@ -44,22 +44,26 @@ class LazyAttr:
         self.module = module
         self._realattr = None
 
-    def __call__(self, *args, **kwargs):
+    @property
+    def realattr(self):
         if not self._realattr:
-            self.module._import()  # noqa
-            self._realattr = getattr(self.module._module, self.attr)  # noqa
+            self.module._import()
+            self._realattr = getattr(self.module._module, self.attr)
 
-        return self._realattr(*args, **kwargs)
+        return self._realattr
+
+    @property
+    def __doc__(self):
+        return self.realattr.__doc__
+
+    def __call__(self, *args, **kwargs):
+        return self.realattr(*args, **kwargs)
 
     def __getattr__(self, attr):
-        if not self._realattr:
-            self.module._import()  # noqa
-            self._realattr = getattr(self.module._module, self.attr)  # noqa
-
-        return getattr(self._realattr, attr)
+        return getattr(self.realattr, attr)
 
     def __repr__(self):
-        return f"<lazyattr {self.module._module_name}.{self.attr} loaded={self._realattr is not None}>"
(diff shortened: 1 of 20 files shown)


dissect/target/tools/utils.py
```diff
@@ -196,3 +198,24 @@ def persist_execution_report(output_dir: Path, report_data: Dict, timestamp: dat
     report_full_path = output_dir / report_filename
     report_full_path.write_text(json.dumps(report_data, sort_keys=True, indent=4))
     return report_full_path
+
+
+def catch_sigpipe(func: Callable) -> Callable:
```
cecinestpasunepipe, line 203: Not sure if this implementation is correct, what is the practical use case?

According to the docs I found (http://www.pixelbeat.org/programming/sigpipe_handling.html), the CLI result for:

`target-query -l | head -n1`

should be:

`Available plugins:`

(like:  yes 1234 | head -n1)

But the actual output seems to be:

```
Available plugins:
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```
Not sure if that is fixable or if we even want to change that.
So, can you help me understand, what is the practical value of this change?

cecinestpasunepipe (dismissed):
More a bit of a discussion topic I think...

dissect/target/tools/utils.py
```diff
@@ -196,3 +198,24 @@ def persist_execution_report(output_dir: Path, report_data: Dict, timestamp: dat
...
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except KeyboardInterrupt:
+            print("Aborted!", file=sys.stderr)
+            return 1
+        except OSError as e:
+            exc_type = type(e)
+            # Only catch BrokenPipeError or OSError 22
+            if (exc_type is BrokenPipeError) or (exc_type is OSError and e.errno == 22):
```
pyrco, line 216: Maybe nice to use the `errno.EINVAL` instead of the magic number 22?

dissect/target/tools/utils.py
pyrco: If you want to get rid of those messages, apparently you have to close `sys.stderr` just before exiting https://bugs.python.org/issue11380,#msg198345

The idea was to not have backtraces or extraneous output confusing the user when dealing with an external situation where we know things are "broken" and we know we should stop the whole process anyway.

pyrco (changes requested):

yunzheng:
Please also see a refactored version of this decorator in `dissect.cobaltstrike`:

https://github.com/fox-it/dissect.cobaltstrike/blob/4bbe48f8847177fa76e346eca00344f681079e64/dissect/cobaltstrike/utils.py#L92

It looks like I forgot to merge these changes in `flow.record` as well.

pyrco (approved):

PR #259, opened 2023-05-23
Refine plugin system

cecinestpasunepipe:
- make plugin system compatible with plugins that require output files/dirs by checking output types early (None)
- allow users to get an OS summary in records by referencing the OS header in the -l output (-f OS)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -843,8 +843,8 @@ def plugin_function_index(target: Target) -> tuple[dict[str, Any], set[str]]:
         modulepath = available["module"]
         if modulepath.endswith("._os"):
             if not target._os:
-                # if no target available mention general OS instead of specific one
-                available["module"] = "OS"
+                # if no target available add a namespaceless section
+                available["module"] = ""
             elif target._os.__class__.__name__ != available["class"]:
                 continue
             rootset.add(modulepath.split(".")[0])
--- a/dissect/target/plugins/general/osinfo.py
+++ b/dissect/target/plugins/general/osinfo.py
@@ -1,0 +1,39 @@
+from typing import Callable, Generator
+
+from flow.record import GroupedRecord, Record
+
+from dissect.target import plugin
+from dissect.target.exceptions import PluginError
+from dissect.target.helpers.record import TargetRecordDescriptor
+
+OSInfoRecord = TargetRecordDescriptor(
+    "generic/osinfo",
+    [
+        ("string", "name"),
+        ("string", "value"),
+    ],
+)
+
+
+class OSInfoPlugin(plugin.Plugin):
+    """Convenience plugin that wraps _os.* functions in records."""
+
+    def check_compatible(self) -> bool:
+        return True
+
+    @plugin.export(record=OSInfoRecord)
+    def osinfo(self) -> Generator[Record, None, None]:
+        for os_func in self.target._os.__functions__:
+            if os_func in ["is_compatible", "get_all_records"]:
+                continue
+            value = getattr(self.target._os, os_func)
+            record = OSInfoRecord(name=os_func, value=repr(value), _target=self.target)
+            yield record
+            if isinstance(value, Callable):
+                try:
+                    for subvalue in value():
+                        yield GroupedRecord("generic/osinfo/grouped", [record, subvalue])
+                except (PluginError, TypeError):
+                    # Ignore exceptions triggered by functions
+                    # that cannot be executed in this context
+                    continue
--- a/dissect/target/plugins/general/osinfo.py
+++ b/dissect/target/plugins/general/osinfo.py
@@ -1,4 +1,4 @@
-from typing import Callable, Generator
+from typing import Callable, Generator, Iterator, Union
 
 from flow.record import GroupedRecord, Record
 
@@ -22,7 +22,7 @@ class OSInfoPlugin(plugin.Plugin):
         return True
 
     @plugin.export(record=OSInfoRecord)
-    def osinfo(self) -> Generator[Record, None, None]:
+    def osinfo(self) -> Iterator[Union[OSInfoRecord, GroupedRecord]]:
         for os_func in self.target._os.__functions__:
             if os_func in ["is_compatible", "get_all_records"]:
                 continue
@@ -30,9 +30,8 @@ class OSInfoPlugin(plugin.Plugin):
             record = OSInfoRecord(name=os_func, value=None, _target=self.target)
             if isinstance(value, Callable) and isinstance(subrecords := value(), Generator):
                 try:
-                    yield record
                     yield GroupedRecord("generic/osinfo/grouped", [record] + list(subrecords))
-                except (PluginError, TypeError):
+                except Exception:
                     # Ignore exceptions triggered by functions
                     # that cannot be executed in this context
                     continue
--- a/dissect/target/plugins/general/osinfo.py
+++ b/dissect/target/plugins/general/osinfo.py
@@ -1,9 +1,8 @@
 from typing import Callable, Generator, Iterator, Union
 
-from flow.record import GroupedRecord, Record
+from flow.record import GroupedRecord
 
 from dissect.target import plugin
-from dissect.target.exceptions import PluginError
 from dissect.target.helpers.record import TargetRecordDescriptor
 
 OSInfoRecord = TargetRecordDescriptor(
--- a/dissect/target/plugins/general/osinfo.py
+++ b/dissect/target/plugins/general/osinfo.py
@@ -27,13 +27,15 @@ class OSInfoPlugin(plugin.Plugin):
             if os_func in ["is_compatible", "get_all_records"]:
                 continue
             value = getattr(self.target._os, os_func)
-            record = OSInfoRecord(name=os_func, value=repr(value), _target=self.target)
-            yield record
-            if isinstance(value, Callable):
+            record = OSInfoRecord(name=os_func, value=None, _target=self.target)
+            if isinstance(value, Callable) and isinstance(subrecords := value(), Generator):
                 try:
-                    for subvalue in value():
-                        yield GroupedRecord("generic/osinfo/grouped", [record, subvalue])
+                    yield record
+                    yield GroupedRecord("generic/osinfo/grouped", [record] + list(subrecords))
                 except (PluginError, TypeError):
                     # Ignore exceptions triggered by functions
                     # that cannot be executed in this context
                     continue
+            else:
+                record.value = value
+                yield record
--- a/dissect/target/plugins/general/plugins.py
+++ b/dissect/target/plugins/general/plugins.py
@@ -87,7 +87,7 @@ def get_description_dict(
     output_descriptions = []
     for key in structure_dict.keys():
         output_descriptions += [
-            textwrap.indent(key + ":", prefix=" " * indentation_step) if key != "" else "OS plugins (no namespace)"
+            textwrap.indent(key + ":", prefix=" " * indentation_step) if key != "" else "OS plugins"
         ] + output_plugin_description_recursive(
             structure_dict[key],
             print_docs,
--- a/dissect/target/plugins/general/plugins.py
+++ b/dissect/target/plugins/general/plugins.py
@@ -87,7 +87,7 @@ def get_description_dict(
     output_descriptions = []
     for key in structure_dict.keys():
         output_descriptions += [
-            textwrap.indent(key + ":", prefix=" " * indentation_step)
+            textwrap.indent(key + ":", prefix=" " * indentation_step) if key != "" else "OS plugins (no namespace)"
         ] + output_plugin_description_recursive(
             structure_dict[key],
             print_docs,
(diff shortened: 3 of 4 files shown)


cecinestpasunepipe:
> do not forget to remove the references to `os` in `dissect-docs`
> 
>     * https://docs.dissect.tools/en/latest/usage/introduction.html
> 
>     * https://github.com/fox-it/dissect-docs/blob/main/docs/source/usage/introduction.rst

No longer necessary, proposal for a better solution

cecinestpasunepipe:
> > > do not forget to remove the references to `os` in `dissect-docs`
> > > ```
> > > * https://docs.dissect.tools/en/latest/usage/introduction.html
> > > 
> > > * https://github.com/fox-it/dissect-docs/blob/main/docs/source/usage/introduction.rst
> > > ```
> > 
> > 
> > No longer necessary, proposal for a better solution
> 
> It would probably still be necessary.. As the docs refer to it in combination with `hostname`, `domain` etc.
> 
> As an example line:
> 
> > target-query /mnt/SCHARDT.001 -f hostname,domain,os,version,ips -d ';'
> 
> Which would fail due to the mixture of records and `strings` in the output
> 
> Edit: the introduction was working alright. it went wrong in. Where it uses `OS` instead of `os`.
> 
> > https://docs.dissect.tools/en/latest/usage/use-cases.html

Yes, the docs already have been updated I see?

dissect/target/plugin.py
```diff
@@ -870,6 +870,12 @@ def find_plugin_functions(target: Target, patterns: str, compatibility: bool = F
     functions, rootset = plugin_function_index(target)
 
     for pattern in patterns.split(","):
+        # OS is used as a header in the -l output of CLI
+        # this is not a valid plugin pattern but let's be friendly to the user and
+        # give all OS properties as records (osinfo is a little wrapper) (users are excluded because those are records)
+        if pattern.startswith("OS"):
```
Schamper, line 876: Perhaps consider also allowing lowercase `os`?

dissect/target/plugins/general/osinfo.py
```diff
@@ -0,0 +1,36 @@
...
+from flow.record import Record
+
+from dissect.target import plugin
+from dissect.target.exceptions import PluginError
+from dissect.target.helpers.record import TargetRecordDescriptor
+
+OSInfoRecord = TargetRecordDescriptor(
+    "generic/osinfo",
+    [
+        ("string", "prop"),
```
Schamper, line 12:
```suggestion
        ("string", "name"),
```
Perhaps more clear than `prop`? Or `key`.

dissect/target/plugins/general/osinfo.py
```diff
@@ -0,0 +1,36 @@
...
+            value = getattr(self.target._os, os_func)
+            # Aggregate functions that yield records of their own
+            if isinstance(value, Callable):
+                try:
+                    value = len(list(value()))
+                except (PluginError, TypeError):
+                    # Ignore exceptions triggered by functions
+                    # that cannot be executed in this context
+                    continue
+            yield OSInfoRecord(prop=os_func, value=value)
```
Schamper, line 36:
```suggestion
            yield OSInfoRecord(prop=os_func, value=value, _target=self.target)
```
Or was that deliberate?

dissect/target/tools/query.py
```diff
@@ -174,13 +174,30 @@ def main():
         parser.error("argument -f/--function is required")
 
     # Verify uniformity of output types, otherwise default to records
+    # note that this is a hueristic, the targets are not opened yet because of
```
Schamper, line 177:
```suggestion
    # Note that this is a heuristic, the targets are not opened yet because of
```

dissect/target/tools/query.py
```diff
@@ -174,13 +174,30 @@ def main():
...
+    # performance, so it might generate a false positive
+    # (os.* on Windows includes other OS plugins),
+    # however this is highly hypothetical, most plugins across OSes have
+    # the same output types and most output types are records anyway.
+    # Furthermore we really want the notification at the top, so this is the only
+    # way forward. In the very unlikely case you have a
+    # collection of non-record plugins that have record counterparts for
+    # other OSes just refine the wildcard to exclude other Oses.
```
Schamper, line 185:
```suggestion
    # other OSes just refine the wildcard to exclude other OSes.
```

dissect/target/tools/query.py
```diff
@@ -174,13 +174,30 @@ def main():
         parser.error("argument -f/--function is required")
 
     # Verify uniformity of output types, otherwise default to records
```
Schamper, line 176:
```suggestion
    # Verify uniformity of output types, otherwise default to records.
```

dissect/target/tools/query.py
```diff
@@ -174,13 +174,30 @@ def main():
...
+    # note that this is a hueristic, the targets are not opened yet because of
+    # performance, so it might generate a false positive
+    # (os.* on Windows includes other OS plugins),
+    # however this is highly hypothetical, most plugins across OSes have
+    # the same output types and most output types are records anyway.
+    # Furthermore we really want the notification at the top, so this is the only
+    # way forward. In the very unlikely case you have a
+    # collection of non-record plugins that have record counterparts for
+    # other OSes just refine the wildcard to exclude other Oses.
+    # The only scenario I can think of that might cause this is with
```
Schamper, line 186:
```suggestion
    # The only scenario that might cause this is with
```

dissect/target/tools/query.py
```diff
@@ -174,13 +174,30 @@ def main():
...
 
     default_output_type = None
-    if len(output_types) > 1:
+
+    # Otherwise you see a confusing mixed output types message, can still generate output
+    # because some plugins have output file/dir options
+    if not len(output_types):
+        pass
+    elif len(output_types) > 1:
+        # give this warning beforehand, if mixed, set default to record (no errors)
```
Schamper, line 200:
```suggestion
        # Give this warning beforehand, if mixed, set default to record (no errors)
```

Schamper (dismissed):

dissect/target/plugin.py
cecinestpasunepipe: Not sure if that is a good idea. OS does not exist but "os" does. So this interferes with existing functionality.

Schamper: Is this not a very weird edge case to support in that case? I think the `osinfo` is a good idea, just wondering if it's good to really put a hard `if EDGE_CASE` for one specific header in some documentation, if the intended behavior between `OS` and `os` are different. `OS` should behave similar to `os` in that case, in my opinion, as casing shouldn't have this big of a difference.

dissect/target/plugins/general/osinfo.py
```diff
@@ -0,0 +1,36 @@
...
+        return True
+
+    @plugin.export(record=OSInfoRecord)
+    def osinfo(self) -> Generator[Record, None, None]:
+        for os_func in self.target._os.__functions__:
+            value = getattr(self.target._os, os_func)
+            # Aggregate functions that yield records of their own
+            if isinstance(value, Callable):
+                try:
+                    value = len(list(value()))
```
Schamper, line 31: Would the `repr` of the result not the a better representation of the result?

dissect/target/tools/query.py
```diff
@@ -173,14 +173,31 @@ def main():
...
     output_types = set()
     funcs = find_plugin_functions(Target(), args.function, False)
     for func in funcs:
         output_types.add(func.output_type)
 
     default_output_type = None
-    if len(output_types) > 1:
+
+    # Otherwise you see a confusing mixed output types message, can still generate output
+    # because some plugins have output file/dir options
```
pyrco, line 196: This only applies to the first `if` and not the whole block? Maybe add these comments then after the `if`

dissect/target/plugin.py
pyrco: Then we should just lower case all patterns (or do case insensitive match)? And keep `osinfo` available as its own `os.info` plugin or something like that.

pyrco (changes requested):

dissect/target/plugin.py
cecinestpasunepipe: Ok adjusted

dissect/target/plugins/general/osinfo.py
```diff
@@ -0,0 +1,39 @@
...
+    def check_compatible(self) -> bool:
+        return True
+
+    @plugin.export(record=OSInfoRecord)
+    def osinfo(self) -> Generator[Record, None, None]:
+        for os_func in self.target._os.__functions__:
+            if os_func in ["is_compatible", "get_all_records"]:
+                continue
+            value = getattr(self.target._os, os_func)
+            record = OSInfoRecord(name=os_func, value=repr(value), _target=self.target)
```
pyrco, line 30: I don't like the `repr(value)`  here as it adds extra quotes to values that don't need them. Looking at the output I noticed a few things that I think we should address:

```
$ target-query -q -f osinfo ./WinDev2007Eval-disk1.vmdk 
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='architecture' value="'amd64-win64'">
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='hostname' value="'WINDEV2007EVAL'">
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='ips' value="['172.17.19.136', '192.168.73.134', '10.0.2.15']">
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='os' value="'windows'">
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>
<generic/osinfo/grouped [<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>, <windows/user hostname='WINDEV2007EVAL' domain=None sid='S-1-5-18' name='systemprofile' home='%systemroot%\\system32\\config\\systemprofile'>]>
<generic/osinfo/grouped [<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>, <windows/user hostname='WINDEV2007EVAL' domain=None sid='S-1-5-19' name='LocalService' home='%systemroot%\\ServiceProfiles\\LocalService'>]>
<generic/osinfo/grouped [<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>, <windows/user hostname='WINDEV2007EVAL' domain=None sid='S-1-5-20' name='NetworkService' home='%systemroot%\\ServiceProfiles\\NetworkService'>]>
<generic/osinfo/grouped [<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>, <windows/user hostname='WINDEV2007EVAL' domain=None sid='S-1-5-21-2409144726-2152242204-271938582-1001' name='User' home='C:\\Users\\User'>]>
<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='version' value="'Windows 10 Enterprise Evaluation (NT 10.0) 19041.329'">
```

- The separate `<generic/osinfo hostname='WINDEV2007EVAL' domain=None name='users' value='<bound method WindowsPlugin.users of <dissect.target.plugins.os.windows._os.WindowsPlugin object at 0x7d3d6467f0a0>>'>` entry has little value I think.
- The way grouping turns out means there is a lot of duplication.

We can go at it in different ways. One way is to skip the grouping at all and output sub-records verbatim. That does mean the relation to osinfo is lost, but that may be ok.

Another way is to group all sub-records into a single group. This has as a downside that if the amount of subrecords gets very large, memory consumption will also increase. As `users` is currently the only plugin in `_os.py` that outputs records, this may not be a problem.

I think the best `value` for the `generic/osinfo` record in case it is a `Callable` should be empty as the `repr` of some function is not very useful. Also `Callable` can be a bit more refined to `Generator`, `isinstance` will return True for functions that yield values (as is the case with `users()`), but not for lists etc. (as is the case for `ips`).

pyrco (changes requested):

dissect/target/plugins/general/plugins.py
```diff
@@ -87,7 +87,7 @@ def get_description_dict(
     output_descriptions = []
     for key in structure_dict.keys():
         output_descriptions += [
-            textwrap.indent(key + ":", prefix=" " * indentation_step)
+            textwrap.indent(key + ":", prefix=" " * indentation_step) if key != "" else "OS plugins (no namespace)"
```
Schamper, line 90:
```suggestion
            textwrap.indent(key + ":", prefix=" " * indentation_step) if key != "" else "OS plugins"
```
I don't believe it's necessary to display this.

pyrco (approved):

pyrco (approved):

PR #263, opened 2023-05-31
Add target-query improvements

JSCU-CNI:
This PR adds the following:
- Add target-query test to assert no existing functions failed to import during runtime
- target-query now checks if all given functions exist
- Add target-query test to assert the tool exits when given non-existing functions as input


dissect/target/tools/query.py
```diff
@@ -179,6 +179,30 @@ def main():
     for func in funcs:
         output_types.add(func.output_type)
 
+    # Verify if all provided plugins exist. Exits if a
+    # given function was not found in `find_plugin_functions`.
+    # TODO: Are there any other ways to call a function using target-query?
+    invalid_funcs = [func.strip(" *!?[]") for func in args.function.split(",")]
```
cecinestpasunepipe, line 185: This will break somewhat more complex patterns like: `browsers.*.downloads` (i.e. give me all downloads)

cecinestpasunepipe (changes requested):

dissect/target/plugins/general/loaders.py
```diff
@@ -14,10 +12,8 @@ def check_compatible(self):
     def loaders(self):
         """List the available loaders."""
 
-        loaders = itertools.chain(LOADERS, [DirLoader])
-        loader_names = [loader.attr for loader in loaders]
-        loader_names.sort()
+        loader_names = sorted(getattr(loader, "attr", loader.__name__) for loader in LOADERS + [DirLoader])
```
pyrco, line 15: As the `attr` property is part of the "lazy loading" properties of these modules, using it is dubious to begin with. So replacing it all together with the property that is actually meant is not a problem:

```suggestion
        loader_names = sorted(loader.__name__ for loader in LOADERS + [DirLoader])
```

Note that the code on HEAD has already changed significantly, so this change will need some reworking.

dissect/target/plugins/general/loaders.py
```diff
@@ -14,10 +12,8 @@ def check_compatible(self):
...
 
         print("Available loaders:")
         for loader_name in loader_names:
-            print(f"    {loader_name}")
+            print(f"\t{loader_name}")
```
pyrco, line 19: All other output is 4 spaces indented, so this should too. For consistency, it is better to replace it with `dissect.target.helpers.docs.INDENT_STEP`:

```suggestion
from dissect.target.helpers.docs import INDENT_STEP

...

            print(f"{INDENT_STEP}{loader_name}")
```

pyrco (changes requested):

dissect/target/tools/query.py
JSCU-CNI: Added handling for patterns such as `a.*.b` in commit [e3e4dbd](https://github.com/fox-it/dissect.target/pull/263/commits/e3e4dbd68c29187a406d1e1af37febcfdb749cf5). 

Are there any other ways to call a function using target-query?

cecinestpasunepipe: Yes, many unfortunately because the plugin finder uses fnmatch. It might be a better approach to use a return parameter in find_plugin_functions(... missing_plugins=[]) and check that?

cecinestpasunepipe (changes requested):

dissect/target/tools/query.py
JSCU-CNI: Thanks for the pointer. Implemented your suggestion in af13998f7255e1dfbd948afa09ec008f7632c0f2.

dissect/target/plugin.py
```diff
@@ -859,7 +859,9 @@ def all_plugins():
     return index, rootset
 
 
-def find_plugin_functions(target: Target, patterns: str, compatibility: bool = False) -> list[PluginFunction]:
+def find_plugin_functions(
+    target: Target, patterns: str, compatibility: bool = False
```
pyrco, line 863:
```suggestion
    target: Target,
    patterns: str,
    compatibility: bool = False,
```

pyrco (changes requested):

dissect/target/tools/query.py
```diff
@@ -237,7 +211,7 @@ def main():
         first_seen_output_type = default_output_type
         cli_params_unparsed = rest
 
-        for func_def in find_plugin_functions(target, args.function, False):
+        for func_def, _ in find_plugin_functions(target, args.function, False):
```
cecinestpasunepipe, line 214:
```suggestion
        func_defs, _ = find_plugin_functions(target, args.function, False)

        for func_def in func_defs:
```
Otherwise we get a "too many values to unpack" error...

dissect/target/plugin.py
```diff
@@ -930,6 +934,9 @@ def find_plugin_functions(target: Target, patterns: str, compatibility: bool = F
                 if nsmatch or fmatch:
                     plugin_descriptions.append(func)
 
+            if not plugin_descriptions:
+                invalid_funcs.add(pattern)
+
```
cecinestpasunepipe, line 941: This will work for classic notation. For tree notation you can use something like this (I add a suggestion for you to apply directly unfortunately :-( )

```
if wildcard or treematch:
    matches = 0
    ...
    matches += 1
    result.append( .... PluginFunction( .....
    ...

if not matches:
    invalid_funcs.add(pattern)
```

JSCU-CNI: Thanks for the suggestion, applied and added test in [3f0eebd](https://github.com/fox-it/dissect.target/pull/263/commits/3f0eebdae6f18e0d6969661ac0da20e2970e477d).

cecinestpasunepipe (dismissed):
LGTM

cecinestpasunepipe (dismissed):
Checked again, nothing seems to have been changed.
Tested again, still seems to work.
LGTM.

dissect/target/plugin.py
```diff
@@ -889,6 +895,7 @@ def find_plugin_functions(target: Target, patterns: str, compatibility: bool = F
             pattern += "*"
 
         if wildcard or treematch:
+            matches = 0
```
pyrco, line 898: Why not use a boolean for this?

```suggestion
            matches = False
```

dissect/target/plugin.py
```diff
@@ -908,6 +915,7 @@ def find_plugin_functions(target: Target, patterns: str, compatibility: bool = F
                     except Exception:
                         continue
 
+                matches += 1
```
pyrco, line 918:
```suggestion
                matches = True
```

pyrco (approved):

PR #264, opened 2023-05-31
Fix assumption in Hyper-V child plugin

JSCU-CNI:
This PR fixes an assumption in the Hyper-V child plugin. The key `VirtualMachines` does not always exist in a Windows `data.vmcx` file.


dissect/target/plugins/child/hyperv.py
```diff
@@ -40,19 +40,21 @@ def check_compatible(self):
...
             data = hyperv.HyperVFile(self.data_vmcx.open()).as_dict()
+            if "VirtualMachines" not in data["Configurations"]:
+                return
 
             for vm_path in data["Configurations"]["VirtualMachines"].values():
                 yield ChildTargetRecord(
                     type=self.__type__,
                     path=path.from_windows(vm_path),
                     _target=self.target,
                 )
```
Schamper, line 60:
```suggestion
            if virtual_machines := data["Configurations"].get("VirtualMachines"):
                for vm_path in virtual_machines.values():
                    yield ChildTargetRecord(
                        type=self.__type__,
                        path=path.from_windows(vm_path),
                        _target=self.target,
                    )
```
Or something along those lines, then you can keep the other code where it was.

Schamper (changes requested):

dissect/target/plugins/child/hyperv.py
```diff
@@ -43,16 +43,17 @@ def list_children(self) -> Iterator[ChildTargetRecord]:
...
+                        type=self.__type__,
+                        path=path.from_windows(vm_path),
+                        _target=self.target,
+                    )
 
         for xml_path in self.vm_xml:
             yield ChildTargetRecord(
                 type=self.__type__,
-                path=xml_path.resolve(),
+                path=path.from_windows(xml_path.resolve()),
```
Schamper, line 57: I don't believe this to be necessary, at least not with the latest changes in `flow.record`.

tests/test_plugins_child_hyperv.py
```diff
@@ -18,6 +18,7 @@ def test_plugins_child_wsl(target_win, fs_win):
     children = list(target_win.list_children())
 
     assert len(children) == 5
+
```
Schamper, line 21:
```suggestion
```
No need to make changes to this file then :)

Schamper (changes requested):

dissect/target/plugins/child/hyperv.py
JSCU-CNI: Using `flow.record==3.10` removing the `path.from_windows` call causes the test to fail:

```
>       assert (
            str(children[4].path)
            == "\\sysvol\\ProgramData\\Microsoft\\Windows\\Hyper-V\\Virtual Machines\\B90AC31B-C6F8-479F-9B91-07B894A6A3F6.xml"  # noqa E501
        )
E       AssertionError: assert '/sysvol/Prog...894A6A3F6.xml' == '\\sysvol\\Pr...894A6A3F6.xml'
E         - \sysvol\ProgramData\Microsoft\Windows\Hyper-V\Virtual Machines\B90AC31B-C6F8-479F-9B91-07B894A6A3F6.xml
E         ? ^      ^           ^         ^       ^       ^                ^
E         + /sysvol/ProgramData/Microsoft/Windows/Hyper-V/Virtual Machines/B90AC31B-C6F8-479F-9B91-07B894A6A3F6.xml
E         ? ^      ^           ^         ^       ^       ^                ^

tests/test_plugins_child_hyperv.py:35: AssertionError
```

Schamper: Hm I'm confused now on the expected behaviour of this 😄 @pyrco can you jump in?

pyrco: That's interesting, I did some testing and debugging and it behaves as expected with flow.record 3.10 on my system and on the GitHub test runners, no need for `.from_windows()` here.

The behavior you describe is what you would get if you had a flow.record <= 3.9. I changed my tox py3 venv and explicitly installed flow.record==3.9 and saw the exact behavior you describe.

Perhaps tox did not detect properly it should update its py3 venv, this happens sometimes. If you could rerun the tests with `tox -re py3` and see if that changes anything. The `-r` will force the recreation of tox' py3 test venv.

JSCU-CNI: Thanks @pyrco. It seems tox's aggressive caching was the culprit here indeed.

Schamper (approved):

PR #269, opened 2023-06-05
Update the Phobos loader to use the new NTFS API

pyrco:
(DIS-1990)

### DIFF
--- a/dissect/target/loaders/phobos.py
+++ b/dissect/target/loaders/phobos.py
@@ -1,7 +1,15 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, BinaryIO, Iterator
+
 from dissect.util import stream
 
 from dissect.target import filesystem
 from dissect.target.loader import Loader
+
+if TYPE_CHECKING:
+    from dissect.target import Target
 
 BLOCK_SIZE = 64 * 0x100000  # 64 MiB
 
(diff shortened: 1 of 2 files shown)


dissect/target/loaders/phobos.py
```diff
@@ -39,10 +51,10 @@ class PhobosLoader(Loader):
     """
 
     @staticmethod
-    def detect(path):
+    def detect(path: Path):
```
Schamper, line 54:
```suggestion
    def detect(path: Path) -> bool:
```

dissect/target/loaders/phobos.py
```diff
@@ -39,10 +51,10 @@ class PhobosLoader(Loader):
...
         return path.suffix.lower() == ".eight"
 
-    def map(self, target):
+    def map(self, target: Target):
```
Schamper, line 57:
```suggestion
    def map(self, target: Target) -> None:
```

Schamper (dismissed):

dissect/target/loaders/phobos.py
```diff
@@ -39,10 +51,10 @@ class PhobosLoader(Loader):
     """
 
     @staticmethod
-    def detect(path):
+    def detect(path: Path):
         return path.suffix.lower() == ".eight"
```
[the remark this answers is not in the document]
pyrco: If we want to go that route it is better to transform it into a kind of `Scrape` loader, as there is nothing really specifically "Phobos" about this loader. It could be used as one of the fallback loaders.

I think that is a good idea, but maybe not in the context of this PR/ticket.

cecinestpasunepipe (approved):

PR #270, opened 2023-06-07
add handling for non-ascii characters

janstarke:
Some WER entries have fields with Non-identifier-characters in it. This codes handles the problem.

### DIFF
--- a/dissect/target/plugins/os/windows/wer.py
+++ b/dissect/target/plugins/os/windows/wer.py
@@ -15,6 +15,7 @@ camel_case_patterns = [re.compile(r"(\S)([A-Z][a-z]+)"), re.compile(r"([a-z0-9])
 arraylike_pattern = re.compile(r"([_a-zA-Z][a-zA-Z0-9]*)\[(\d+)\]")
 
 
+@functools.lru_cache()
 def _clean_key(key: str) -> str:
     def _clean(char: str) -> str:
         return f"_{ord(char):x}"
--- a/dissect/target/plugins/os/windows/wer.py
+++ b/dissect/target/plugins/os/windows/wer.py
@@ -76,6 +76,9 @@ def _collect_wer_data(wer_file: Path) -> tuple[list[tuple[str, str]], dict[str, 
             # to be used.
             key = _clean_key(key)
 
+            # Fields that start with an underscore are reserved, add a prefix.
+            key = f"n{key}" if key.startswith("_") else key
+
             record_values[key] = value
             record_fields.append((record_type, key)) if key != "ts" else record_fields.insert(0, (record_type, key))
             # reset key necessary for dynamic entries and ts
--- a/dissect/target/plugins/os/windows/wer.py
+++ b/dissect/target/plugins/os/windows/wer.py
@@ -12,6 +12,20 @@ from dissect.target.plugin import Plugin, export
 from dissect.target.target import Target
 
 camel_case_patterns = [re.compile(r"(\S)([A-Z][a-z]+)"), re.compile(r"([a-z0-9])([A-Z])"), re.compile(r"(\w)[.\s](\w)")]
+arraylike_pattern = re.compile(r"([_a-zA-Z][a-zA-Z0-9]*)\[(\d+)\]")
+
+
+def _clean_key(key: str) -> str:
+    def _clean(c):
+        return "_" + key[0].encode("unicode_escape").hex()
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/wer.py
```diff
@@ -34,13 +48,23 @@ def _collect_wer_data(wer_file: Path) -> tuple[list[tuple[str, str]], dict[str,
...
                 value = wintimestamp(int(value))
                 record_type = "datetime"
                 key = "ts"
 
             key = _key_to_snake_case(key if key else name)
 
+            # Windows wer files might contain any valid unicode character in its field names
+            # but we use the field name as python identifier. So, some character are not allowed
+            # to be used.
+            key = _clean_key(key)
```
yunzheng, line 66: I had a windows10 vm that had keys starting with an underscore. However these are not allowed as fieldnames in flow.record, example exception:

`flow.record.exceptions.RecordDescriptorError: Field '_5c7566656666_version' is an invalid or reserved field name.`

I fixed it with:

```suggestion
            key = _clean_key(key)

            # Fields that start with an underscore are reserved, add a prefix.
            key = f"n{key}" if key.startswith("_") else key
```

yunzheng: Ignore this commit suggestion, I now see that `clean_key` actually converts the the key to hex representation for non ascii bytes and the hex bytes was the representation for the BOM header which I fixed in #272 . 

Adding docstrings and unit tests would be nice.

dissect/target/plugins/os/windows/wer.py
```diff
@@ -11,6 +11,20 @@
 from dissect.target.target import Target
 
 camel_case_patterns = [re.compile(r"(\S)([A-Z][a-z]+)"), re.compile(r"([a-z0-9])([A-Z])"), re.compile(r"(\w)[.\s](\w)")]
+arraylike_pattern = re.compile(r"([_a-zA-Z][a-zA-Z0-9]*)\[(\d+)\]")
+
+
+def _clean_key(key: str) -> str:
+    def _clean(c):
+        return "_" + key[0].encode("unicode_escape").hex()
```
yunzheng, line 19: field names that start with an underscore are not allowed in `flow.record`. Consider using the prefix `n_` or something else.

dissect/target/plugins/os/windows/wer.py
```diff
@@ -11,6 +11,20 @@
...
+    first = key[0] if (key[0].isidentifier() and key[0].isascii()) else _clean(key[0])
+
+    def _is_valid(c):
+        if c.isascii():
+            return c.isidentifier() or c.isdecimal()
+        return False
+
+    return first + ''.join([c if _is_valid(c) else _clean(c) for c in key[1:]])
```
yunzheng, line 27: checking every character will most likely add some performance overhead, if there is no other way consider decorating the function with `@functools.cache`

dissect/target/plugins/os/windows/wer.py
```diff
@@ -12,6 +12,21 @@
 from dissect.target.target import Target
 
 camel_case_patterns = [re.compile(r"(\S)([A-Z][a-z]+)"), re.compile(r"([a-z0-9])([A-Z])"), re.compile(r"(\w)[.\s](\w)")]
+arraylike_pattern = re.compile(r"([_a-zA-Z][a-zA-Z0-9]*)\[(\d+)\]")
+
+
+def _clean_key(key: str) -> str:
```
pyrco, line 18:
```suggestion
@functools.lru_cache()
def _clean_key(key: str) -> str:
```
As it seems most of the keys are shared between .WER files and this will speed things up a bit, as @yunzheng mentioned.

pyrco (changes requested):
Looks good, but I think we should add the caching on `_clean_key()`

yunzheng (changes requested):
Can you add a few unit tests or examples for `_clean_key()` and for arraylike_pattern?

Some example keys from actual WER files that needs to be cleaned would already really help and make sure that parsing keeps working as intended.  The same applies for arraylike_pattern keys.

pyrco:
The core of this issue was already partially addressed by #544.
#938 will make these changes more robust. The end result will be a similar, though different in details, solution as this PR. Therefore this PR can/will be closed once #938 is merged.

pyrco:
Hi @janstarke, as the issue is now resolved by the PRs mentioned above, I will go ahead and close this PR.

PR #271, opened 2023-06-08
Remove Loaders from -f in bash autocompletion script

Miauwkeru:
(DIS-2023)

### DIFF
--- a/dissect/target/data/autocompletion/target_bash_completion.sh
+++ b/dissect/target/data/autocompletion/target_bash_completion.sh
@@ -97,7 +97,7 @@ __target_function ()
 if command -v target-query &> /dev/null
 then
     echo Loading Dissect plugin list
-    DISSECT_PLUGINS=$(target-query --quiet --list | grep -e '\s\+[a-z].\+ -' | awk '{print $1}')
+    DISSECT_PLUGINS=$(target-query --quiet --list | grep -e '^\s\+[a-z][a-z\._]\+ -' | awk '{print $1}')
 
     echo Loading Dissect help prompts
     declare -A DISSECT_HELP


dissect/target/data/autocompletion/target_bash_completion.sh
```diff
@@ -97,7 +97,7 @@ __target_function ()
 if command -v target-query &> /dev/null
 then
     echo Loading Dissect plugin list
-    DISSECT_PLUGINS=$(target-query --quiet --list | grep ' -' | awk '{print $1}')
+    DISSECT_PLUGINS=$(target-query --quiet --list | grep -e '\s\+[a-z].\+ -' | awk '{print $1}')
```
pyrco, line 100: This will trigger on the line:
```
Reason: ImportError: Please install 'yara-python' to use 'target-query -f yara'.
```

```suggestion
    DISSECT_PLUGINS=$(target-query --quiet --list | grep -e '\s\+[a-z][a-z\.]\+ -' | awk '{print $1}')
```

is probably a bit more robust.

pyrco (changes requested):

dissect/target/data/autocompletion/target_bash_completion.sh
pyrco: `--quiet` removes the errors, but this line was not an error line, it is part of the normal list output to stdout:

```
...

          wer - Return information from Windows Error Reporting (WER) files. (output: records)

Failed to load:
    Module: dissect.target.plugins.os.windows.catroot
    Reason: ModuleNotFoundError: No module named 'asn1crypto'

    Module: dissect.target.plugins.filesystem.yara
    Reason: ImportError: Please install 'yara-python' to use 'target-query -f yara'.

Available loaders:
    AsdfLoader - No documentation.

...
```

PR #272, opened 2023-06-09
Fix parsing WER files that do not have a BOM header

yunzheng:
WER files are not guaranteed to have a byte order mark (BOM).

If no BOM header is present `utf-16-le` is used as encoding, otherwise `utf-16`.


dissect/target/plugins/os/windows/wer.py
```diff
@@ -18,7 +19,16 @@ def _collect_wer_data(wer_file: Path) -> tuple[list[tuple[str, str]], dict[str,
...
     record_fields = []
     key = None
-    for line in wer_file.read_text("utf-16").splitlines():
+
+    # Default encoding, when no BOM is present
+    encoding = "utf-16-le"
+
+    # Check for the presence of a BOM header and use the correct encoding
+    with wer_file.open("rb") as fh:
+        if fh.read(len(codecs.BOM)) == codecs.BOM_UTF16_LE:
```
Schamper, line 28:
```suggestion
        if fh.read(len(codecs.BOM)) in (codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE):
```
Or is is guaranteed to be LE?

Schamper (approved):

PR #278, opened 2023-06-13
Add IIS logs to LogLoader

cecinestpasunepipe:
(DIS-2018)

### DIFF
--- a/dissect/target/loaders/log.py
+++ b/dissect/target/loaders/log.py
@@ -1,5 +1,4 @@
 import urllib
-from datetime import datetime
 from pathlib import Path
 from typing import Union
 
--- a/dissect/target/loaders/log.py
+++ b/dissect/target/loaders/log.py
@@ -36,4 +36,3 @@ class LogLoader(Loader):
             vfs.map_file(mapping, str(entry))
         target.filesystems.add(vfs)
         target.fs = vfs
-        target.datetime = datetime.today()
--- a/dissect/target/loaders/log.py
+++ b/dissect/target/loaders/log.py
@@ -15,12 +15,11 @@ class LogLoader(Loader):
         "iis": "sysvol/files/logs/",
     }
 
-    def __init__(self, path: Union[Path, str], **kwargs):
+    def __init__(self, path: Union[Path, str], parsed_path=None):
         super().__init__(path)
-        uri = kwargs.get("parsed_path")
         self.options = {}
-        if uri:
-            self.options = dict(urllib.parse.parse_qsl(uri.query, keep_blank_values=True))
+        if parsed_path:
+            self.options = dict(urllib.parse.parse_qsl(parsed_path.query, keep_blank_values=True))
 
     @staticmethod
     def detect(path: Path) -> bool:
--- a/dissect/target/loaders/log.py
+++ b/dissect/target/loaders/log.py
@@ -1,4 +1,7 @@
+import urllib
+from datetime import datetime
 from pathlib import Path
+from typing import Union
 
 from dissect.target import Target
 from dissect.target.filesystem import VirtualFilesystem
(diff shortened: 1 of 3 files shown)


dissect/target/loaders/log.py
```diff
@@ -7,10 +10,18 @@
...
-        ".evtx": "sysvol/windows/system32/winevt/logs",
-        ".evt": "sysvol/windows/system32/config",
+        "evtx": "sysvol/windows/system32/winevt/logs",
+        "evt": "sysvol/windows/system32/config",
+        "iis": "sysvol/files/logs/",
     }
 
+    def __init__(self, path: Union[Path, str], **kwargs):
+        super().__init__(path)
+        uri = kwargs.get("parsed_path")
```
Poeloe, line 20: If I'm correct, you can also add a optional parameter uri to the init method. If uri is in kwargs, it will be automatically filled. That way, you don't have to do the additional get ;)

dissect/target/plugins/apps/webservers/iis.py
```diff
@@ -85,6 +87,11 @@ def iter_log_format_path_pairs(self) -> list[tuple[str, str]]:
             for log_file in self.iter_log_paths(log_dir_path):
                 yield (log_format, log_file)
 
+    @plugin.internal
+    def parse_autodetect_format_log(self, path: Path) -> Iterator[BasicRecordDescriptor]:
+        yield from self.parse_iis_format_log(path)
+        yield from self.parse_w3c_format_log(path)
```
Poeloe, line 93: To my understanding, we now also will get some noise because W3C logs will sort of be parsed when analysed with the IIS format code. So I guess we do need logic here to determine the log file, but I might be wrong.

Main difference is W3C logs are separated with spaces (and all values are quoted if they contain spaces), IIS logs contain commas instead. So maybe this can be used for fast logic?

Poeloe (changes requested):

dissect/target/plugins/apps/webservers/iis.py
Poeloe: Looking more into the code, it will probably just give a whole lot of debug log lines instead. It will still go over each line in the log, so generating some overhead there. Therefore, I would still suggest to add some logic here.

dissect/target/loaders/log.py
```diff
@@ -19,10 +30,11 @@ def map(self, target: Target) -> None:
...
         for entry in self.path.parent.glob(self.path.name):
-            ext = entry.suffix.lower()
+            ext = self.options.get("hint", entry.suffix.lower()).strip(".")
             if (mapping := self.LOGS_DIRS.get(ext, None)) is None:
                 continue
             mapping = str(Path(mapping).joinpath(entry.name))
             vfs.map_file(mapping, str(entry))
         target.filesystems.add(vfs)
         target.fs = vfs
+        target.datetime = datetime.today()
```
Schamper, line 40: What’s this for?

cecinestpasunepipe: The plugin expects time zone information but because this is not a regular target, no such thing is available, this is just to make sure the plugin does not crash.

cecinestpasunepipe: But you are right, this is not a very clean solution, I will improve it.

tests/test_loaders_log.py
```diff
@@ -9,20 +10,21 @@
...
 
 @pytest.mark.parametrize(
-    "input_file, expected_mapping",
+    "path, uri, input_file, expected_mapping",
     [
-        ("/dir/test.evtx", "/sysvol/windows/system32/winevt/logs/test.evtx"),
-        ("/dir/test.evt", "/sysvol/windows/system32/config/test.evt"),
+        ("/dir/*.evt*", None, "/dir/test.evtx", "/sysvol/windows/system32/winevt/logs/test.evtx"),
+        ("/dir/*.evt*", None, "/dir/test.evt", "/sysvol/windows/system32/config/test.evt"),
+        ("/source/iis.log", "log:///dir/with/files/*.log?hint=iis", "/source/iis.log", "/sysvol/files/logs/iis.log"),
```
Poeloe, line 17: I do like this "hint" parameter! It makes it possible to specify the type of *.log file (or whatever standard log output format) that should be interpreted for each file separately. 

We need to document this clearly somewhere, such that it is option is generally known

Poeloe (approved):

PR #282, opened 2023-06-16
Fix sequential bug caused by copy-by-reference mistake in plugin system

cecinestpasunepipe:
Because the plugin finder modifies the dict _os plugins become inaccessible.


dissect/target/plugin.py
```diff
@@ -837,7 +837,9 @@ def all_plugins():
         yield from os_plugins()
         yield from child_plugins()  # Doesn't export anything but added for completeness.
 
-    for available in all_plugins():
+    for available_original in all_plugins():
+        # Otherwise -f os.windows._os.users fails for instance
```
pyrco, line 841:
```suggestion
        # Prevent modifying the global PLUGINS dict, otherwise -f os.windows._os.users fails for instance.
```

Gives maybe a bit more direction, on what this actually would do.

pyrco (changes requested):

pyrco (approved):

PR #290, opened 2023-06-23
Update WSL plugin to use registry to find all WSL instances

sulonl:
Updated wsl.py to not only look for WSL distribution disks (.vhdx) in default install location, but to use the Windows Registry of the Windows target to find other disk files for WSL on other locations as well which may not be on the default location.

### DIFF
--- a/dissect/target/plugins/child/wsl.py
+++ b/dissect/target/plugins/child/wsl.py
@@ -22,9 +22,7 @@ def find_wsl_installs(target: Target) -> Iterator[Path]:
         for distribution_key in lxss_key.subkeys():
             if not distribution_key.name.startswith("{"):
                 continue
-            base_path = distribution_key.value("BasePath").value.replace(
-                "\\\\?\\", ""
-            )  # "\\?\" not resolved by Dissect, temporary replace.
+            base_path = target.resolve(distribution_key.value("BasePath").value)
             # WSL needs diskname to be ext4.vhdx, but they can be renamed when WSL is not active
             yield from target.fs.path(base_path).glob("*.vhdx")
 
--- a/dissect/target/plugins/child/wsl.py
+++ b/dissect/target/plugins/child/wsl.py
@@ -8,11 +8,11 @@ from dissect.target.target import Target
 
 def find_wsl_installs(target: Target) -> Iterator[Path]:
     """Find all WSL disk files.
-    
+
     Disk files for working (custom) Linux distributions can be located anywhere on the system.
     Locations to disk files for each user's WSL instance is stored in the Windows registry at
     ``HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Lxss``.
-    
+
     References:
         - https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro
         - https://learn.microsoft.com/en-us/windows/wsl/enterprise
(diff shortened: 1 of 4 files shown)


dissect/target/plugins/child/wsl.py
```diff
@@ -7,33 +7,37 @@
...
+                yield disk_file
 
 
 class WSLChildTargetPlugin(ChildTargetPlugin):
     """Child target plugin that yields WSL VHDX file locations.
 
-    Windows WSL VHDX conatiners are stored at ``%AppData%\\Local\\Packages\\$DistFolder\\LocalState\\ext4.vhdx``,
-    where ``$DistFolder`` will be substituted with a unix distribution folder.
+    Windows WSL VHDX disk file locations are stored in the Windows Registry in `HKEY_USERS/$sid/Software/Microsoft/Windows/CurrentVersion/Lxss`,
+    where $sid is substituted with any SID of users on the Windows machine.
```
Schamper, line 35:
```suggestion
    Windows WSL VHDX disk file locations are stored in the Windows registry in ``HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Lxss``.
```

dissect/target/plugins/child/wsl.py
```diff
@@ -7,33 +7,37 @@
...
-            for install_path in user_details.home_path.joinpath("AppData/Local/Packages").glob(dist_folder):
-                if (vhdx := install_path.joinpath("LocalState/ext4.vhdx")).exists():
-                    yield vhdx
+    """Disk files for working (custom) Linux distributions can be located anywhere on the system.
+    Locations to disk files for each user's WSL instance is stored in the Windows Registry in the following subkey:
+    - HKCU\Software\Microsoft\Windows\CurrentVersion\Lxss
+    References:
+        - https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro
+        - https://learn.microsoft.com/en-us/windows/wsl/enterprise
+    """
```
Schamper, line 16:
```suggestion
    """Find all WSL disk files.
    
    Disk files for working (custom) Linux distributions can be located anywhere on the system.
    Locations to disk files for each user's WSL instance is stored in the Windows registry at
    ``HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Lxss``.
    
    References:
        - https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro
        - https://learn.microsoft.com/en-us/windows/wsl/enterprise
    """
```
General structure of a `"""` style docstring is:

```
"""Short description of function/method.

Longer explanation goes here.

References/args/other
"""
```

Schamper (changes requested):

dissect/target/plugins/child/wsl.py
```diff
@@ -7,33 +7,37 @@
...
+
+    for lxss_key in target.registry.keys("HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Lxss"):
+        for distribution_key in lxss_key.subkeys():
+            if not distribution_key.name.startswith("{"):
+                continue
+            base_path = distribution_key.value("BasePath").value.replace(
```
Schamper, line 25: My only comment is that this should be properly fixed, otherwise this is good to go!

Schamper (changes requested):

Schamper (approved):

PR #292, opened 2023-06-26
Fix UTMP IPv6

Zawadidone:


Zawadidone:
What is the best place for this code? So that the code can be re-used by the BTMP parser.

Schamper:
> What is the best place for this code? So that the code can be re-used by the BTMP parser.

Perhaps here, and yield some named tuples with this pre-parsed field instead of the direct cstruct entry? https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/unix/log/utmp.py

dissect/target/plugins/os/unix/log/utmp.py
```diff
@@ -68,6 +86,30 @@ def __iter__(self):
...
+
+                utmp_entry = UTMP_ENTRY(
+                    ts=from_unix(entry.ut_tv.tv_sec),
+                    ut_type=r_type,
+                    ut_pid=entry.ut_pid,
+                    ut_user=entry.ut_user.decode().strip("\x00"),
+                    ut_line=entry.ut_line.decode().strip("\x00"),
+                    ut_id=entry.ut_id.decode().strip("\x00"),
+                    ut_host=entry.ut_host.decode().strip("\x00"),
+                    ut_addr=ipaddress.ip_address(ut_addr),
```
cecinestpasunepipe, line 110:
```suggestion
                    ut_user=entry.ut_user.decode(errors="surrogateescape").strip("\x00"),
                    ut_line=entry.ut_line.decode(errors="surrogateescape").strip("\x00"),
                    ut_id=entry.ut_id.decode(errors="surrogateescape").strip("\x00"),
                    ut_host=entry.ut_host.decode(errors="surrogateescape").strip("\x00"),
                    ut_addr=ipaddress.ip_address(ut_addr),
```

dissect/target/plugins/os/unix/log/utmp.py
```diff
@@ -68,6 +86,30 @@ def __iter__(self):
...
+                r_type = ""
+                if entry.ut_type in utmp.Type.reverse:
+                    r_type = utmp.Type.reverse[entry.ut_type]
+
+                # UTMP misuses the field ut_addr_v6 for IPv4 and IPv6 addresses, because of this
+                # if the last 12 bytes are zero the value is read as an IPv4-address.
+                if entry.ut_addr_v6[1:] == [0, 0, 0]:
+                    ut_addr = struct.pack("<i", entry.ut_addr_v6[0])
+                else:
+                    ut_addr = struct.pack("<4i", *entry.ut_addr_v6)
```
cecinestpasunepipe, line 100: should this logic not be behind a check?
"System V has no ut_host or ut_addr_v6 fields."
https://man7.org/linux/man-pages/man5/utmp.5.html

dissect/target/plugins/os/unix/log/utmp.py
```diff
@@ -68,6 +86,30 @@ def __iter__(self):
...
         while True:
             try:
-                yield utmp.entry(byte_stream)
+                entry = utmp.entry(byte_stream)
+
+                r_type = ""
+                if entry.ut_type in utmp.Type.reverse:
+                    r_type = utmp.Type.reverse[entry.ut_type]
+
+                # UTMP misuses the field ut_addr_v6 for IPv4 and IPv6 addresses, because of this
```
cecinestpasunepipe, line 95: Can you write a little test maybe for this fix? So we can easily spot a regression?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/utmp.py
Zawadidone: Yes it should

cecinestpasunepipe: I tried to run this plugin on BSD and while the file was parsed without errors the contents did not seem to make sense. Also there seem to be a lot of variations among Linux distros. The only way to determine the file format seems to be to check the version of libc (since that is where utmp.h resides) but that might be quite difficult (i.e. muslc etc..). Another option is to check if the contents contains these 0-bytes because they are a good indicator that the plugin is incompatible, maybe just issue a warning in that case.

Pragmatic approach:
1. Allow the plugin to run on all systems, check for difference SystemV
2. If output records contain 0-bytes (see error handling), emit warning

We can always refine this plug-in later of course, along with all the utmp variations we encounter.

Zawadidone: Can you create an issue for these utmp/libc variation’s?

dissect/target/plugins/os/unix/log/utmp.py
Zawadidone: The fix introduces the same bug in case the IPv6 address is 4 bytes long, but works for IPv6 address that are longer than 4 bytes.

The IPv6 address is `1337:1::'` and in the WTP log is b'\x137\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', the last 12 bytes are zeroes. Because of this IPv6 addresses of which the binary representation is 4 bytes long is parsed as IPv4 address which leads to the address `19.55.0.1`:

```
>>> import ipaddress
>>> ipv6 = "1337:1::"
>>> ipaddress.ip_address(ipv6)
IPv6Address('1337:1::')
>>> ipaddress.ip_address(ipv6).packed
b'\x137\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
>>> ipaddress.ip_address(ipv6).packed[:4]
b'\x137\x00\x01'
>>> ipaddress.ip_address(b"\x137\x00\x01")
IPv4Address('19.55.0.1')
```

I would like to make the suggestion to make the UTMP parser a bit smarter to fix this issue.

* `ut_addr_v6` field uses more then 4 bytes -> IPv6 address
* `ut_addr_v6` field uses 4 bytes
	* If `ut_host is a valid IPv6 address the full 16 bytes of `ut_addr_v6` are parsed as IPv6 address
	* Otherwise the first 4 bytes are parsed as IPv4 address

Zawadidone:
@cecinestpasunepipe should the files `wtmp.py` and `btmp.py` be combined in `utmp.py` with the namespace `utmp` with the functiosn wtmp and btmp? Because the code is identical and only the path and record name is different.

Schamper:
> @cecinestpasunepipe should the files `wtmp.py` and `btmp.py` be combined in `utmp.py` with the namespace `utmp` with the functiosn wtmp and btmp? Because the code is identical and only the path and record name is different.

You could also just put a `UtmpPlugin` in `utmp.py` with a function `btmp` and `wtmp`, skipping the namespace.

dissect/target/plugins/os/unix/log/utmp.py
Zawadidone: Can this comment be added to a Github issue outside of this PR?

dissect/target/plugins/os/unix/log/utmp.py
Zawadidone: Ah my mistake thanks for seeing that.

I have created a new test file that includes the field `ut->ut_addr_v6`.

As shown in the screenshot utmpdump contains the same issue.

I have used SSH using the IPv6 address `1337::1` which uses 16 bytes, utmpdump correctly shows that the hostname is `1337::1` and the IP-address is `1337::1`.

I have used SSH using the IPv6 address `1337:1::` which uses 4 bytes lead by 12 bytes of trailing zeroes, utmpdump correctly shows that the hostname is `1337:1::` which is correct and the IP-address is `19.55.0.1` which is incorrect. 

image

The last 12 bytes of leading zeroes causes utmpdump to read the buffer as `AF_INET` instead of `AF_INET6`.

```python
>>> import ipaddress
>>> ipv6 = ipaddress.ip_address("1337:1::")
>>> ipv6
IPv6Address('1337:1::')
>>> ipv6.packed
b'\x137\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
>>> ipv6.packed[:4]
b'\x137\x00\x01'
>>> ipaddress.ip_address(ipv6.packed[:4])
IPv4Address('19.55.0.1')
```

To prevent this issue as shown in the test the `ut_host` field is used to determine the type of the IP-address

Zawadidone:
Okay I have combined the functions

dissect/target/plugins/os/unix/log/utmp.py
Zawadidone: I don't known how the last case can be reproduced on a Linux machine and correctly parsed by the parser.

| `ut_host` | `ut_addr_v6` | result |
|---|---|---|
| IPv4 address  | IPv4 address |  `ut_addr_v6[0]` is parsed as IPv4 address |
| Random string  | IPv4 address |  `ut_addr_v6[0]` is parsed as IPv4 address |
| Pv6 address  | IPv6 address > 4 bytes)|  `ut_addr_v6` is parsed as IPv6 address |
| Random string  | IPv6 address > 4 bytes)|  `ut_addr_v6` is parsed as IPv6 address |
|  IPv6 address   | IPv6 address <= 4 bytes |  `ut_addr_v6` is parsed as IPv6 address |
|  Random string   | IPv6 address <= 4 bytes |  `ut_addr_v6[0]` is parsed as IPv6 address which is incorrect https://github.com/fox-it/dissect.target/pull/292/files#diff-4676dfeb3cf44a047e82d239cd74ead57f206d968ce3a428c24e4c7b92869272R152-R154 |

pyrco: Ah I see, the utmpdump tool itself also puts out the wrong information. So basically the way the utmp structure is set filled is fundamentally broken unfortunately.

I don't think the final case in your table is easily fixable as there is just too little information. On the plus side I guess that encountering a IP6 address with only the first 32 bits set will probably be rare.

pyrco (approved):

cecinestpasunepipe:
> Okay I have combined the functions

Yes, I will review this PR right now.

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/log/utmp.py
```diff
@@ -68,6 +122,120 @@ def __iter__(self):
...
             except EOFError:
                 break
+
+
+class UtmpPlugin(Plugin):
+    WTMP_GLOB = "/var/log/wtmp*"
+    BTMP_GLOB = "/var/log/btmp*"
+
+    def check_compatible(self) -> None:
+        if not any(
```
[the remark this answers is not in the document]
Zawadidone: In case BSD support is chosen and modification of the c-struct is needed, can this be done in another PR? I would like to use this plugin for Linux host investigations without observing random IPv4 addresses that should be IPv6 addresses.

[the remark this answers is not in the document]
pyrco: I think doing BSD support in a separate PR is ok. I think if you just check `self.target.os == linux` in `check_compatible()` for now, we won't need to move it to the linux subdir and users on BSD platforms won't get surprises.

cecinestpasunepipe (approved):
LGTM

pyrco (approved):

Zawadidone:
@pycro @cecinestpasunepipe thank you for reviewing the PR and coming with suggestions to fix the problem.

PR #298, opened 2023-06-29
Add support for gpo tasks that are wrapped in other tasks

cecinestpasunepipe:
- make the xml search query more generic so it also matches wrapped tasks
- make the plugin more robust (do not fail if appname is not present, happens sometimes)

### DIFF
--- a/dissect/target/plugins/os/windows/adpolicy.py
+++ b/dissect/target/plugins/os/windows/adpolicy.py
@@ -69,7 +69,7 @@ class ADPolicyPlugin(Plugin):
                 task_file_stat = task_file.stat()
                 xml = task_file.read_text()
                 tree = ElementTree.fromstring(xml)
-                for task in tree.findall(".//{}Task"):
+                for task in tree.findall(".//{*}Task"):
                     properties = task.find("Properties") or task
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -69,16 +69,16 @@ def _xmltasks(self, policy_dir):
                 task_file_stat = task_file.stat()
                 xml = task_file.read_text()
                 tree = ElementTree.fromstring(xml)
-                for task in tree.findall("Task"):
-                    properties = task.find("Properties")
+                for task in tree.findall(".//{}Task"):
```
pyrco, line 72: I don't know much about XPath, but I couldn't find what the `{}` does in this context.

cecinestpasunepipe: It's to access nodes with empty namespaces. {*} would be better yet maybe.

cecinestpasunepipe: For some reason the xml contained empty namespaced nodes. {*} will allow the query to access all namespaced nodes.

pyrco (approved):

PR #307, opened 2023-07-10
Fix Dirloader to work with acquired dirs (-ot)

cecinestpasunepipe:
(DIS-1856)

### DIFF
--- a/dissect/target/loaders/dir.py
+++ b/dissect/target/loaders/dir.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -19,7 +18,7 @@ class DirLoader(Loader):
     PREFIXES = ["", "fs"]
 
     @classmethod
-    def _find_entry_path(cls, path: Path):
+    def _find_entry_path(cls, path: Path) -> str | None:
         for prefix in cls.PREFIXES:
             if find_dirs(path / prefix)[0] is not None:
                 return prefix
--- a/dissect/target/loaders/dir.py
+++ b/dissect/target/loaders/dir.py
@@ -11,24 +11,24 @@ from dissect.target.plugin import OperatingSystem
 if TYPE_CHECKING:
     from dissect.target import Target
 
+PREFIXES = ["", "fs"]
+
+
+def find_entry_path(path: Path) -> str | None:
+    for prefix in PREFIXES:
+        if find_dirs(path / prefix)[0] is not None:
+            return prefix
+
 
 class DirLoader(Loader):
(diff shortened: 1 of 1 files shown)


dissect/target/loaders/dir.py
```diff
@@ -17,9 +17,13 @@ class DirLoader(Loader):
 
     @staticmethod
     def detect(path: Path) -> bool:
+        if (path / "fs").exists():
```
Schamper, line 20: I think this check is a bit too assuming. Could you add some more conditions? E.g. only do this if `find_dirs` returns `None` first?

dissect/target/loaders/dir.py
```diff
@@ -15,15 +16,24 @@
...
+
+    @classmethod
+    @cache
+    def _find_entry_path(cls, path: Path):
+        found = None
+        for prefix in cls.PREFIXES:
+            if find_dirs(path / prefix)[0] is not None:
+                found = prefix
+                break
+        return found
```
Schamper, line 29:
```suggestion
        for prefix in cls.PREFIXES:
            if find_dirs(path / prefix)[0] is not None:
                return prefix
```

cecinestpasunepipe: I always try to avoid multiple returns in one function, but this is more compact. Trade-off I guess.

cecinestpasunepipe:
Unfortunately the prefix-function can't be cached due to side-effects.

Horofic (dismissed):

dissect/target/loaders/dir.py
```diff
@@ -15,11 +15,20 @@
 class DirLoader(Loader):
     """Load a directory as a filesystem."""
 
+    PREFIXES = ["", "fs"]
+
+    @classmethod
+    def _find_entry_path(cls, path: Path) -> str | None:
```
Schamper, line 21: Maybe you can make this a function in the global scope, similar to the other functions here. There's not much reason for it to be a classmethod, is there?

dissect/target/loaders/dir.py
```diff
@@ -11,15 +11,24 @@
 if TYPE_CHECKING:
     from dissect.target import Target
 
+PREFIXES = ["", "fs"]
+
+
+def find_entry_path(path: Path) -> str | None:
```
Schamper, line 17:
```suggestion
def find_entry_path(path: Path) -> Optional[str]:
```
I don't think using `|` in type hints is legal in Python 3.9?

Could you also move this below the `DirLoader` class itself? That way it's more consistent placing with the other functions.

Schamper (changes requested):

Schamper (changes requested):

Schamper (approved):

PR #312, opened 2023-07-13
Only support simple assignments in config parser

Schamper:


dissect/target/helpers/config.py
```diff
@@ -1,23 +1,52 @@
...
 
 
-def _find_config_file(path):
+def _parse_ast(code: str) -> dict[str, Union[str, int]]:
+    # Only allow basic value assignments for backwards compatibility
+    obj = {}
+
+    module = ast.parse(code)
+    if not isinstance(module, ast.Module):
+        return obj
```
pyrco, line 27: Would it be an idea to have some debug logging here and at the `continue`s so you have some feedback when things don't parse?

Schamper: Added some.

pyrco (approved):

PR #314, opened 2023-07-14
Support more log data types for etl

Miauwkeru:
(DIS-1999)

### DIFF
--- a/dissect/target/plugins/os/windows/log/etl.py
+++ b/dissect/target/plugins/os/windows/log/etl.py
@@ -1,12 +1,12 @@
 from functools import lru_cache
 
-from dissect.cstruct import Instance
 from dissect.etl.etl import ETL, Event
 
 from dissect.target import Target
 from dissect.target.exceptions import FilesystemError
 from dissect.target.helpers.record import DynamicDescriptor, TargetRecordDescriptor
 from dissect.target.plugin import Plugin, export
+from dissect.target.plugins.os.windows.datetime import parse_tzi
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/log/etl.py
```diff
@@ -31,12 +31,15 @@ def _build_record(self, etl_event: Event, etl_path: str, target: Target):
...
+            record_type = "bytes"
             if isinstance(value, list):
-                record_fields.append(("string[]", key))
+                record_type = "string[]"
             elif isinstance(value, int):
-                record_fields.append(("varint", key))
-            else:
-                record_fields.append(("string", key))
+                record_type = "varint"
+            elif isinstance(value, str) and value.isprintable() or isinstance(value, Instance):
```
pyrco, line 39: If I understand the etl code correct, `etl_event.event_values()` are either key,value pairs from normal Python `dict()`s (from the headers `additional_header_fields()`) or cstruct `Instance._values` key,value pairs (from the parsed structure).

In that light I don't see how a value that is a `str` instance can ever not be a `"string"`. So I don't see why the `is_printable()` is needed. Also classifying any value that itself is a cstruct `Instance` as a `"string"` seems strange. Could you explain?

Finally is the precendece correct or should there be some brackets like:
```suggestion
            elif (isinstance(value, str) and value.isprintable()) or isinstance(value, Instance):
```

pyrco (approved):

PR #315, opened 2023-07-15
Fix 'test_fs_attrs_no_os_listxattr' on Darwin

nrhtr:
Without this I get the following test failure on aarch64-darwin:

> FAILED tests/test_helpers_fsutil.py::test_fs_attrs_no_os_listxattr
> - AttributeError: module 'os' has no attribute 'listxattr'


pyrco (changes requested):
Thanks for the useful fix! There is only a linting issue, but if that is fixed it looks good to go.

I see the Bot already asked about the Contributor License Agreement, so if that and the linting error are fixed I'll merge it in.

nrhtr:
@DissectBot agree

PR #321, opened 2023-07-19
Add support for more history files

JSCU-CNI:
This PR adds support for fish, mongodb, mysql, postgresql and sqlite history files.

### DIFF
--- a/dissect/target/plugins/os/unix/history.py
+++ b/dissect/target/plugins/os/unix/history.py
@@ -1,42 +1,61 @@
 import re
-from typing import Iterator
+from typing import Iterator, List, Tuple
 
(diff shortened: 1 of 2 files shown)


pyrco:
Thanks for the addition and excuses for the long turn-around. It seems there are some linting errors in the code.
Otherwise the patch itself looks good, so if you could fix these errors, I'll merge the PR.

pyrco (changes requested):
Thanks for the addition and excuses for the long turn-around. It seems there are some linting errors in the code.
Otherwise the patch itself looks good, so if you could fix these errors, I'll merge the PR.

pyrco (approved):

PR #322, opened 2023-07-19
Add unix sshd plugin

JSCU-CNI:
This PR adds a parser for sshd configuration files.


dissect/target/plugins/os/unix/sshd.py
```diff
@@ -0,0 +1,160 @@
+import logging
```
cecinestpasunepipe, line 1: This is a very interesting contribution, thank you.

We very much appreciate this contribution and all the effort but this potential new plugin has sparked an important internal debate in our development team about the future of configuration plugins on UNIX and Linux. We would like to propose to keep this plugin and potential future plugins regarding  UNIX configuration **on-hold** for now, just to prevent you from wasting any of your valuable time.

Please allow us some time to reflect on our current strategy before we proceed.

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/unix/sshd.py
cecinestpasunepipe: To clarify, we are not rejecting this plugin but we are researching if there can be a more generic approach to querying configurations for Unix/Linux systems and how to integrate your contributions in this regard properly. But before we can be more clear about it, we need to have a better idea of how to approach this architectural challenge.

cecinestpasunepipe:
Update: 

@JSCU-CNI We have a PoC for a generic UNIX config parser here: https://github.com/fox-it/dissect.target/commit/f6b76dfb14cf47ea52be0addf133fcc9765a5b7b - we also have an implementation plan for other types of config files and we will start incorporating your PR here on top of the generic parser soon.

Our schedule:

- Port your PR to the new generic UNIX config parser (first priority!)
- Implement other well known formats inside /etc like CSV, XML, YAML, JSON and MISC

You can also give the new parser a try with target_shell-registry.
Please let us know if you have any questions / comments.
All feedback is welcome!

JSCU-CNI:
Thank you for all the work in #422 and #354. I think this PR can be closed now :tada:

PR #325, opened 2023-07-21
Clean the systemd service configuration parser

Miauwkeru:
(DIS-1907)

### DIFF
--- a/tests/test_plugins_os_unix_services.py
+++ b/tests/test_plugins_os_unix_services.py
@@ -50,6 +50,7 @@ def test_unix_services(target_unix_users, fs_unix):
         ("[Unit]\nnew_lines=hello \\\nworld\\\ntest", 'Unit_new_lines="hello world test"'),
         ("[Unit]\ntest", 'Unit_test="None"'),
         ("[Unit]\nnew_lines=hello \\\n#Comment\n;Comment2\nworld", 'Unit_new_lines="hello world"'),
+        ("[Unit]\nlines=hello \\\nworld\\\n\ntest", 'Unit_lines="hello world test"'),
     ],
 )
 def test_unix_systemd_parser(assignment, expected_value):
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/unix/services.py
```diff
@@ -112,37 +113,19 @@ def parse_systemd_config(fh: BinaryIO) -> str:
     This should probably be rewritten to return a proper dict as in
     its current form this is only useful when used in Splunk.
     """
-    variables = ""
+    parser = RawConfigParser(strict=False)
```
pyrco, line 116: The default delimiters are `=` and `:`, while systemd only allows `=`.

Also multi line values seem to must have a `\` at the end of the line (and possibly indenting of the next line), where the default for `RawConfigParser` is just indenting of the next line. How does it handle this?

systemd config allows empty values but with a delimiter (`=`), which is different from the (disabled by default) option `allow_no_value` which allows for keys without a delimiter all together. Are these empty values handled properly?

dissect/target/plugins/os/unix/services.py
```diff
@@ -106,43 +107,45 @@ def should_ignore_file(needle: str, haystack: list) -> bool:
...
 
-def parse_systemd_config(fh: BinaryIO) -> str:
+def parse_systemd_config(fh: TextIO) -> str:
     """Returns a string of key/value pairs from a toml/ini-like string.
 
     This should probably be rewritten to return a proper dict as in
     its current form this is only useful when used in Splunk.
     """
-    variables = ""
+    parser = ConfigParser(strict=False, delimiters=("=",), allow_no_value=True)
```
pyrco, line 116:
```suggestion
    parser = ConfigParser(strict=False, delimiters=("=",), allow_no_value=True)
```

`allow_no_value=True` is not realy what I meant, as that would allow

```
[Unit]
test
```
which I don't think is allowed in systemd configs (never tested it so I'm not sure). I was referring to the case of:

```
[Unit]
test=
```
Maybe this is already handled by the default settings.

pyrco (changes requested):

tests/test_plugins_os_unix_services.py
```diff
@@ -47,10 +48,26 @@ def test_unix_services(target_unix_users, fs_unix):
...
     ],
 )
 def test_unix_systemd_parser(assignment, expected_value):
     data = parse_systemd_config(StringIO(assignment))
     assert data == expected_value
+
+
+@pytest.mark.xfail
+def test_unix_systemd_known_fail():
+    # While this should return `Hello world test help` it fails because it cannot find something
```
pyrco, line 62: `something` is a bit vague, could you make that a bit more explicit. Maybe also say how it is parsed.

Probably also good to add this case:

```
lines=hello \\
world\\

test
```
as the empty line should be ignored like the comments but is not if I understood correctly.

pyrco (changes requested):

pyrco (approved):

PR #327, opened 2023-07-25
Add support for Linux MD RAID

Schamper:
Also includes some small fixes for LVM2 and BDE, as well as support for nested logical volumes.

Depends on https://github.com/fox-it/dissect.volume/pull/18.

(DIS-1388)


Schamper:
I was thinking if it might make sense to move the LVM/FDE logic into the `DiskCollection` logic. Thoughts?

pyrco (approved):

PR #329, opened 2023-07-26
Remove duplicate Unix log functions.

Zawadidone:
The duplicate functions behave the same. This creates duplicate records when executing all `target-query` functions in an automated manner.

### DIFF
--- a/dissect/target/plugins/os/unix/log/auth.py
+++ b/dissect/target/plugins/os/unix/log/auth.py
@@ -26,11 +26,6 @@ class AuthPlugin(Plugin):
         return any(var_log.glob("auth.log*")) or any(var_log.glob("secure*"))
 
     @export(record=[AuthLogRecord])
-    def securelog(self) -> Iterator[AuthLogRecord]:
-        """Return contents of /var/log/auth.log* and /var/log/secure*."""
-        return self.authlog()
-
-    @export(record=[AuthLogRecord])
     def authlog(self) -> Iterator[AuthLogRecord]:
         """Return contents of /var/log/auth.log* and /var/log/secure*."""
 
--- a/dissect/target/plugins/os/unix/log/messages.py
+++ b/dissect/target/plugins/os/unix/log/messages.py
@@ -30,14 +30,6 @@ class MessagesPlugin(Plugin):
         return any(var_log.glob("syslog*")) or any(var_log.glob("messages*"))
 
     @export(record=MessagesRecord)
-    def syslog(self) -> Iterator[MessagesRecord]:
-        """Return contents of /var/log/messages* and /var/log/syslog*.
-
-        See ``messages`` for more information.
-        """
-        return self.messages()
-
-    @export(record=MessagesRecord)
     def messages(self) -> Iterator[MessagesRecord]:
         """Return contents of /var/log/messages* and /var/log/syslog*.
 
--- a/tests/test_plugins_os_unix_log_messages.py
+++ b/tests/test_plugins_os_unix_log_messages.py
@@ -45,11 +45,6 @@ def test_unix_log_messages_plugin(target_unix_users, fs_unix):
         assert results[1].pid == 1
         assert results[1].source == path.from_posix("/var/log/messages")
 
-    # assure syslog() behaves the same as messages()
-    syslogs = list(target_unix_users.syslog())
-    assert len(syslogs) == len(results)
-    assert isinstance(syslogs[0], type(MessagesRecord()))
-
 
 @pytest.mark.skipif(platform.system() == "Windows", reason="ZoneInfoNotFoundError. Needs to be fixed.")
 def test_unix_log_messages_compressed_timezone_year_rollover():


Zawadidone:
@JSCU-CNI was there a particular reasons to add extra function that behaves the same?

```
target-query -l
[...]
	messages:
    	messages - Return contents of /var/log/messages* and /var/log/syslog*. (output: records)
        syslog - Return contents of /var/log/messages* and /var/log/syslog*. (output: records)
        
    auth:
            authlog - Return contents of /var/log/auth.log* and /var/log/secure*. (output: records)
            securelog - Return contents of /var/log/auth.log* and /var/log/secure*. (output: records)
[...]
```

JSCU-CNI:
> @JSCU-CNI was there a particular reason to add extra functions that behaves the same?

Yes, this was chosen deliberately, mainly for UX reasons.

Different Unix distributions have different common names for artifacts. For example `/var/log/secure` is used in RHEL/CentOS and `/var/log/auth.log` is used in Debian based systems. As far as I'm aware it's not possible (yet) to add 'symlink' plugin names, so we added both.

Zawadidone:
@JSCU-CNI I have seperated the used log paths of the functions, this way the UX shows separate functions as shown below and when all `target-query` functions are executed those functions don't create duplicate records.

```bash
target-query -l | grep -e "messages|syslog|securelog|authlog" | grep output
            authlog - Return contents of /var/log/auth.log*. (output: records)
            securelog - Return contents of /var/log/secure*. (output: records)
            messages - Return contents of /var/log/messages*. (output: records)
            syslog - Return contents of /var/log/syslog*. (output: records)
```

Zawadidone:
@JSCU-CNI should the glob `/var/log/messages*` and  `/var/log/syslog*` be added to this function?

https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/unix/_os.py#L71

JSCU-CNI:
This seems to be a problem because you are trying to blindly run all target-query functions at once on your targets. Instead of changing all plugins causing problems with this approach, I would change your approach. 

In our opinion separating the functionality of the `messages` and `authlog` functions is not preferred. Perhaps a single name should be chosen instead of having a second function that merely redirects to another function. However you will still see issues with other functions. For example with `commandhistory` and `bashhistory`, which behave the same as `messages` and `authlog` do, but in this instance for backwards compatibility reasons.

Perhaps @Schamper can weigh in here :)

Zawadidone:
> This seems to be a problem because you are trying to blindly run all target-query functions at once on your targets. Instead of changing all plugins causing problems with this approach, I would change your approach.

Can you elaborate on this or explain what is a better approach?

This how I do it to maximize the usage of available processors, almost the same as explained in the [documentation](https://docs.dissect.tools/en/latest/usage/first-steps/incident-handler.html#write-target-query-functions-output-to-a-file):

```bash
target-query --list | grep ' -' | grep 'output: records' | grep -vE 'yara|remoteaccess\.remoteaccess|browser\.history|get_all_records|example' | awk '{print $1}' > FUNCTIONS

# Execute the functions and output in split JSON files with a maximum of 10000 records
cat FUNCTIONS | xargs -I {} -P 10 sh -c 'target-query -q -t MSEDGEWIN10.tar -f {} --no-cache --children 2>>/dev/null | rdump --split 10000 -w export/plugins/{}.json --suffix-length 10 2>>/dev/null'

# Create records based on the datetime fields in the records and save the output in JSONL files
find export/plugins -type f -print0 | xargs -r0I {} -P 10 sh -c 'rdump {} --multi-timestamp -w jsonfile://export/$(basename {} .json).jsonl?descriptors=True'
```

> In our opinion separating the functionality of the `messages` and `authlog` functions is not preferred. Perhaps a single name should be chosen instead of having a second function that merely redirects to another function. However you will still see issues with other functions. For example with `commandhistory` and `bashhistory`, which behave the same as `messages` and `authlog` do, but in this instance for backwards compatibility reasons.

In case of messages, is using the name `messages` preferred as shown in this commit [`e731c5d` (#329)](https://github.com/fox-it/dissect.target/pull/329/commits/e731c5d942a461470800c882aba7766cb7fc5dd9)? I don't mind to adjust the other plugins `authlog` and `commandhistory`.

Zawadidone:
@JSCU-CNI ?

Schamper:
There are of course different ways of approaching a problem, and it looks like we currently have two conflicting approaches with conflicting resolutions.

As far as I'm concerned, having multiple "aliases" for a plugin is a valid use-case for UX purposes.

Of course there are other solutions to this problem, and I can think of at least two that I'd like to test here:
1) Add a formal way to alias plugins (and that will be de-duplicated in e.g. wildcard searches)
2) Add a plugin grouping/category system with both a in-source maintained list of categories (by us) as well as the ability for users to maintain their own lists of categories

The first one is because it was mentioned here. The second one is already something we have had a meeting about internally. As we are still working on migrating our tickets to GitHub (I swear it's close now :smile:), I'll copy a description of the current ticket here:

> The category system is created to serve two adjacent goals:
>
> - Define a small number (3-ish) of categories which are useful for “default” investigations. These can be a useful starting point for beginning investigators, or for default cases which have nothing special. Think of categories like “default” for your bog-standard case, “volatile” for cases involving volatile data etc. 
>
> - Have the user define their own categories specific for their investigation (e.g. plugins that output artifacts that hold clues to persistency)
>
> Note that the words tags/categories can be used interchangeably: plugins are given a tag, while categories are sets of plugins. The difference is in the technical implementation.
> 
> In this case we go for the category implementation. A category is a list of strings of plugin (wildcard) names, which can be selected using a commandline parameter like --category <category_name>. This is then fed into the default plugin selection and execution mechanism.
>
 > The default categories should be defined somewhere in query.py or a related file. The user defined categories can go into .targetcfg.py as variables. These variables must be prefixed with a well known prefix, like plugin_category_<category_name>.

As we also dump plugins in bulk (but not blind-fire), this at least solves some issues we're currently having ourselves. For example, maybe we define a category for plugins that are generally quick to run, and prioritize those over ones that take a long time.

This is still a rough description, but hopefully you get the idea. Would this be something that could work for you @Zawadidone?

Zawadidone:
@Schamper yes that works for me. 

What are the issue you encounter when executing all plugins on a large amount of targets?

Poeloe:
@Zawadidone there are no real issues for us at the moment, but more improvements mainly related to efficiency. The category system comes in handy as a sort of decision points for external logic, especially for prioritisation.

For example, imagine that you have limited processing resources to work with. In that case, it is useful for your external logic to know if you're dealing with a lets say "heavy" plugin or a "light" plugin. Based on that, you can prioritise the execution of the plugins making sure that the "heavy" plugins are executed last.

Or you might at first be interested in all the plugins may give you some information about "Lateral Movement", because that is for example your main concern when starting the investigation.

PR #333, opened 2023-07-28
`target-fs` also load targets as directory

Zawadidone:


Zawadidone:
If the `Target` is a directory in the case of [KAPE](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/kape.py) or [Velociraptor](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/velociraptor.py) the tool `target-fs` logs an error and stops with execution.

```bash
target-fs  <TARGET> ls /
usage: fs.py [-h] [-K KEYCHAIN_FILE] [-Kv KEYCHAIN_VALUE] [-v] [-q] [--plugin-path PLUGIN_PATH [PLUGIN_PATH ...]] TARGET {ls,cat,walk,cp} ...
fs.py: error: Specified target <TARGET>is not a file or does not exist

# With the current change
target-fs  <TARGET> ls /
c:
sysvol
```

Zawadidone:
I am not aware why this error is logged and if this change introduces other bugs

Issue #266, opened 2023-06-02
Order of target-query output varies


mischw:
The order of the output of target-query seems to be in random order. Three executions yield different results each time:

```shell
$ target-query $IMAGE -f architecture,ips,hostname -d ';'
2023-06-02T08:28:10.577451Z [warning  ] <Target debian11.vmdk>: Can't identify filesystem: <Volume name='part_3763f00000' size=8999927808 fs=None> [dissect.target.target] 
2023-06-02T08:28:10.595940Z [warning  ] <Target debian11.vmdk>: Skipped FS type: swap, UUID=cb8add39-4d4b-4580-9560-8e1b803717f1, none [dissect.target.target] 
2023-06-02T08:28:10.595989Z [warning  ] <Target debian11.vmdk>: Unsupported mount device: /dev/sr0 /media/cdrom0 [dissect.target.target] 
<Target debian11.vmdk> x86_64-linux;['192.168.175.162'];osboxes
```

```shell
$ target-query $IMAGE -f architecture,ips,hostname -d ';'
2023-06-02T08:28:14.153905Z [warning  ] <Target debian11.vmdk>: Can't identify filesystem: <Volume name='part_3763f00000' size=8999927808 fs=None> [dissect.target.target] 
2023-06-02T08:28:14.172428Z [warning  ] <Target debian11.vmdk>: Skipped FS type: swap, UUID=cb8add39-4d4b-4580-9560-8e1b803717f1, none [dissect.target.target] 
2023-06-02T08:28:14.172478Z [warning  ] <Target debian11.vmdk>: Unsupported mount device: /dev/sr0 /media/cdrom0 [dissect.target.target] 
<Target debian11.vmdk> x86_64-linux;osboxes;['192.168.175.162']
```

```shell
$ target-query $IMAGE -f architecture,ips,hostname -d ';'
2023-06-02T08:28:19.582197Z [warning  ] <Target debian11.vmdk>: Can't identify filesystem: <Volume name='part_3763f00000' size=8999927808 fs=None> [dissect.target.target] 
2023-06-02T08:28:19.600684Z [warning  ] <Target debian11.vmdk>: Skipped FS type: swap, UUID=cb8add39-4d4b-4580-9560-8e1b803717f1, none [dissect.target.target] 
2023-06-02T08:28:19.600739Z [warning  ] <Target debian11.vmdk>: Unsupported mount device: /dev/sr0 /media/cdrom0 [dissect.target.target] 
[...]

PR #334, opened 2023-07-28
Add NamespacePlugin

cecinestpasunepipe:
(DIS-1681)


Schamper (commented):
Can you add automatic generation of shared functions that have the same (name, record) signature? So that you don't have to manually define functions such as these:

```
@export(record=BrowserDownloadRecord)
def downloads(self):
    yield from self._func("downloads")
```

cecinestpasunepipe:
> Can you add automatic generation of shared functions that have the same (name, record) signature? So that you don't have to manually define functions such as these:
> 
> ```
> @export(record=BrowserDownloadRecord)
> def downloads(self):
>     yield from self._func("downloads")
> ```

Seems that we have to do this statically, maybe using __init_subclass__ otherwise the autogenerated funcs will be listed as invalid. Also, we used to put record descriptions as docs in those functions, so we lose those then I guess?

Schamper:
> Seems that we have to do this statically, maybe using `__init_subclass__` otherwise the autogenerated funcs will be listed as invalid. Also, we used to put record descriptions as docs in those functions, so we lose those then I guess? Looking into this now.

You can set `__doc__`.

cecinestpasunepipe:
> > Also, we used to put record descriptions as docs in those functions, so we lose those then I guess? Looking into this now.
> 
> You can set `__doc__`.

Yes, I know, but the developer still needs a place to write the documentation?
Or do you mean to generate the record docs automatically? But what about any manual documentation / comments then?

Schamper:
> Yes, I know, but the developer still needs a place to write the documentation? (Of course the developer could add the docs to all the separate plugins but then we have duplicate docs?) Or do you mean to generate the record docs automatically? But what about any manual documentation / comments then?

I believe the long-term goal is to automatically generate the record descriptor documentation. The body of the docstring could be automatically generated to say something like "Convenience function to call {function} on {subplugins}".

dissect/target/plugins/browsers/edge.py
```diff
@@ -32,18 +32,3 @@ class EdgePlugin(ChromiumMixin, Plugin):
     BrowserHistoryRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
         "browser/edge/history", GENERIC_HISTORY_RECORD_FIELDS
     )
-
-    @export(record=BrowserDownloadRecord)
-    def downloads(self):
```
Schamper, line 37: Is this deletion intentional?

dissect/target/plugin.py
```diff
@@ -793,8 +796,123 @@ def _modulepath(cls) -> str:
...
+            # The method needs to have a single record descriptor as output
+            if not isinstance(getattr(subplugin_func, "__record__", None), RecordDescriptor):
+                continue
+
+            # The generic template for the aggregator method
+            def generate_aggregator(method_name):
+                def aggregator(self):
+                    yield from self._func(method_name)
+
+                aggregator.__name__ = "_aggr"
```
Schamper, line 879: What is this necessary for?

dissect/target/plugin.py
```diff
@@ -793,8 +796,123 @@ def _modulepath(cls) -> str:
...
+
+        self._subplugins = []
+        for entry in self.SUBPLUGINS:
+            try:
+                subplugin = getattr(self.target, entry)
+                self._subplugins.append(subplugin)
+            except Exception:  # noqa
+                target.log.exception("Failed to load subplugin: %s", entry)
+
+    def _func(self, func_name: str) -> None:
```
Schamper, line 822: Return type is wrong.

Schamper (changes requested):

dissect/target/plugins/browsers/edge.py
cecinestpasunepipe: This PR was still in draft mode. Unfortunately since you have started to react while it was in draft there is no means within GH anymore to specify that something is WIP, so please consider this a race-condition ;-)

dissect/target/plugin.py
cecinestpasunepipe: Also removed, artifact of a previous solution.

dissect/target/plugin.py
cecinestpasunepipe: Changed.

dissect/target/plugins/browsers/edge.py
cecinestpasunepipe: Are you sure you want to go down this road though? It looks like the reduction of LoC comes at the cost of additional complexity. In your opinion, is this worth the trade-off?

cecinestpasunepipe:
Note that the auto-generated functions will have generated docs that reflect the actual aggregation:

So

`target-query -l browser`

yields:

```
Available plugins:
    browsers:
      browser:
        browser.downloads - Return aggregated records for: chromium,chrome,iexplore,edge,firefox (output: records)
        browser.extensions - Return aggregated records for: chromium,chrome,edge (output: records)
        browser.history - Return aggregated records for: chromium,chrome,iexplore,edge,firefox (output: records)
```

dissect/target/plugins/browsers/chrome.py
```diff
@@ -1,15 +1,16 @@
...
+    BrowserPlugin,
 )
 from dissect.target.plugins.browsers.chromium import ChromiumMixin
 
 
-class ChromePlugin(ChromiumMixin, Plugin):
+class ChromePlugin(ChromiumMixin, BrowserPlugin):
     """Chrome browser plugin."""
 
     __namespace__ = "chrome"
```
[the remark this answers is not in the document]
cecinestpasunepipe: ?

tests/test_prefetch_time.py
```diff
@@ -95,5 +95,5 @@ def test_prefetch_read_filename(mocked_prefetch):
     filename = mocked_prefetch.read_filename(0x10, 0x10)
 
     mocked_fileheader.read.assert_called_with(0x10 * 2)
-    assert type(filename) == bytes
+    assert isinstance(filename, bytes)
```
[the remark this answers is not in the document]
cecinestpasunepipe: Racecondition ;-)

dissect/target/plugin.py
```diff
@@ -793,8 +797,154 @@ def _modulepath(cls) -> str:
...
+        # If this is a direct subclass of a Namespace Plugin,
+        # create a reference to the current class for indirect subclasses
+        # so that the can autogenerate aggregate methods there
+        cls.__nsplugin__ = cls
+
+    def __init_subclass_subplugin__(cls, **kwargs):
+        # Does the direct subclass already have a SUBPLUGINS attribute?
+        # if not, create this attribute on the direct subclass
+        if not getattr(cls.__nsplugin__, "SUBPLUGINS", None):
+            cls.__nsplugin__.SUBPLUGINS = set()
```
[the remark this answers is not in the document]
cecinestpasunepipe: Wouldn't that cause potential collisions?

dissect/target/plugins/browsers/chrome.py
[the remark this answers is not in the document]
cecinestpasunepipe: a.A and B.b?
This is already happening: browser.downloads, firefox.downloads

dissect/target/plugin.py
```diff
@@ -793,8 +797,154 @@ def _modulepath(cls) -> str:
...
+
+        # Register the current plugin class as a subplugin with
+        # the direct subclass of NamespacePlugin
+        cls.__nsplugin__.SUBPLUGINS.add(cls.__namespace__)
+
+        # Collect the public attrs of the subplugin
+        for subplugin_func_name in get_nonprivate_attribute_names(cls):
```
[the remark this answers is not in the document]
cecinestpasunepipe: Implemented

dissect/target/plugin.py
```diff
@@ -793,8 +797,154 @@ def _modulepath(cls) -> str:
...
+
+            # If we already have an aggregate method, skip
+            if existing_aggregator := getattr(cls.__nsplugin__, subplugin_func_name, None):
+                existing_aggregator.__subplugins__.append(cls.__namespace__)
+                continue
+
+            # The generic template for the aggregator method
+            def generate_aggregator(method_name):
+                def aggregator(self) -> Iterator[Record]:
+                    yield from self._func(method_name, aggregator.__subplugins__)
```
Schamper, line 884: Perhaps inline the `_func` logic here? I'm not sure I see the benefit anymore of having that separate.

dissect/target/plugin.py
```diff
@@ -793,8 +797,154 @@ def _modulepath(cls) -> str:
...
+            # Register the newly auto-created method
+            cls.__nsplugin__.__exports__.append(subplugin_func_name)
+            cls.__nsplugin__.__functions__.append(subplugin_func_name)
+
+    def __init_subclass__(cls, **kwargs):
+        # Upon subclassing,
+        # decide whether this is a direct subclass of NamespacePlugin
+        # If this is not the case, autogenerate aggregate methods for
+        # methods with similar record types and signatures in the
+        # direct subclass of NamespacePlugin
```
Schamper, line 939:
```suggestion
        # Upon subclassing, decide whether this is a direct subclass of NamespacePlugin
        # If this is not the case, autogenerate aggregate methods for methods with similar record types and
        # signatures in the direct subclass of NamespacePlugin
```

Schamper (changes requested):

dissect/target/plugins/browsers/edge.py
Schamper: > Are you sure you want to go down this road though? It looks like the reduction of LoC comes at the cost of additional complexity. In your opinion, is this worth the trade-off?

I'm not sure what this comment is in reply of exactly, can you elaborate?

Back on topic of this original comment, this deletion isn't a namespace function, but actual functionality of the Edge plugin, is it not? Or am I missing something?

cecinestpasunepipe: The code was deleted by accident and I put it back. It should have not been visible in the PR but I created a draft PR (not ready for review) to check the code myself.

The other comment was just a generic observation and has nothing to do with the deleted lines (that are now back).

Schamper: Ah my bad, GitHub seemed to still show it as deleted when I made that last comment.

dissect/target/plugin.py
cecinestpasunepipe: ok

dissect/target/plugin.py
```diff
@@ -964,4 +1117,4 @@ def find_plugin_functions(
                     )
                 )
 
-    return list(set(result)), invalid_funcs
+    return sorted(list(set(result)), key=lambda key: key.name), invalid_funcs
```
Schamper, line 1120: What exactly does this sorting add?

cecinestpasunepipe: It solves #266

Schamper: Wouldn't the order still be wrong? The order should be the same as specified on the command line.

dissect/target/plugin.py
```diff
@@ -965,4 +1118,4 @@ def find_plugin_functions(
                     )
                 )
 
-    return list(set(result)), invalid_funcs
+    return sorted(set(result), key=result.index), invalid_funcs
```
Schamper, line 1121: Can you add a unit test?

cecinestpasunepipe: Added

Schamper (changes requested):

dissect/target/plugin.py
```diff
@@ -793,8 +797,146 @@ def _modulepath(cls) -> str:
...
+        # Collect the public attrs of the subplugin
+        for subplugin_func_name in cls.__exports__:
+            subplugin_func = inspect.getattr_static(cls, subplugin_func_name)
+
+            # The attr need to be callable and exported
+            if not isinstance(subplugin_func, Callable):
+                continue
+
+            # The method needs to have a single record descriptor as output
+            if getattr(subplugin_func, "__output__", None) != "record":
```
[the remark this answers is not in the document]
cecinestpasunepipe: O, yes, code has changed by a previous suggestion.

dissect/target/plugin.py
```diff
@@ -886,8 +1030,20 @@ def find_plugin_functions(
...
+        exact_match = pattern in functions
+
+        # Allow for exact matches, otherwise you cannot reach documented
+        # namespace plugins like browsers.browser.downloads.
+        # You can *always* run these using the namespace/classic-style
+        # like: browser.downloads (but -l lists them in the tree for
+        # documentation purposes so it would be misleading not to allow
+        # tree access as well). Note that these tree items will never
+        # respond to wildcards though (browsers.browser.* won't work)
+        # to avoid duplicate results.
```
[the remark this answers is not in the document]
cecinestpasunepipe: I find it very hard to read but okay.

dissect/target/plugin.py
```diff
@@ -793,8 +797,146 @@ def _modulepath(cls) -> str:
...
+            # Copy the meta descriptors of the function attribute
+            copy_attrs = ["__output__", "__record__", "__doc__", "__exported__"]
+            for copy_attr in copy_attrs:
+                setattr(generated_aggregator, copy_attr, getattr(subplugin_func, copy_attr, None))
+
+            # Add subplugin to aggregator
+            generated_aggregator.__subplugins__.append(cls.__namespace__)
+
+            # Mark the function as being autogenerated
+            setattr(generated_aggregator, "__autogen__", True)
```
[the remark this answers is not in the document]
cecinestpasunepipe: It is set on the aggregation function, not the class.

dissect/target/plugin.py
```diff
@@ -844,6 +978,9 @@ def all_plugins():
         if "get_all_records" in available["exports"]:
             available["exports"].remove("get_all_records")
         modulepath = available["module"]
+        # Always skip these
+        if available["class"] in ["DefaultPlugin", "ExamplePlugin"]:
```
Schamper, line 982: I don't really like this being a string based skip, is there nowhere else this could be more accurately checked?

dissect/target/plugin.py
```diff
@@ -874,9 +1007,19 @@ def find_plugin_functions(
     a list of plugin function descriptors (including output types).
     """
     result = []
+
+    # Avoid cyclic import
+    from dissect.target.plugins.general.default import DefaultPlugin  # noqa
+    from dissect.target.plugins.general.example import ExamplePlugin  # noqa
+
+    def add_to_result(func: PluginFunction) -> None:
+        if func not in result and func.class_object not in [DefaultPlugin, ExamplePlugin]:
```
Schamper, line 1016: Why not add another `__magic__` class variable? I.e. `__skip__`?

Schamper (changes requested):

dissect/target/plugin.py
cecinestpasunepipe: I tried to avoid adding more class vars like that, but it seems inevitable.

Schamper: I agree that we should not keep adding infinite amounts of variables, but it seems cleaner than doing inline imports in this case 😄

dissect/target/plugins/general/example.py
```diff
@@ -53,7 +53,8 @@ def __init__(self, target):
             super().__init__(target)
     """
 
-    __findable__ = False
+    __findable__: bool = False
+    __skip__: bool = True
```
Schamper, line 57:
```suggestion
    # IMPORTANT: Remove these attributes when using this as boilerplate for your own plugin!
    __findable__ = False
    __skip__ = True
```
Don't think these are necessary? Also adding a small comment

dissect/target/plugins/general/default.py
```diff
@@ -3,6 +3,8 @@
 
 
 class DefaultPlugin(OSPlugin):
+    __skip__: bool = True
```
Schamper, line 6:
```suggestion
    __skip__ = True
```
Same here.

Schamper (changes requested):

dissect/target/plugins/general/example.py
cecinestpasunepipe: I was in doubt.

Schamper (approved):

PR #336, opened 2023-07-28
Add InfoPlugin

### LINKED ISSUE #330: Output OS version as record
I would like to change the current version functions of the operating systems to output the version as a record instead of text or add an additional function to output the version as a record.

How can this be implemented? 

https://github.com/fox-it/dissect.target/blob/29b54d7dbbe45cc76a9188470529e5e56cc94744/dissect/target/plugin.py#L348-L356

Zawadidone:
Close #330


Zawadidone:
The intended solution for this PR is still discussed in #330

PR #339, opened 2023-08-01
Turn `interpolation` off for the systemd configparser

Miauwkeru:
Interpolation causes the ConfigParser to try and resolve arguments with %
for example. Systemd uses those characters for certain specifiers.

(DIS-1907)


Schamper (approved):

JSCU-CNI:
Thank you for the quick fix!

PR #344, opened 2023-08-03
IIS only add logfiles if it exists in the xml

Miauwkeru:
- Add tests for different logfile notation in webservers
- Add iis log directory if it is definedi

(DIS-2116)


tests/test_plugins_apps_webservers_iis.py
```diff
@@ -10,7 +10,10 @@
 from ._utils import absolute_path
 
 
-def test_iis_plugin_iis_format(target_win_tzinfo, fs_win):
+def test_iis_plugin_iis_format(
```
Schamper, line 13: Can you add type hints?

Schamper (changes requested):

tests/data/webservers/iis/iis-applicationHost-w3c-alternative.config
Poeloe, line 1: Maybe make the log filename a bit more descriptive, such that it is more clear what is different.

I would suggest: `iis-applicationHost-w3c-additional-logFile-element.config`

tests/data/webservers/iis/iis-applicationHost-w3c-undefined-logfile-directory.config
Poeloe, line 1: Would suggest : iis-applicationHost-w3c-logFile-without-directory-attr.config

Poeloe (approved):

Schamper (approved):

PR #346, opened 2023-08-04
Allow Default/None-OS to access plugins

cecinestpasunepipe:
(DIS-2138)

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -10,6 +10,7 @@ from dissect.target.plugin import (
     get_external_module_paths,
     save_plugin_import_failure,
 )
+from dissect.target.target import Target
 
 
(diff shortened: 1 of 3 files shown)


Poeloe (changes requested):
Please include a test, since we did not notice this bug with our tests

tests/test_plugin.py
```diff
@@ -118,3 +118,12 @@ def test_find_plugin_function_unix(target_unix):
 
     assert len(found) == 1
     assert found[0].name == "os.unix.services.services"
+
+
+def test_find_plugin_function_default(target_default):
```
Schamper, line 123: Can you add type hints to the test?

Poeloe (approved):
LGTM!

Schamper (approved):

PR #349, opened 2023-08-07
Fix duplicate files upon saving dirs in target-shell

cecinestpasunepipe:
(DIS-2024)


dissect/target/tools/shell.py
```diff
@@ -663,7 +663,7 @@ def save_path(
                 dst_path = dst_path.joinpath(src_name)
                 dst_path.mkdir(parents=True, exist_ok=True)
                 log_saved_path(src_path, dst_path)
-                for path in src_path.rglob("*"):
+                for path in src_path.glob("*"):
```
Schamper, line 666: Could you add a regression test?

Schamper (changes requested):

PR #353, opened 2023-08-09
Improve tar loader for windows

JSCU-CNI:
This PR adds improved support for relative and absolute tar member name paths when dealing with windows tar filesystems.

In the future we'd like to replace this `map` function with the mapping functionality of `TarFilesystem`. For now this is a quick fix to improve tar loader detection on our end.


dissect/target/loaders/tar.py
```diff
@@ -52,8 +53,10 @@ def map(self, target: target.Target) -> None:
                 volume = volumes["/"]
                 mname = member.name
             else:
-                if not member.name.startswith("/sysvol"):
+                if not "sysvol" in member.name:
```
Schamper, line 56: I feel like a contains check is a bit too broad. Can you make this more strict?

dissect/target/loaders/tar.py
```diff
@@ -42,7 +42,8 @@ def map(self, target: target.Target) -> None:
             if member.name == ".":
                 continue
 
-            if not member.name.startswith("fs/") and not member.name.startswith("/sysvol"):
+            sysvols = ["/fs", "fs/", "/sysvol", "sysvol/"]
+            if not any([member.name.startswith(s) for s in sysvols]):
```
Schamper, line 46:
```suggestion
            if not member.name.startswith(("/fs", "fs/", "/sysvol", "sysvol/")):
```
`startswith` also accepts a tuple of prefixes to check.

Schamper (changes requested):

Schamper (dismissed):

Schamper (commented):
Missed the linting error in the CI. If you can fix that it's good to go.

Schamper (approved):

PR #355, opened 2023-08-13
Zip loader for Velociraptor packages

OlafHaalstra:
Implemented a zip loader for Velociraptor packages.

Currently works on Velociraptor packages that have been created with release >=0.6.7

Still looking for advise on:
* How to implement Linux / OSX filesystems
* Combining both Velociraptor parsers. I have chosen to create a new loader to support the zips since parsing zips might punish performance. Can probably also be implemented in such a way that it supports both ZIPs and regular folders


OlafHaalstra:
@Zawadidone it would be nice if you could quickly review this as well given your history with parsing Velociraptor

dissect/target/loaders/velociraptorzip.py
```diff
@@ -0,0 +1,89 @@
...
+
+            entry_cls = ZipFilesystemDirectoryEntry if member.is_dir() else ZipFilesystemEntry
+            entry = entry_cls(volume, fsutil.normpath(mname), member)
+            volume.map_file_entry(entry.path, entry)
+
+        for vol_name, vol in volumes.items():
+            loaderutil.add_virtual_ntfs_filesystem(
+                target,
+                vol,
+                usnjrnl_path="$Extend/$UsnJrnl%3A$J",
```
Zawadidone, line 84: The path `$Secure%3A$SDS` is part of the NTFS filesystem (https://github.com/Velocidex/velociraptor/pull/2849).

Zawadidone:
I have tested the Velociraptor features of version https://github.com/Velocidex/velociraptor/releases/tag/v0.7.0.

| Velociraptor Artifact | Options | Result | Fix |
|---|---|---|---|
| Windows.KapeFiles.Targets |  `UseAutoAccessor:Y,_SANS_Triage:Y,VSSAnalysisAge:365` | The `ntfs_vss` which adds drives to `uploads/ntfs_vss/`, e.g. `uploads/ntfs_vss/%5C%5C.%5CC%3A/` | Map/mount the drivers/folders in  `uploads/ntfs_vss/` |
| Windows.KapeFiles.Targets |  `UseAutoAccessor:N,_SANS_Triage:Y,VSSAnalysisAge:365` | The `ntfs_vss` which adds drives to `uploads/ntfs_vss/`, e.g. `uploads/ntfs_vss/%5C%5C.%5CC%3A/`, but in case | Map/mount the drivers/folders in  `uploads/ntfs_vss/` |
| Windows.KapeFiles.Targets |  `UseAutoAccessor:N,_SANS_Triage:Y,VSSAnalysisAge:0` | Depending on the path the accessors `lazy_ntfs` or `ntfs` is used | Map/mount drives in `uploads/<accessor>/` |
| Windows.KapeFiles.Targets |  `UseAutoAccessor:Y,_SANS_Triage:Y,VSSAnalysisAge:0` | Depending on the path the accessors `lazy_ntfs` or `ntfs` is used | Map/mount drives in `uploads/<accessor>/` |

For Linux, MacOS, etc see the table below.

| Velociraptor Artifact | Options | Result | Fix |
|---|---|---|---|
| Generic.Collectors.File |  `root:/, accessor:file` | The root directory starts at `uploads/file/` | Map/mount the root directory `uploads/file/` |
| Generic.Collectors.File |  `root:/, accessor:file` | The root directory starts at `uploads/auto/` | Map/mount the root directory `uploads/auto/` |

I have created a PR (https://github.com/fox-it/dissect.target/pull/358) to fix this in the `VelociraptorLoader` but there is something wrong with the acquired files related to the NTFS filesystem (https://github.com/fox-it/dissect.target/pull/358#issuecomment-1678039676).

tests/test_loaders_velociraptorzip.py
```diff
@@ -0,0 +1,44 @@
...
+    with open(absolute_path("data/mft.raw"), "rb") as fh:
+        (root / "uploads/ntfs/%5C%5C.%5CC%3A/$MFT").write_bytes(fh.read(10 * 1025))
+
+    # Add one record so we can test if it works
+    data = bytes.fromhex(
+        "5800000002000000c100000000000100bf000000000001002003010000000000"
+        "6252641a86a4d7010381008000000000000000002000000018003c0069007300"
+        "2d00310035005000320036002e0074006d00700000000000"
+    )
+    (root / "uploads/ntfs/%5C%5C.%5CC%3A/$Extend/$UsnJrnl%3A$J").write_bytes(data)
```
Zawadidone, line 33: Could the tests be extended as shown in https://github.com/fox-it/dissect.target/pull/358/files#diff-99a977c474ffb59498d39b478cac2cba2abc63f4686d314ee3c6f155ffe286ee, because when using the VSS Analyzer the accessor `ntfs_vss` is used to acquire files related to NTFS 

```bash
ls */uploads/*/*/

autoaccessor-N.exe/uploads/ntfs_vss/%5C%5C.%5CC%3A/:
'$Boot'  '$Boot.idx'  '$Extend'  '$LogFile'  '$MFT'  '$Recycle.Bin'  '$Secure%3A$SDS'   ProgramData   Users   Windows

autoaccessor-Y/uploads/ntfs_vss/%5C%5C.%5CC%3A/:
'$Boot'  '$Boot.idx'  '$Extend'  '$LogFile'  '$MFT'  '$Recycle.Bin'  '$Secure%3A$SDS'   ProgramData   Users   Windows
```

Schamper:
> Combining both Velociraptor parsers. I have chosen to create a new loader to support the zips since parsing zips might punish performance. Can probably also be implemented in such a way that it supports both ZIPs and regular folders

The `zipfile` module partially implements the `pathlib` API through the `zipfile.Path` object (which I see you're already using. Perhaps the logic in the original Velociraptor loader could be tweaked to be generic on a `pathlib.Path`-like object, so that it works on both zip files and directories.

> How to implement Linux / OSX filesystems

Combining the loaders probably solves this problem too?

Zawadidone:
@OlafHaalstra is it possible to add the ZIP Loader to the updated VelociraptorLoader (https://github.com/fox-it/dissect.target/pull/358) that supports version Velociraptor version 0.7.0?

OlafHaalstra:
With the new commit I tried to combine both loaders into a single file, due to the fact that `map_dirs` is not able to correctly parse the correct Windows drives I reimplemented it partially inside the Velociraptor loader to support the `%5C%5C.%5CD%3A` and `C%3A` style of writing drives. Potentially we can also move this functionality to expand `map_dirs` in `dir.py`.

Ideally you would combine more code for zips and non zips but I wasn't fully able to do so. The `zip` folder parsing follows the way `tar` is also parsed. Folders are parsed the same way as they were with the exception of determining the correct drive.

Schamper:
@OlafHaalstra I've made some changes to re-use more existing code. The unit tests pass, can you test if this works on real Velociraptor ZIPs too?

I looked briefly into making `DirFilesystem` compatible with `zipfile.Path` but unfortunately it's missing too much functionality for that.

Zawadidone:
@OlafHaalstra it works thanks for the feature!

OlafHaalstra:
@Schamper your improvements work on the real zip from Velociraptor! Thanks 😄

OlafHaalstra:
@DissectBot agree

tests/test_loaders_velociraptor.py
```diff
@@ -79,21 +78,117 @@ def test_velociraptor_loader_windows_ntfs(paths: list[str], mock_target: Target,
...
+            [
+                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/",
+                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/$Extend",
+                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/windows/system32",
+                "uploads/lazy_ntfs/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1",
+            ]
+        ),
+    ],
+)
+def test_velociraptor_loader_windows_ntfs_zip(paths: list[str], mock_target: Target, tmp_path: Path) -> None:
```
Poeloe, line 115: I would suggest to add the following helper function at the beginning of the file:

```
def create_paths(sub_dir: str) -> list[str]:
    return [
        f"uploads/{sub_dir}/%5C%5C.%5CC%3A/",
        f"uploads/{sub_dir}/%5C%5C.%5CC%3A/$Extend",
        f"uploads/{sub_dir}/%5C%5C.%5CC%3A/windows/system32",
        f"uploads/{sub_dir}/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1",
    ]
```

And then you can do the following for both the `test_velociraptor_loader_windows_ntfs` and `test_velociraptor_loader_windows_ntfs_zip` test:    

```
@pytest.mark.parametrize(
    "sub_dir", 
    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs"],
)
def test_velociraptor_loader_windows_ntfs(sub_dir: str, mock_target: Target, tmp_path: Path) -> None:
    paths = create_paths(sub_dir)
```

```
@pytest.mark.parametrize(
    "sub_dir", 
    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs"],
)
def test_velociraptor_loader_windows_ntfs_zip(sub_dir: str, mock_target: Target, tmp_path: Path) -> None:
    paths = create_paths(sub_dir)
```

This reduces the amount of lines and makes it more readable in my opinion :)

Poeloe (changes requested):

tests/test_loaders_velociraptor.py
OlafHaalstra: Would also be nice to do it on a `class` level but I believe that the way the testing is setup that's not supported, see: https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

OlafHaalstra: I updated the testing strategy as you suggested!

tests/test_loaders_velociraptor.py
```diff
@@ -8,61 +9,36 @@
...
-                "uploads.json",
-                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/",
-                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/$Extend",
-                "uploads/lazy_ntfs/%5C%5C.%5CC%3A/windows/system32",
-                "uploads/lazy_ntfs/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1",
-            ]
-        ),
-    ],
+    "sub_dir",
+    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs"],
```
Poeloe, line 23: Since I see that "auto" has been added for Windows, it would be nice to add it to the test as well.

```suggestion
    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs", "auto"],
```

tests/test_loaders_velociraptor.py
```diff
@@ -76,24 +52,88 @@ def test_velociraptor_loader_windows_ntfs(paths: list[str], mock_target: Target,
     assert len(mock_target.filesystems) == 3
 
 
+@pytest.mark.parametrize(
+    "sub_dir",
+    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs"],
```
Poeloe, line 57: Same here

```suggestion
    ["mft", "ntfs", "ntfs_vss", "lazy_ntfs", "auto"],
```

Poeloe (changes requested):

Poeloe (approved):
LGTM!

PR #364, opened 2023-08-18
Add Modules/Lsmod plugin

DevJoost:
(DIS-2092)

Adds a module plugin with two different types of output (records and text in lsmod format). Plugin can be used to list information about current loaded kernel modules.


tests/data/unix/module/modulea/coresize
Schamper, line 1: Could you move these test files to `tests/data/plugins/os/unix/modules`?

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+class ModulePlugin(Plugin):
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self._module_paths = list(self.target.fs.glob("/sys/module/*"))
+
+    def check_compatible(self) -> bool:
+        return len(self._module_paths) > 0
+
+    @internal
+    def iterate_modules(self) -> Iterator[Module]:
```
Schamper, line 38:
```suggestion
    def _iterate_modules(self) -> Iterator[Module]:
```
And rename accordingly where used.

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+    name: Optional[str] = None
+    size: Optional[int] = None
+    refcnt: Optional[int] = None
+    modules_referred: Optional[list[str]] = None
+
+
+class ModulePlugin(Plugin):
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self._module_paths = list(self.target.fs.glob("/sys/module/*"))
```
Schamper, line 32:
```suggestion
        self._module_paths = list(self.target.fs.path("/sys/module").iterdir())
```

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+                module = Module(module_path, module_folder.name)
+                module.size = int(module_folder.joinpath("coresize").read_text())
+                module.refcnt = int(module_folder.joinpath("refcnt").read_text())
+                if module_folder.joinpath(
+                    "holders"
+                ).exists():  # Normally the holders folder should exist, however Acquire doesn't acquire a folder if its empty
+                    module.modules_referred = [item.name for item in module_folder.joinpath("holders").iterdir()]
+                else:
+                    module.modules_referred = []
+                yield module
```
Schamper, line 51:
```suggestion
            if module_path.joinpath("initstate").exists():
                # Normally the holders folder should exist, however Acquire currently doesn't collect a folder if its empty
                holders = []
                if (holders_path := module_path.joinpath("holders")).exists():
                    holders = [item.name for item in holders_path.iterdir()]

                yield Module(
                      module_path,
                      module_path.name,
                      int(module_path.joinpath("coresize").read_text()),
                      int(module_path.joinpath("refcnt").read_text()),
                      holders,
                )
```

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+                source=module.path,
+            )
+
+    @export(output="yield")
+    def lsmod(self):
+        """Return information about active kernel modules in lsmod format"""
+        lsmod_output = f"{'Module ':<28} {'Size':<7}  Used by\n"
+        for module in self.iterate_modules():
+            lsmod_output += f"{module.name:<28} {module.size:<7}  {module.refcnt} {','.join(module.modules_referred)}\n"
+        yield lsmod_output
```
Schamper, line 71:
```suggestion
        yield f"{'Module ':<28} {'Size':<7}  Used by\n"
        for module in self.iterate_modules():
            yield f"{module.name:<28} {module.size:<7}  {module.refcnt} {','.join(module.modules_referred)}\n"
```

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+from dissect.target.plugin import Plugin, export, internal
+from dissect.target.target import Target
+
+ModuleRecord = TargetRecordDescriptor(
+    "linux/module",
+    [
+        ("string", "name"),
+        ("varint", "size"),
+        ("varint", "refcount"),
+        ("stringlist", "modules_referred"),
```
Schamper, line 14:
```suggestion
        ("string[]", "modules_referred"),
```
Also, since it's originating from a directory called `holders`, does it not make more sense to rename this field to `holders`? Or `used_by`, to adhere close to the `lsmod` output?

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
+from dataclasses import dataclass
+from typing import Iterator, Optional
+
+from dissect.target.helpers.record import TargetRecordDescriptor
+from dissect.target.plugin import Plugin, export, internal
+from dissect.target.target import Target
+
+ModuleRecord = TargetRecordDescriptor(
+    "linux/module",
```
Schamper, line 9:
```suggestion
    "unix/module",
```
I think?

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+
+@dataclass
+class Module:
+    path: Optional[str] = None
+    name: Optional[str] = None
+    size: Optional[int] = None
+    refcnt: Optional[int] = None
+    modules_referred: Optional[list[str]] = None
```
Schamper, line 26:
```suggestion
    path: str
    name: str
    size: int
    refcnt: int
    modules_referred: list[str]
```
As far as I can tell, none of these fields are actually optional and you always populate all of them.

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+from dissect.target.target import Target
+
+ModuleRecord = TargetRecordDescriptor(
+    "linux/module",
+    [
+        ("string", "name"),
+        ("varint", "size"),
+        ("varint", "refcount"),
+        ("stringlist", "modules_referred"),
+        ("uri", "source"),
```
Schamper, line 15:
```suggestion
        ("path", "source"),
```

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,71 @@
...
+            yield ModuleRecord(
+                name=module.name,
+                size=module.size,
+                refcount=module.refcnt,
+                modules_referred=module.modules_referred,
+                source=module.path,
+            )
+
+    @export(output="yield")
+    def lsmod(self):
```
Schamper, line 66:
```suggestion
    def lsmod(self) -> Iterator[str]:
```

Schamper (changes requested):

dissect/target/plugins/os/unix/modules.py
DevJoost: Agreed. Both name and type fixed.

dissect/target/plugins/os/unix/modules.py
```diff
@@ -0,0 +1,70 @@
...
+
+class ModulePlugin(Plugin):
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self._module_paths = list(self.target.fs.path("/sys/module").iterdir())
+
+    def check_compatible(self) -> bool:
+        return len(self._module_paths) > 0
+
+    @internal
```
Schamper, line 37:
```suggestion
```
The `@internal` decorator is only necessary if you wish to expose a function to the Python API (`t._iterate_modules` in this case) but not be usable by `target-query`.

Schamper (changes requested):

Schamper (approved):

dissect/target/plugins/os/unix/modules.py
Schamper, line 1: One last thing actually. I believe this is pretty specific to Linux, not other Unix systems. Can you move the plugin to the `linux` directory? And move the test files accordingly too.

Schamper (changes requested):

dissect/target/plugins/os/unix/modules.py
DevJoost: I'll leave that open for @Horofic. Discussed this topic with him and we eventually ended up with placing it in the unix folder.

Horofic: Not sure how, (kernel)modules are handled for bsd-based systems. Imo, this plugin could be extended to also yield results from those systems. The `lsmod` could be moved to the `linux` though.

DevJoost: For now moving the entire plugin to `linux`. We could at one point choose to add BSD modules to de modules plugin.

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -0,0 +1,70 @@
...
+    @export(record=ModuleRecord)
+    def sysmodules(self) -> Iterator[ModuleRecord]:
+        """Return information about active kernel modules."""
+        for module in self._iterate_modules():
+            yield ModuleRecord(
+                name=module.name,
+                size=module.size,
+                refcount=module.refcnt,
+                used_by=module.used_by,
+                source=module.path,
```
Schamper, line 62:
```suggestion
                source=module.path,
                _target=self.target,
```

Schamper (changes requested):

dissect/target/plugins/os/unix/linux/modules.py
DevJoost: Good catch

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -0,0 +1,70 @@
+from dataclasses import dataclass
+from typing import Iterator
+
+from dissect.target.helpers.record import TargetRecordDescriptor
+from dissect.target.plugin import Plugin, export
+from dissect.target.target import Target
+
+ModuleRecord = TargetRecordDescriptor(
+    "unix/module",
```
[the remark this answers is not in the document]
Horofic:
```suggestion
    "linux/module",
```

Horofic (changes requested):

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -0,0 +1,71 @@
...
+    used_by: list[str]
+
+
+class ModulePlugin(Plugin):
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self._module_paths = list(self.target.fs.path("/sys/module").iterdir())
+
+    def check_compatible(self) -> bool:
+        return len(self._module_paths) > 0
```
pyrco, line 35: Check module should raise an `UnsupportedPluginError` instead of returning a `bool`:

```suggestion
    def check_compatible(self) -> None:
        if not len(self._module_paths) > 0:
            raise UnsupportedPluginError("No module paths found.")
```

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -0,0 +1,71 @@
...
+    name: str
+    size: int
+    refcnt: int
+    used_by: list[str]
+
+
+class ModulePlugin(Plugin):
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self._module_paths = list(self.target.fs.path("/sys/module").iterdir())
```
pyrco, line 32:
```suggestion
        self._module_paths = self.target.fs.path("/sys/module").iterdir()
```

I'd not resolve the full iterator here, just keep it as is.
The `check_compatible()` could then become:

```python
        try:
            next(self._module_paths) :
        except StopIteration:
            raise UnsupportedPluginError("No module paths found.")
```

Schamper: Wouldn't this advance the generator, and you'd miss the first entry in the plugin itself? You'd also only be able to run the plugin once.

pyrco: Hmm, yes it would. Then just define the base path in `__init__` and `iterdir()` separately in each function.

DevJoost: Alright I tried to follow the conversation. I think I fixed it by adding a base path in `__init__` and `next(base_path.iterdir())` in check_compatible. In the plugin I iterated again over the base path with `iterdir()`

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -29,13 +30,14 @@ class Module:
 class ModulePlugin(Plugin):
     def __init__(self, target: Target):
         super().__init__(target)
-        self._module_paths = list(self.target.fs.path("/sys/module").iterdir())
+        self._module_base_path = self.target.fs.path("/sys/module")
 
     def check_compatible(self) -> bool:
-        return len(self._module_paths) > 0
+        if not next(self._module_base_path.iterdir()):
```
Schamper, line 36:
```suggestion
        if not self._module_base_path.is_dir() or not next(self._module_base_path.iterdir(), None):
```

Schamper (changes requested):

Schamper (approved):

dissect/target/plugins/os/unix/linux/modules.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def check_compatible(self) -> bool:
+        if not self._module_base_path.is_dir() or not next(self._module_base_path.iterdir(), None):
+            raise UnsupportedPluginError("No module paths found.")
+
+    def _iterate_modules(self) -> Iterator[Module]:
+        for module_path in self._module_base_path.iterdir():
+            if module_path.joinpath("initstate").exists():
+                # Normally the holders folder should exist,
+                # however Acquire currently doesn't collect a folder if its empty
```
pyrco, line 43:
```suggestion
```

This is no longer the case, so the comment can be removed.

pyrco (changes requested):

pyrco (approved):

PR #375, opened 2023-08-25
Mount volatile Linux directories when running on a local target

pyrco:
(DIS-1965)

### DIFF
--- a/dissect/target/loaders/local.py
+++ b/dissect/target/loaders/local.py
@@ -13,8 +13,18 @@ from dissect.target.filesystems.dir import DirectoryFilesystem
 from dissect.target.helpers.utils import parse_path_uri
 from dissect.target.loader import Loader
 
+SOLARIS_DEV_DIR = Path("/dev/dsk")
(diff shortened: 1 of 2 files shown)


dissect/target/loaders/local.py
```diff
@@ -59,21 +69,30 @@ def map(self, target):
...
     Iterate through /dev and match raw device names (not partitions).
+
+    /proc and /sys are mounted if they exists, allowing access to volatile files.
     """
-    for drive in Path("/dev").iterdir():
+    for drive in LINUX_DEV_DIR.iterdir():
         if LINUX_DRIVE_REGEX.match(drive.name):
             _add_disk_as_raw_container_to_target(drive, target)
 
+    # mount volatile directories
```
cecinestpasunepipe, line 82: Instead of telling what this code does, I recommend to explain why we do it...

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #377, opened 2023-08-25
Fix/make journal plugin compatible with new expression parser

cecinestpasunepipe:


dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -214,7 +214,7 @@
     le64_t      monotonic;
     sd_id128_t  boot_id;
     le64_t      xor_hash;
-    EntryItem   items[size - 64 / 16];                // The size minus the previous members divided by the size of the items
+    EntryItem   items[(size - 64) / 16];                // The size minus the previous members divided by the size of the items
```
Schamper, line 217:
```suggestion
    EntryItem   items[(size - 64) / 16];                  // The size minus the previous members divided by the size of the items
```

dissect/target/plugins/os/unix/log/journal.py
```diff
@@ -239,7 +239,7 @@
     uint8_t     reserved[6];
     le64_t      size;
     le64_t      next_entry_array_offset;
-    le64_t      entry_object_offsets[size - 24 / 8];  // The size minus the previous members divided by the size of the offset
+    le64_t      entry_object_offsets[(size - 24) / 8];  // The size minus the previous members divided by the size of the offset
```
Schamper, line 242:
```suggestion
    le64_t      entry_object_offsets[(size - 24) / 8];    // The size minus the previous members divided by the size of the offset
```

Schamper (changes requested):
And can you indent the other comments in the definition one level so that they're all aligned again? I chose to increase the indentation rather than decrease because otherwise it'd be attached to the field on line 242.

Schamper:
Can you also change all the other comment indentations to align?

PR #380, opened 2023-08-31
Harmonize Compatibility Checks

cecinestpasunepipe:
(DIS-2206)

### DIFF
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -517,7 +517,7 @@ class Target:
 
         if check_compatible:
             try:
-                if p.is_compatible() is False:
+                if not p.is_compatible():
                     self.send_event(Event.INCOMPATIBLE_PLUGIN, plugin_cls=plugin_cls)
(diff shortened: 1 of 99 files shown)


dissect/target/target.py
```diff
@@ -329,7 +329,7 @@ def _load_child_plugins(self) -> None:
                 continue
 
             try:
-                if child_plugin.check_compatible() is False:
+                if child_plugin.is_compatible() is False:
```
pyrco, line 332:
```suggestion
                if not child_plugin.is_compatible():
```

dissect/target/target.py
```diff
@@ -517,7 +517,7 @@ def add_plugin(
 
         if check_compatible:
             try:
-                if p.check_compatible() is False:
+                if p.is_compatible() is False:
```
pyrco, line 520:
```suggestion
                if not p.is_compatible():
```

pyrco (changes requested):

pyrco (approved):

PR #384, opened 2023-09-02
Add support for Citrix Netscaler access & error logs

MaxGroot:
This pull request adds support for access logs as Citrix produces them by adding a few regexes to the Apache access plugin. In doing so, the `WebserverAccessLogRecord` descriptor is extended with the fields `local_ip` and `pid`. 

Moreover, this PR adds a generic WebserverErrorLogRecord and extends the Apache plugin with a `error` method. To keep the plugin readable I changed most regex strings to a multi-line version with comments that are parsed using `re.VERBOSE`. 

It might be worth considering to 'parse' the configuration files for Apache to construct regexes dynamically. However, those files might not always be available. Having said that, I'm unsure whether it is desirable to keep adding / extending regexes to keep covering other log formats in the future. Open to feedback in that regard.


dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
+        self.access_log_paths, self.error_log_paths = self.get_log_paths()
 
     def check_compatible(self) -> None:
-        if not len(self.log_paths):
+        if not len(self.access_log_paths) and not len(self.error_log_paths):
             raise UnsupportedPluginError("No Apache directories found")
 
     @plugin.internal
-    def get_log_paths(self) -> list[Path]:
+    def get_log_paths(self) -> Tuple[list[Path], list[Path]]:
```
Schamper, line 200:
```suggestion
    def get_log_paths(self) -> tuple[list[Path], list[Path]]:
```

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -104,56 +238,115 @@ def get_log_paths(self) -> list[Path]:
...
+
+                # Both error_source and error_code follow the same logformat. When both are present, the error source
+                # goes before the client and the error code goes after. However, it is also possible that only the error
+                # code is available, in which case it is situated *after* the client. In such situations our regex match
+                # has assigned the variables wrong, and we need to do a swap.
+                if error_source and error_code is None:
+                    error_source, error_code = error_code, error_source
+
+                yield WebserverErrorLogRecord(
+                    ts=datetime.strptime(log["ts"], "%a %b %d %H:%M:%S.%f %Y"),
```
Schamper, line 325: Does this include a timezone? Otherwise can you add one (UTC or local time)?

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -104,56 +238,115 @@ def get_log_paths(self) -> list[Path]:
...
+                    response_time_ms=response_time,
+                    source=path,
+                    _target=self.target,
+                )
+            except Exception as e:
+                self.target.log.warning("An error occured parsing Apache log file %s: %s", path, str(e))
+                self.target.log.debug("", exc_info=e)
+
+    @plugin.export(record=WebserverAccessLogRecord)
+    def error(self) -> Iterator[WebserverAccessLogRecord]:
```
Schamper, line 301:
```suggestion
    @plugin.export(record=WebserverErrorLogRecord)
    def error(self) -> Iterator[WebserverErrorLogRecord]:
```

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -104,56 +238,115 @@ def get_log_paths(self) -> list[Path]:
...
+                    self.target.log.warning(
+                        "Could not match Apache log format %s for log line: %s (%s)", fmt, line, path
+                    )
+                    continue
+
+                log = match.groupdict()
+                response_time = log.get("response_time")
+                if response_time:
+                    response_time = apache_response_time_to_ms(response_time)
+                yield WebserverAccessLogRecord(
```
Schamper, line 279:
```suggestion

                yield WebserverAccessLogRecord(
```

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -3,25 +3,121 @@
 import re
 from datetime import datetime
 from pathlib import Path
-from typing import Iterator, Optional
+from typing import Iterator, Optional, Tuple
```
Schamper, line 6:
```suggestion
from typing import Iterator, Optional
```

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -104,56 +238,115 @@ def get_log_paths(self) -> list[Path]:
...
+                    message=log["message"],
+                    source=path,
+                    _target=self.target,
+                )
+
+            except Exception as e:
+                self.target.log.warning("An error occured parsing Apache log file %s: %s", path, str(e))
+                self.target.log.debug("", exc_info=e)
+
+    def _iterate_log_lines(self, paths: list[Path]) -> Iterator[Tuple[str, Path]]:
```
Schamper, line 341:
```suggestion
    def _iterate_log_lines(self, paths: list[Path]) -> Iterator[tuple[str, Path]]:
```

tests/test_plugins_apps_webservers_apache.py
```diff
@@ -165,6 +213,57 @@ def test_plugins_apps_webservers_apache_config_commented_logs(target_unix, fs_un
...
-    log_paths = target_unix.apache.get_log_paths()
-    assert str(log_paths[0]) == "/custom/log/location/old.log"
-    assert str(log_paths[1]) == "/custom/log/location/new.log"
+    access_log_paths, error_log_paths = target_unix.apache.get_log_paths()
+
+    assert str(access_log_paths[0]) == "/custom/log/location/old.log"
+    assert str(access_log_paths[1]) == "/custom/log/location/new.log"
+
+
+def test_plugins_apps_webservers_apache_error_txt(target_unix, fs_unix):
```
Schamper, line 222: Can you add some type hints?

Schamper (changes requested):

Schamper:
> It might be worth considering to 'parse' the configuration files for Apache to construct regexes dynamically. However, those files might not always be available. Having said that, I'm unsure whether it is desirable to keep adding / extending regexes to keep covering other log formats in the future. Open to feedback in that regard.

I was thinking about this too. Besides keeping some defaults, this might be worth pursuing. Perhaps @JSCU-CNI has some thoughts on that, since they originally wrote this plugin.

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -3,25 +3,121 @@
...
+        rf"""
+        (?P<remote_ip>.*?)          # Client IP address of the request.
+        \s
+        ->
+        \s
+        (?P<local_ip>.*?)           # Local IP of the Netscaler.
+        \s
+        (?P<remote_logname>.*?)     # Remote logname (from identd, if supplied).
+        \s
+        (?P<remote_user>.*?)        # Remote user if the request was authenticated.
```
Poeloe, line 106: Maybe you can think of an elegant way to re use the `REMOTE_REGEX` pattern here, since it is quite similar. So use a function that returns a specific version based on a boolean or something in that direction.

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
+
+
+        combined_resptime                   = '127.0.0.1 - - [19/Dec/2022:17:25:12 +0100] [69195] "GET / HTTP/1.1" 200
+                                               18705 "-" "curl/7.78.0" "Time: 49835 microsecs"'
+
+    """
+    splitted_line = line.split(" ")
+    first_part = splitted_line[0]
+    second_part = splitted_line[1]
     if ":" in first_part and "." in first_part:
```
Poeloe, line 146: Is it necessary to also check for the `.`? Cannot think of a scenario (other than a corrupt file)

Or maybe you mean `second_part` instead, to check for an IP-address?

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
         # does not start with IP, hence it must be a vhost typed log
         return LogFormat.VHOST_COMBINED
+    elif second_part == "->":
+        # Citrix-like
+        return LogFormat.CITRIX_NETSCALER_COMBINED_RESPONSE_TIME_WITH_HEADERS
     elif line[-1:] == '"':
```
Poeloe, line 152: Not your code, but isn't the `:` unnecessary since your getting the last element with -1?

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
-        # ends with a quotation mark, meaning three is a user agent
+        if "Time: " in line:
+            return LogFormat.CITRIX_NETSCALER_COMBINED_RESPONSE_TIME
+        # ends with a quotation mark but does not contain a response time, meaning there is only a user agent
         return LogFormat.COMBINED
     elif line[-1:].isdigit():
```
Poeloe, line 157: Same here

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
+                                               "GET / HTTP/1.1" 200 712 "-" "Mozilla/5.0
+                                               (Windows NT 10.0; Win64; x64); AppleWebKit/537.36 (KHTML, like Gecko)
+                                               Chrome/108.0.0.0 Safari/537.36" "Time: 11264 microsecs"'
+
+
+        combined_resptime                   = '127.0.0.1 - - [19/Dec/2022:17:25:12 +0100] [69195] "GET / HTTP/1.1" 200
+                                               18705 "-" "curl/7.78.0" "Time: 49835 microsecs"'
+
+    """
+    splitted_line = line.split(" ")
```
Poeloe, line 143: You can just use `split()` here, since it splits on spaces by default (and I think you don't mind the fact that multiple consecutive whitespaces are seen as one separator ;)) -> https://docs.python.org/3/library/stdtypes.html#str.split
```suggestion
    splitted_line = line.split()
```

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -32,43 +128,76 @@ def infer_log_format(line: str) -> Optional[LogFormat]:
...
+        if "Time: " in line:
+            return LogFormat.CITRIX_NETSCALER_COMBINED_RESPONSE_TIME
+        # ends with a quotation mark but does not contain a response time, meaning there is only a user agent
         return LogFormat.COMBINED
     elif line[-1:].isdigit():
         return LogFormat.COMMON
     return None
 
 
+def apache_response_time_to_ms(time_str: str) -> int:
```
Poeloe, line 162: You're not always returning an `int` here, due to the division by 1000 case you're dealing with `microsecs`. So either change the return type to `Union[int, float]` or parse the `microsecs` case to an `int`

Poeloe (changes requested):
Did not entirely finished my review, will continue at a later time!

Poeloe (dismissed):

JSCU-CNI:
>>    It might be worth considering to 'parse' the configuration files for Apache to construct regexes dynamically. However, those files might not always be available. Having said that, I'm unsure whether it is desirable to keep adding / extending regexes to keep covering other log formats in the future. Open to feedback in that regard.
>
>I was thinking about this too. Besides keeping some defaults, this might be worth pursuing. Perhaps @JSCU-CNI has some thoughts on that, since they originally wrote this plugin.

Thanks for the mention. Deriving the log format from Apache config files definitely sounds like a nice to have feature to us. I would propose to keep the current Apache regexes if no config file is found on a target.

On another note, we feel the current ApachePlugin would get a little bloated with Citrix functionality once this is merged. 
Looking at the current proposal of this extension of the Apache webserver plugin, we wonder if a new plugin class (eg. CitrixPlugin) would be beneficial. This plugin could inherit the current ApachePlugin and build on existing logic, while keeping Apache specific logic separate.

dissect/target/plugins/apps/webservers/apache.py
```diff
@@ -104,56 +238,115 @@ def get_log_paths(self) -> list[Path]:
...
                         log_path = line.split("CustomLog")[1].strip().split(" ")[0].replace('"', "")
                         custom_log = self.target.fs.path(log_path)
-                        log_paths.extend(
-                            path for path in custom_log.parent.glob(f"{custom_log.name}*") if path not in log_paths
-                        )
+                        if "error" in log_path:
+                            error_log_paths.extend(
+                                path
+                                for path in custom_log.parent.glob(f"{custom_log.name}*")
+                                if path not in error_log_paths
```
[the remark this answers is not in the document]
MaxGroot: Refactored into sets, however this does mean that does change the ordering of the values. In the old version, the values of `access_log_paths` would be sorted by when they were encountered in the configuration. Now it's sorted alphabetically. I don't see a problem with this myself (the log paths themselves aren't returned to end-users), but still thought it's worth mentioning.

dissect/target/plugins/apps/webservers/apache.py
MaxGroot: I think with vhosts, you have something like `example.com:80` as the first part of a line, which contains both a `:` and a `.`

dissect/target/plugins/apps/webservers/apache.py
MaxGroot: I agree it's quite similar, but didn't see a way of re-using components here. Considering the regex pattern has been moved to a separate file, I think the code overlap is a little less glaring. Still open to suggestions on this though.

MaxGroot:
Thanks for your input, I agree that the plugin would get too bloated with Citrix-specific stuff with the PR in its initially proposed form. Subclassing the `ApachePlugin` with a Citrix-specific plugin seems like the right approach. Apologies it took a while, I went out for a bit to get some milk.

To be able to subclass `ApachePlugin` into a `CitrixWebserverPlugin`, substantial changes were necessary to the `ApachePlugin`. I initially considered splitting this PR into one for refactoring the `ApachePlugin` while adding error log support, and creating a second PR for the `CitrixWebserverPlugin`. However, I think some of the design choices can best be discussed with having both the superclass and the subclass in the same PR.

To make this work, some things were moved around or changed:

* `infer_access_log_format` is now a static method of `ApachePlugin`. This way, the `CitrixWebserverPlugin` can override it for inferring citrix-specific log formats.
* The `LogFormat` enum was removed. As enums are not extendible, the `CitrixWebserverPlugin` could not return a log format of the same type as its superclass. Instead, log formats are now noted using a `LogFormat` named tuple, which contain both a name and a regex pattern.
* Some constants, such as `default_log_dirs` and `access_log_names` are now properties of the `ApachePlugin` rather than hardcoded lists contained within methods. This way, the subclass can easily extend these properties.

The refactor also takes the initial review suggestions into account. One additional thing that has been changed compared to [56f8821](https://github.com/fox-it/dissect.target/pull/384/commits/56f882145248ca16fa49ab0f0f20deeb268f2cd8) is a small rewrite of how Apache configurations are parsed for the presence of `CustomLog` directives. To also support `ErrorLog` directives this code had to be supplemented anyway, and the parsing would previously use a mix-and-match of `split`, `strip`, `replace` and indexing:

`log_path = line.split("CustomLog")[1].strip().split(" ")[0].replace('"', "")`

This has been altered to a regex, to be more in line with the rest of the plugin structure. 
 
I realize this is a substantial refactor of the `ApachePlugin`. @JSCU-CNI feel free to leave your thoughts on the proposed changes, as you originally wrote the plugin.

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
-        # Check default Apache configs for their CustomLog directive
-        default_config_paths = [
-            "/etc/apache2/apache2.conf",
-            "/usr/local/etc/apache22/httpd.conf",
-            "/etc/httpd/conf/httpd.conf",
-        ]
-
-        for config in default_config_paths:
+        for log_dir, log_name in itertools.product(self.DEFAULT_LOG_DIRS, self.ACCESS_LOG_NAMES):
+            access_log_paths.update(self.target.fs.path(log_dir).glob(log_name + "*"))
```
Schamper, line 204:
```suggestion
            access_log_paths.update(self.target.fs.path(log_dir).glob(f"{log_name}*"))
```
This reads a bit nicer IMO.

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
+
+        for log_dir, log_name in itertools.product(self.DEFAULT_LOG_DIRS, self.ERROR_LOG_NAMES):
+            error_log_paths.update(self.target.fs.path(log_dir).glob(log_name + "*"))
```
Schamper, line 207:
```suggestion
            error_log_paths.update(self.target.fs.path(log_dir).glob(f"{log_name}*"))
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
+                        pattern_to_use = RE_CONFIG_CUSTOM_LOG_DIRECTIVE
+
+                    match = pattern_to_use.match(line)
+                    if not match:
                         self.target.log.warning("Unexpected Apache log configuration: %s (%s)", line, path)
+                        continue
+
+                    directive = match.groupdict()
+                    custom_log = self.target.fs.path(directive["location"])
+                    set_to_update.update(path for path in custom_log.parent.glob(f"{custom_log.name}*"))
```
Schamper, line 232: And you already do it here too :smile:

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
 
-        return log_paths
+        return sorted(access_log_paths), sorted(error_log_paths)
 
     @plugin.export(record=WebserverAccessLogRecord)
     def access(self) -> Iterator[WebserverAccessLogRecord]:
         """Return contents of Apache access log files in unified WebserverAccessLogRecord format."""
```
Schamper, line 238:
```suggestion
        """Return contents of Apache access log files in unified ``WebserverAccessLogRecord`` format."""
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
+                    _target=self.target,
+                )
+
+            except Exception as e:
+                self.target.log.warning("An error occured parsing Apache log file %s: %s", path, str(e))
+                self.target.log.debug("", exc_info=e)
+
+    @plugin.export(record=WebserverErrorLogRecord)
+    def error(self) -> Iterator[WebserverErrorLogRecord]:
+        """Return contents of Apache error log files in unified WebserverErrorLogRecord format."""
```
Schamper, line 281:
```suggestion
        """Return contents of Apache error log files in unified ``WebserverErrorLogRecord`` format."""
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
-                        )
-                        continue
+        Three default log type examples from Apache (note that the ipv4 could also be ipv6)::
+            combined       = '1.2.3.4 - - [19/Dec/2022:17:25:12 +0100] "GET / HTTP/1.1" 304 247 "-" "Mozilla/5.0
+                            (Windows NT 10.0; Win64; x64); AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0
+                            Safari/537.36"'
+            common         = '1.2.3.4 - - [19/Dec/2022:17:25:40 +0100] "GET / HTTP/1.1" 200 312'
+            vhost_combined = 'example.com:80 1.2.3.4 - - [19/Dec/2022:17:25:40 +0100] "GET / HTTP/1.1" 200 312 "-"
+                            "Mozilla/5.0 (Windows NT 10.0; Win64; x64); AppleWebKit/537.36 (KHTML, like Gecko)
+                            Chrome/108.0.0.0 Safari/537.36"'
```
Schamper, line 347: Did you check how this looks like in Sphinx?

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
+        """
+        splitted_line = line.split()
+        first_part = splitted_line[0]
```
Schamper, line 350:
```suggestion
        parts = line.split()
        first_part = parts[0]
```

dissect/target/plugins/os/unix/bsd/citrix/webserver.py
```diff
@@ -0,0 +1,74 @@
...
+    ),
+)
+
+
+class CitrixWebserverPlugin(ApachePlugin):
+    """Citrix uses Apache with custom access log formats. These are:
+    LogFormat "%{Citrix-ns-orig-srcip}i -> %{Citrix-ns-orig-destip}i %l %u %t [%P] \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\" \"Time: %D microsecs\"" combined_resptime_with_citrix_hdrs
+    LogFormat "%a %l %u %t [%P] \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\" \"Time: %D microsecs\"" combined_resptime
+
+    """  # noqa: E501, W605
```
Schamper, line 58:
```suggestion
    """Apache log parsing plugin for Citrix specific logs.
    
    Citrix uses Apache with custom access log formats. These are::
        LogFormat "%{Citrix-ns-orig-srcip}i -> %{Citrix-ns-orig-destip}i %l %u %t [%P] \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\" \"Time: %D microsecs\"" combined_resptime_with_citrix_hdrs
        LogFormat "%a %l %u %t [%P] \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\" \"Time: %D microsecs\"" combined_resptime
    """  # noqa: E501, W605
```
Would be nice to also verify how this renders in Sphinx.

tests/plugins/apps/webserver/test_apache.py
```diff
@@ -2,56 +2,59 @@
...
     ApachePlugin,
-    LogFormat,
-    infer_log_format,
 )
+from dissect.target.target import Target
 from tests._utils import absolute_path
 
 
-def test_plugins_apps_webservers_apache_infer_log_format_combined():
+def test_plugins_apps_webservers_apache_infer_access_log_format_combined():
```
Schamper, line 16: While you're at it, can you shorten these function names?

tests/plugins/os/unix/bsd/citrix/test_webserver.py
```diff
@@ -0,0 +1,78 @@
...
+from dissect.target.plugins.os.unix.bsd.citrix.webserver import (
+    LOG_FORMAT_CITRIX_NETSCALER_ACCESS_COMBINED_RESPONSE_TIME,
+    LOG_FORMAT_CITRIX_NETSCALER_ACCESS_COMBINED_RESPONSE_TIME_WITH_HEADERS,
+    CitrixWebserverPlugin,
+)
+from dissect.target.target import Target
+from tests._utils import absolute_path
+
+
+def test_citrix_webserver_infer_access_log_citrix_netscaler_combined_response_time():
```
Schamper, line 13: Same here.

dissect/target/plugins/os/unix/bsd/citrix/webserver.py
Schamper, line 1: The current location of this plugin (and the exclusion in `webserver.py` means it won't get run if you run `webserver.access`, `webserver.logs`. Is that desirable? You can now _only_ invoke it with `citrix.access` and `citrix.error`.

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -77,83 +196,165 @@ def get_log_paths(self) -> list[Path]:
...
+                    source=path,
+                    _target=self.target,
+                )
+
+            except Exception as e:
+                self.target.log.warning("An error occured parsing Apache log file %s: %s", path, str(e))
+                self.target.log.debug("", exc_info=e)
+
+    @plugin.export(record=WebserverErrorLogRecord)
+    def error(self) -> Iterator[WebserverErrorLogRecord]:
```
Schamper, line 280: Can you add a `error` namespace function to `webserver` too? If you're feeling daring, you could also upgrade that namespace plugin to a proper [`NamespacePlugin`](https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugin.py#L823).

Schamper (changes requested):

dissect/target/plugins/apps/webserver/apache.py
MaxGroot: Made some small changes so it is properly formatted in Sphinx.

dissect/target/plugins/os/unix/bsd/citrix/webserver.py
MaxGroot: Slightly altered to fit within Sphinx formatting.

dissect/target/plugins/os/unix/bsd/citrix/webserver.py
MaxGroot: Moved to the same directory as other `webserver` plugins.

dissect/target/plugins/apps/webserver/apache.py
MaxGroot: Took the dare, ran into a few bugs with the `NamespacePlugin`, which is why the tests fail until https://github.com/fox-it/dissect.target/pull/463 is merged into main and this branch is rebased.

MaxGroot:
Depends on https://github.com/fox-it/dissect.target/pull/463.

Schamper (approved):

PR #385, opened 2023-09-04
Citrix netscaler OS fixes and add support for netscaler cli history

MaxGroot:
The previous approach returned a virtualfilesystem on detect(). As I understand it, an OS plugin should return the system volume on detect, and bootstrap the filesystem in the create classmethod. Also added a check for /log to distinguish ramdisk from system volume. Moreover, account for usage on *only* the ramdisk by falling back to the ramdisk as system volume if the harddisk is not available.

Lastly, this PR adds support for netscaler-cli to the command history plugin.


dissect/target/plugins/os/unix/bsd/citrix/_os.py
Schamper, line 1: Can you rename `CitrixBsdPlugin` to `CitrixPlugin`?

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -49,22 +50,36 @@ def _parse_netscaler_configs(self) -> None:
...
+    @classmethod
+    def create(cls, target: Target, sysvol: Filesystem) -> UnixPlugin:
         for fs in target.filesystems:
             if fs.exists("/bin/freebsd-version"):
-                newfilesystem.map_fs("/", fs)
-                break
+                # If available, mount the ramdisk first.
+                target.fs.mount("/", fs)
+
+        target.fs.mount("/var", sysvol)
```
Schamper, line 75: Maybe add a comment here that the "disk" filesystem is actually mounted at `/var` and that in a future iteration we can figure out how to "recreate" the ramdisk from a clean state?

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -96,16 +111,18 @@ def ips(self) -> list[str]:
...
-        nstmp_path = "/var/nstmp/"
+        nstmp_path = self.target.fs.path("/var/nstmp/")
 
-        nstmp_user_path = nstmp_path + "{username}"
+        nstmp_contents = []
+        if nstmp_path.exists():
+            nstmp_contents = list(nstmp_path.iterdir())
 
-        for entry in self.target.fs.scandir(nstmp_path):
+        for entry in nstmp_contents:
```
Schamper, line 120: Why not something like:

```python
if nstmp_path.exists():
    for entry in nstmp_path.iterdir():
        ...
```

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -96,16 +111,18 @@ def ips(self) -> list[str]:
...
             if entry.is_dir() and entry.name != "#nsinternal#":
                 nstmp_users.add(entry.name)
         for username in self._config_usernames:
-            nstmp_home = nstmp_user_path.format(username=username)
-            user_home = nstmp_home if self.target.fs.exists(nstmp_home) else None
+            nstmp_home = nstmp_path / username
```
Schamper, line 124:
```suggestion
            nstmp_home = nstmp_path.joinpath(username)
```

Preferably use the non-magic API in the main library code.

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -120,7 +137,7 @@ def users(self) -> Iterator[UnixUserRecord]:
             yield UnixUserRecord(name=username, home=user_home)
 
         for username in nstmp_users:
-            yield UnixUserRecord(name=username, home=nstmp_user_path.format(username=username))
+            yield UnixUserRecord(name=username, home=nstmp_path / username)
```
Schamper, line 140:
```suggestion
            yield UnixUserRecord(name=username, home=nstmp_path.joinpath(username))
```

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -96,16 +111,18 @@ def ips(self) -> list[str]:
     @export(record=UnixUserRecord)
     def users(self) -> Iterator[UnixUserRecord]:
```
Schamper, line 112: Is there no information available regarding uid?

tests/test_plugins_os_unix_history.py
```diff
@@ -112,6 +112,34 @@ def test_commandhistory_zsh_history(target_unix_users, fs_unix):
     assert results[1].source.as_posix() == "/root/.zsh_history"
 
 
+def test_commandhistory_netscaler_cli_history(target_citrix, fs_bsd):
```
Schamper, line 115: Can you add type hints?

tests/test_plugins_os_unix_bsd_citrix.py
```diff
@@ -3,18 +3,17 @@
 from dissect.target.plugins.os.unix.bsd.citrix._os import CitrixBsdPlugin
 
 
-def test_unix_bsd_citrix_os(target_citrix):
+def test_unix_bsd_citrix_os(target_citrix, fs_bsd):
```
Schamper, line 6: Can you add type hints?

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/citrix/_os.py
MaxGroot: When you add a user to Citrix Netscaler, it doesn't really add a 'user' to the OS (like adding an entry in `/etc/shadow`). Instead, an SSH key is added that can be used to login, and when users then enter `shell` they drop into a shell as `root`. So for 'human' users, there is no uid. We could do a `yield from super().users` at the end of this function to try to get results out of the `/etc/shadow` file though. I opted against it for now as their home folder is then interpreted as `/`, and if you have a `.netscaler_cli` file in the root of the filesystem, you get the same bash history for a bunch of bogus users.

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+)
+
+
+class CitrixCommandHistoryPlugin(CommandHistoryPlugin):
+    COMMAND_HISTORY_ABSOLUTE_PATHS = (("citrix-netscaler-bash", "/var/log/bash.log*"),)
+
+    def _find_history_files(self) -> List[Tuple[str, TargetPath, Any]]:
+        """
+        Find history files on the target that this plugin can parse.
+        """
```
Schamper, line 53:
```suggestion
        """Find history files on the target that this plugin can parse."""
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+        if username is None:
+            return None
+        user_details = self.target.user_details.find(username=username)
+        return user_details.user if user_details else None
+
+    @export(record=CommandHistoryRecord)
+    def commandhistory(self) -> Iterator[CommandHistoryRecord]:
+        """
+        Return shell history for all users.
```
Schamper, line 79:
```suggestion
        """Return shell history for all users.
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+                yield from self.parse_netscaler_bash_history(history_path)
+
+    @internal
+    def parse_netscaler_bash_history(
+        self,
+        path: TargetPath,
+    ) -> Iterator[CommandHistoryRecord]:
+        """
+        Parse bash.log* contents.
+        """
```
Schamper, line 97:
```suggestion
        """Parse bash.log* contents."""
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @internal
+    def parse_netscaler_cli_history(
+        self,
+        history_file: TargetPath,
+        user: UnixUserRecord,
+    ) -> Iterator[CommandHistoryRecord]:
+        """
+        Parses the history file of the Citrix Netscaler CLI.
+        The only difference compared to generic bash history files is that the first line will start with
```
Schamper, line 127:
```suggestion
        """Parses the history file of the Citrix Netscaler CLI.

        The only difference compared to generic bash history files is that the first line will start with
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Also utilize the _find_history_files function of the parent class.
+        history_files.extend(super()._find_history_files())
+        return history_files
+
+    @cache
+    def _find_user_by_name(self, username: str) -> Union[UnixUserRecord, None]:
+        """
+        Cached function to return the matching UnixUserRecord for a given username.
+        """
```
Schamper, line 70:
```suggestion
        """Cached function to return the matching UnixUserRecord for a given username."""
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+        Find history files on the target that this plugin can parse.
+        """
+        history_files = []
+        for shell, history_absolute_path_glob in self.COMMAND_HISTORY_ABSOLUTE_PATHS:
+            for path in self.target.fs.glob(history_absolute_path_glob):
+                target_path = self.target.fs.path(path)
+                history_files.append((shell, target_path, None))
+
+        self.COMMAND_HISTORY_RELATIVE_PATHS += (("citrix-netscaler-cli", ".nscli_history"),)
```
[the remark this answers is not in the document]
Schamper: I assume because @MaxGroot wants to extend the list from the parent class. You could probably also write this as 

```
class CitrixCommandHistoryPlugin(CommandHistoryPlugin):
    ...
    COMMAND_HISTORY_RELATIVE_PATHS = CommandHistoryPlugin.COMMAND_HISTORY_RELATIVE_PATHS + (("citrix-netscaler-cli", ".nscli_history"),)
```

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Also utilize the _find_history_files function of the parent class.
```
Schamper, line 62:
```suggestion
        # Also utilize the _find_history_files function of the parent class
```

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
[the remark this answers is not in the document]
MaxGroot: Implemented using the suggestion from @Schamper .

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _find_history_files(self) -> List[Tuple[str, TargetPath, Any]]:
+        """
+        Find history files on the target that this plugin can parse.
+        """
+        history_files = []
+        for shell, history_absolute_path_glob in self.COMMAND_HISTORY_ABSOLUTE_PATHS:
+            for path in self.target.fs.glob(history_absolute_path_glob):
+                target_path = self.target.fs.path(path)
+                history_files.append((shell, target_path, None))
```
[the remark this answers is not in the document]
MaxGroot: As the `history_absolute_path_glob` is an absolute path, `path().glob()` raises an exception as it expects a relative glob rather than an absolute one. We could of course alter the values of the `COMMAND_HISTORY_ABSOLUTE_PATHS` from absolute to relative paths (e.g. `/var/log/bash.log*` to `var/log/bash.log*`) but in my opinion that is too confusing. Thus, I've kept it as-is for now. Let me know if you have an alternative solution in mind.

dissect/target/plugins/os/unix/bsd/citrix/_os.py
MaxGroot: I ended up iterating through `super().users()` and setting the `.home` to None only if the `.home` would otherwise be `/`.

dissect/target/plugins/os/unix/bsd/citrix/commandhistory.py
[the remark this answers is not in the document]
Schamper: Elsewhere we use `self.target.fs.path().glob(path.lstrip("/"))`

tests/test_plugins_os_unix_bsd_citrix.py
Schamper, line 1: Can you rebase and move these tests?

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -49,22 +49,42 @@ def _parse_netscaler_configs(self) -> None:
...
-                is_citrix = True
-        if is_citrix:
-            return newfilesystem
-        return None
+                target.fs.mount("/flash", fs)
+
+        return cls(target)
 
     @export(property=True)
     def hostname(self) -> Optional[str]:
```
Schamper, line 90: Maybe fall back to `return self._hostname or super().hostname` in case no hostname is explicitly configured?

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -117,10 +138,19 @@ def users(self) -> Iterator[UnixUserRecord]:
...
-            yield UnixUserRecord(name=username, home=nstmp_user_path.format(username=username))
+            seen.add(username)
+            yield UnixUserRecord(name=username, home=nstmp_path.joinpath(username))
+
+        # Yield users from /etc/shadow if we have not seem them in previous loops
+        for user in super().users():
+            if user.name in seen:
+                continue
+            user_home = user.home if user.home != "/" else None
+            yield UnixUserRecord(name=user.name, home=user_home)
```
Schamper, line 153: You lose all the other fields from the original record by doing this (`gecos`, `shell`, etc).

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -117,10 +138,19 @@ def users(self) -> Iterator[UnixUserRecord]:
...
 
+            seen.add(username)
             yield UnixUserRecord(name=username, home=user_home)
 
         for username in nstmp_users:
-            yield UnixUserRecord(name=username, home=nstmp_user_path.format(username=username))
+            seen.add(username)
+            yield UnixUserRecord(name=username, home=nstmp_path.joinpath(username))
+
+        # Yield users from /etc/shadow if we have not seem them in previous loops
```
Schamper, line 148:
```suggestion
        # Yield users from /etc/passwd if we have not seem them in previous loops
```

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -117,10 +138,19 @@ def users(self) -> Iterator[UnixUserRecord]:
...
+        for user in super().users():
+            if user.name in seen:
```
Schamper, line 150: You don't add these to the seen.

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -96,16 +116,17 @@ def ips(self) -> list[str]:
     @export(record=UnixUserRecord)
     def users(self) -> Iterator[UnixUserRecord]:
         nstmp_users = set()
-        nstmp_path = "/var/nstmp/"
+        seen = set()
```
Schamper, line 119: Could you "uniq" on `(name, home, shell)`? Shell will mostly be None for the Citrix specific users, but it's good practice.

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/citrix/_os.py
MaxGroot: As this is the final loop of the method, and we are only receiving users from the `super().users()` function, I don't think we have to add the users to the seen.

dissect/target/plugins/os/unix/bsd/citrix/_os.py
MaxGroot: Done

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -88,7 +88,7 @@ def create(cls, target: Target, sysvol: Filesystem) -> CitrixPlugin:
 
     @export(property=True)
     def hostname(self) -> Optional[str]:
-        return self._hostname
+        return self._hostname if self._hostname else super().hostname
```
Schamper, line 91:
```suggestion
        return self._hostname or super().hostname
```

dissect/target/plugins/os/unix/bsd/citrix/_os.py
```diff
@@ -135,22 +139,28 @@ def users(self) -> Iterator[UnixUserRecord]:
...
             if username == "root" and self.target.fs.exists("/root"):
                 # If we got here, 'root' is present both in /var/nstmp and in /root. In such cases, we yield
-                # the 'root' user as having '/root' as a home, not in /var/nstmp.
-                user_home = "/root"
+                # the 'root' user as having '/root' as a home, not in /var/nstmp, as there is no 'nscli_history'
+                # for the root user in /var/nstmp.
+                user_home = self.target.fs.path("/root")
 
-            seen.add(username)
+            seen.add((username, user_home.as_posix() if user_home else None, None))
```
Schamper, line 146: Why not just `user_home` as tuple member?

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/citrix/_os.py
MaxGroot: `posix_path` instances are unhashable, hence the usage of the string representation

tests/plugins/os/unix/bsd/citrix/test__os.py
```diff
@@ -1,20 +1,31 @@
...
 
-from dissect.target.plugins.os.unix.bsd.citrix._os import CitrixBsdPlugin
+from dissect.target.filesystem import VirtualFilesystem
+from dissect.target.plugins.os.unix.bsd.citrix._os import CitrixPlugin
+from dissect.target.target import Target
 
 
-def test_unix_bsd_citrix_os(target_citrix):
-    target_citrix.add_plugin(CitrixBsdPlugin)
+def test_unix_bsd_citrix_os(target_citrix: Target, fs_bsd: VirtualFilesystem):
```
Schamper, line 9:
```suggestion
def test_citrix_os(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
```

tests/plugins/os/unix/bsd/citrix/test_commandhistory.py
```diff
@@ -0,0 +1,93 @@
...
+from io import BytesIO
+
+from dissect.target.filesystem import VirtualFilesystem
+from dissect.target.plugins.os.unix.bsd.citrix.commandhistory import (
+    CitrixCommandHistoryPlugin,
+)
+from dissect.target.target import Target
+
+
+def test_netscaler_bash_history_absolute_path_glob(target_citrix: Target, fs_bsd: VirtualFilesystem):
```
Schamper, line 13:
```suggestion
def test_netscaler_bash_history_absolute_path_glob(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
```

tests/plugins/os/unix/bsd/citrix/test_commandhistory.py
```diff
@@ -0,0 +1,93 @@
...
+    assert results[1].shell == "citrix-netscaler-bash"
+    assert results[1].source.as_posix() == "/var/log/bash.log"
+
+    assert results[0].ts == datetime.datetime(1970, 8, 10, 11, 57, 39, tzinfo=datetime.timezone.utc)
+    assert results[0].command == 'debug "hello world"'
+    assert results[0].shell == "citrix-netscaler-bash"
+    assert results[0].source.as_posix() == "/var/log/bash.log"
+
+
+def test_netscaler_commandhistory_decompress(target_citrix: Target, fs_bsd: VirtualFilesystem):
```
Schamper, line 41:
```suggestion
def test_netscaler_commandhistory_decompress(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
```

tests/plugins/os/unix/bsd/citrix/test_commandhistory.py
```diff
@@ -0,0 +1,93 @@
...
+    assert results[1].shell == "citrix-netscaler-bash"
+    assert results[1].source.as_posix() == "/var/log/bash.log.0.gz"
+
+    assert results[0].ts == datetime.datetime(1970, 8, 10, 11, 57, 39, tzinfo=datetime.timezone.utc)
+    assert results[0].command == 'debug "hello world"'
+    assert results[0].shell == "citrix-netscaler-bash"
+    assert results[0].source.as_posix() == "/var/log/bash.log.0.gz"
+
+
+def test_netscaler_cli_history(target_citrix: Target, fs_bsd: VirtualFilesystem):
```
Schamper, line 68:
```suggestion
def test_netscaler_cli_history(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
```

Schamper (changes requested):

MaxGroot:
@Politie-SOC Thanks for posting. As you can see at https://github.com/fox-it/dissect.target/pull/385/files#diff-7e4df35dcb23e758fc9bb1932a06fed16bfbf6303e9c514cca69cfaecef7562fR84, a filesystem is currently inspected for the presence of a `/nscconfig` folder and a `/boot` folder, and if both are presented, said filesystem will be mounted at `/flash`. 

In your output, I see the disk image that should contain the `/flash` filesystem has itself a folder in it named `/flash`? I myself have not encountered a situation like that before, as I have only created a simultaneous image of `/dev/sda0` (containing both the `flash` and `var` mounts).

Could you share how you created an image of the flash partition, and why you went that route? It could be that I'm missing something from your posted output, the markdown is a bit iffy. We could add a check for a `/flash` folder and mount its contents at `/flash`, but I'd like to understand in what sort of situations this is necessary. Thanks!

Politie-SOC:
So what we did in this case was create a dd of this:

```
/dev/ar0s1a on /flash (ufs, local)
```

As mentioned here https://github.com/fox-it/dissect.target/issues/484 when we dd /dev/ar0 instead, we also dd the MBR (according to @Schamper ) and Dissect needs that to parse it properly.

The output above shows /dev/ar0s1a instead of /dev/ad0s1a but I'm assuming that shouldn't matter.

MaxGroot:
All right, seems that we don't have to account for a situation where there is a `/flash` folder within the flash filesystem itself then, and we can leave the proposed `detect()` and `create()` as is.

Schamper (approved):

Issue #356, opened 2023-08-13
yara-python doesn't work in Python3.11 and breaks tox testing


OlafHaalstra:
`yara-python` cannot be installed normally under Python 3.11. Whilst this is out of scope for `dissect` this currently breaks the testing script.

Commenting out `"yara-python"` in `pyproject.toml` allows to test it without yara.

When running `tox` for testing I get the following error:
```
Building wheels for collected packages: yara-python
  Building wheel for yara-python (setup.py) ... error
  Running setup.py clean for yara-python
Failed to build yara-python
  error: subprocess-exited-with-error

  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [227 lines of output]

...

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for yara-python
ERROR: Could not build wheels for yara-python, which is required to install pyproject.toml-based projects
```

Which is very similar to the error you get when running `pip install yara-python`:
```
Building wheels for collected packages: yara-python
  Building wheel for yara-python (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for yara-python (pyproject.toml) did not run successfully.
  │ exit code: 1

...

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> yara-python
```
When running `pip install -U git+https://github.com/VirusTotal/yara-python`, installing directly from source. It installs just fine, however still cannot be used with the `tox` command after installing.

Schamper:
Yes, `yara-python` is horrible. I've been pushing to remove it from the dependencies and making it completely optional, perhaps we can finally do so now.

PR #387, opened 2023-09-06
Replace yara-python with yara-python-wheel for prebuilt wheels

Schamper:
Use pre-built wheels for yara-python to speed up the CI pipeline. Coming from https://github.com/Schamper/yara-python-wheel.

Fixes https://github.com/fox-it/dissect.target/issues/356


tox.ini
```diff
@@ -11,11 +11,12 @@ minversion = 4.2.4
 requires = virtualenv>=20.16.6
 
 [testenv]
-extras = yara
+extras = full
```
pyrco, line 14: Wouldn't it be nicer to change the dependency in pyproject.toml? I guess most people who want to use the yara functionality don't want the hassle of compiling the yara-python package?

pyrco (approved):

PR #390, opened 2023-09-08
Set UTF-8 as default encoding for open_decompress

Poeloe:


dissect/target/helpers/fsutil.py
```diff
@@ -1011,6 +1011,9 @@ def open_decompress(
     if "b" in mode:
         # Reset the default errors mode in case of a binary stream
         errors = None
+    else:
+        # Set default encoding as UTF-8 if not binary mode
+        encoding = encoding or "UTF-8"
```
Schamper, line 1016: Perhaps better to set this the same way as the errors? I.e. set it as default argument in the function signature, and reset to `None` in case of a binary stream.

Schamper (changes requested):

dissect/target/helpers/fsutil.py
Poeloe: I did that, but then the raised error assertion at the end fails. I supposed that is necessary, since it is tested.

Schamper: It doesn't strictly have to, it was just for completeness of the test. The current code for the default errors mode already prevents another default Python error from popping up:

```
In [3]: Path("/etc/hostname").open("rb", errors="backslashreplace")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 Path("/etc/hostname").open("rb", errors="backslashreplace")

File /usr/lib64/python3.11/pathlib.py:1044, in Path.open(self, mode, buffering, encoding, errors, newline)
   1042 if "b" not in mode:
   1043     encoding = io.text_encoding(encoding)
-> 1044 return io.open(self, mode, buffering, encoding, errors, newline)

ValueError: binary mode doesn't take an errors argument
```

In `open_decompress`, this will silently be ignored since the errors mode will automatically be set to `None`. I'm personally fine with this being the behavior for the `encoding` argument as well.

Schamper: Yup, basically this:

```diff
diff --git a/dissect/target/helpers/fsutil.py b/dissect/target/helpers/fsutil.py
index 56bc45f..3bc06eb 100644
--- a/dissect/target/helpers/fsutil.py
+++ b/dissect/target/helpers/fsutil.py
@@ -982,7 +982,7 @@ def resolve_link(
 def open_decompress(
     path: TargetPath,
     mode: str = "rb",
-    encoding: Optional[str] = None,
+    encoding: Optional[str] = "UTF-8",
     errors: Optional[str] = "backslashreplace",
     newline: Optional[str] = None,
 ) -> Union[BinaryIO, TextIO]:
@@ -1009,7 +1009,8 @@ def open_decompress(
     file.seek(0)
 
     if "b" in mode:
-        # Reset the default errors mode in case of a binary stream
+        # Reset the default encoding and errors mode in case of a binary stream
+        encoding = None
         errors = None
 
     if magic[:2] == b"\x1f\x8b":
```

Schamper (approved):

Schamper (changes requested):
Actually, can you update the docstring to say we default to UTF-8?

Schamper (approved):

PR #394, opened 2023-09-12
Add a plugin to parse notifications from Windows appdb.dat

pyrco:


dissect/target/plugins/os/windows/notifications.py
```diff
@@ -63,11 +198,149 @@ def __init__(self, target):
...
+                if ts := tile.TS:
+                    tile_ts = wintimestamp(ts)
+                tile_expiry_ts = None
+                if ts := tile.TS_Expiry:
+                    tile_expiry_ts = wintimestamp(ts)
+                name = tile.Name.strip("\x00")
+
+                xml_size = tile.ContentLength
+                tile_xml = chunk.TileXml[tile_no][:xml_size]
+                tile_xml = tile_xml.decode("utf-8", errors="replace")
```
cecinestpasunepipe, line 261: I thought we agreed on using errors="surrogateescape" ? If there is a special reason not do so, maybe we could add a comment?

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -63,11 +198,149 @@ def __init__(self, target):
...
+            toast = chunk.Toasts[toast_no]
+
+            if toast.UniqId:
+                toast_ts = None
+                if ts := toast.TS:
+                    toast_ts = wintimestamp(ts)
+                toast_expiry_ts = None
+                if ts := toast.TS_Expiry:
+                    toast_expiry_ts = wintimestamp(ts)
+                name1 = toast.Name1.strip("\x00")
```
cecinestpasunepipe, line 292: Just a general suggestion... I notice that thi NUL-byte handling is becoming a very common thing (41 matches in target alone!), maybe we should add a convenience function in cstruct for this? I.e. cstr(...) ? because apparently we want to extract the c-string.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -63,11 +198,149 @@ def __init__(self, target):
                 self.wpndb_files.append((user_details.user, wpndb_file))
 
             if appdb_file.exists():
-                self.appdb_files.append((user_details.user, appdb_file))
+                with appdb_file.open(mode="rb") as fp:
+                    chunk = c_appdb.Chunk(fp)
+                    if chunk.Header.Magic == APPDB_MAGIC and chunk.Header.Version == 3:
```
cecinestpasunepipe, line 203: Maybe add a warning for unsupported versions and list the specified version number?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/windows/notifications.py
pyrco: There is such a function in cstruct for all relevant types: `_read_0()` it seems though that it is not exposed.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -1,15 +1,150 @@
...
+        ("datetime", "badge_ts"),
+        ("string", "badge_data"),
+    ],
+)
+
+AppDBTileRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+    "windows/notification/appdb_tile",
+    [
+        ("varint", "id"),
+        ("datetime", "ts"),
```
Schamper, line 125: Does it make sense to swap these so that the `ts` is first? Good for sorting.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -1,15 +1,150 @@
...
+        ("string", "name"),
+        ("string", "content"),
+    ],
+)
+
+AppDBToastRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+    "windows/notification/appdb_toast",
+    [
+        ("varint", "id"),
+        ("datetime", "ts"),
```
Schamper, line 138: Does it make sense to swap these so that the `ts` is first? Good for sorting.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -1,15 +1,150 @@
...
+c_appdb = cstruct(endian="<")
+c_appdb.load(appdb_def)
+
+APPDB_MAGIC = b"DNPW"
+NUM_APPDB_CHUNKS = 256
+
+AppDBChunkRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+    "windows/notification/appdb_chunk",
+    [
+        ("datetime", "last_notification_date"),
```
Schamper, line 110: Is this the field that would make sense to be sorted on to get a chronological ordering of records? Some comments explaining what the various timestamps are would be nice.

Schamper (changes requested):

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/windows/notifications.py
pyrco: Except for ts and ts_expiry the other timestamps are unobserved or unclear. This one doesn't even seem to be the actual last notification date as there are tiles and toasts with more recent dates.
Last notification date is the name the original source gave to this field, so I kept it. I can change it to a generic `ts` if you'd like.

pyrco: On top of that this timestamp is identical for all chunk records as it only appears once in the file (in the first chunk header).

dissect/target/plugins/os/windows/notifications.py
pyrco: These are part of a grouped record so I don't know if sorting makes sense here.

dissect/target/plugins/os/windows/notifications.py
Schamper: Would `chunk_ts` be more appropriate?

(These kinds of summarised comments would help in the source by the way 😄)

pyrco: I added some comments to the cstruct definition.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -1,15 +1,150 @@
...
+  TileDescriptor  Tiles[5];          // start @ 0x958
+  DataXML         TileXml[5];
+
+  // For the in use chunks, 0x14 ToastDiscriptors have an Index, but there
+  // is space for more. Maybe this is used in case of deleted entries?
+  ToastDescriptor Toasts[0x14];      // start @ 0x6e98
+  char            Padding1[0x1e00];  // start @ 0x7618
+  DataXML         ToastXml[0x14];    // start @ 0x9418
+  char            Padding2[0x13f8];  // start @ 0x22418
+} Chunk;                             // size: 0x23810
```
Schamper, line 98:
```suggestion
typedef struct {
    char    Magic[4];                   // Always b"DNPW"
    DWORD   Version;                    // Versions 1 (win8) and 3 (win10) seem to exist
    QWORD   Timestamp;                  // Some sources claim this is LastNotificationDate, but that seems incorrect
    DWORD   NextNotificationId;
    DWORD   Unknown;
    char    Padding[8];
} ChunkHeader;                          // size: 0x20

typedef struct {
    WORD    InUse;                      // ??
    WORD    NextTileWrite;              // ??
    WORD    NextToastWrite;             // ??
    BYTE    Flags[2];
} ChunkInfo;

typedef struct {
    QWORD   Timestamp1;                 // ??
    QWORD   Timestamp2;                 // Is this time to refresh?
    char    Uri[1024];                  // Is this the correct size?
    char    Padding[0x818 - 0x410];
} PushDescriptor;                       // size: 0x818

typedef struct {
    DWORD   Id;
    DWORD   Zero;                       // ??
    QWORD   Timestamp;                  // ??
    WORD    Unknown;
    WORD    DataLength;
    char    Data[DataLength];
    char    Padding[0x118 - 0x14 - DataLength];
} BadgeContent;                         // size: 0x118

typedef struct {
    DWORD   UniqueId;                   // ??
    DWORD   Zero;
    QWORD   ExpiryTime;                 // The time this tile expires
    QWORD   ArrivalTime;                // The time this tile was set
    BYTE    Type;                       // ??
    BYTE    Index;
    WORD    ContentLength;
    wchar_t Name[18];
} TileDescriptor;                       // size: 0x40

typedef struct {
    DWORD   UniqueId;                   // ??
    DWORD   Zero;
    QWORD   ExpiryTime;                 // The time this toast expires
    QWORD   ArrivalTime;                // The time this toast was set
    BYTE    Type;                       // ??
    BYTE    Index;
    WORD    ContentLength;
    wchar_t Name1[17];
    wchar_t Name2[17];
} ToastDescriptor;                      // size: 0x60

typedef struct {
    char    Content[0x1400];
} DataXML;                              // size: 0x1400

typedef struct {
    ChunkHeader     Header;             // Only populated for first chunk, else zeroed
    ChunkInfo       Info;
    PushDescriptor  Push;
    BadgeContent    BadgeXml;
    TileDescriptor  Tiles[5];           // start @ 0x958
    DataXML         TileXml[5];
    // For the in use chunks, 0x14 ToastDiscriptors have an Index, but there
    // is space for more. Maybe this is used in case of deleted entries?
    ToastDescriptor Toasts[0x14];       // start @ 0x6e98
    char            Padding1[0x1e00];   // start @ 0x7618
    DataXML         ToastXml[0x14];     // start @ 0x9418
    char            Padding2[0x13f8];   // start @ 0x22418
} Chunk;                                // size: 0x23810
```
Nitpicky alignment.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -55,19 +192,172 @@ def __init__(self, target):
...
     def check_compatible(self) -> None:
         if not self.wpndb_files and not self.appdb_files:
-            raise UnsupportedPluginError("No wpndatabase.db or appdb.dat files found")
+            raise UnsupportedPluginError("No or incompatible wpndatabase.db or appdb.dat files found")
+
+    def _get_appdb_chunk_record(
+        self,
+        chunk: c_appdb.Chunk,
+        timestamp: Optional[datetime.datetime],
+        user: RecordDescriptor,
```
Schamper, line 231:
```suggestion
        user: WindowsUserRecord,
```
Perhaps we should fix the type hints in the `user_details` plugin.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -55,19 +192,172 @@ def __init__(self, target):
...
+            push_ts2=push_timestamp2,
+            push_uri=push_uri,
+            badge_id=chunk.BadgeXml.Id,
+            badge_ts=badge_ts,
+            badge_data=badge_data,
+            _target=self.target,
+            _user=user,
+        )
+
+    def _get_appdb_tile_records(self, chunk: c_appdb.Chunk, user: RecordDescriptor) -> list[AppDBTileRecord]:
```
Schamper, line 260:
```suggestion
    def _get_appdb_tile_records(self, chunk: c_appdb.Chunk, user: WindowsUserRecord) -> list[AppDBTileRecord]:
```

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -55,19 +192,172 @@ def __init__(self, target):
...
+                    name2=name2,
+                    content=toast_xml,
+                    _target=self.target,
+                    _user=user,
+                )
+
+                toast_records.append(toast_record)
+        return toast_records
+
+    @export(record=GroupedRecord)
```
Schamper, line 332:
```suggestion
    @export(record=[AppDBChunkRecord, AppDBTileRecord, AppDBToastRecord])
```
Might be better, so that we know which records are exported.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -55,19 +192,172 @@ def __init__(self, target):
...
+                            "please consider providing us with a sample.",
+                            chunk.Info.InUse,
+                        )
+                        continue
+
+                    chunk_record = self._get_appdb_chunk_record(chunk, timestamp, user)
+                    tile_records = self._get_appdb_tile_records(chunk, user)
+                    toast_records = self._get_appdb_toast_records(chunk, user)
+
+                    yield GroupedRecord(
```
Schamper, line 357: I haven't checked myself so perhaps I should :smile: but is a GroupedRecord the best method? Have you considered individual records?

Schamper (changes requested):

dissect/target/plugins/os/windows/notifications.py
pyrco: I'm not sure if we can say that the tiles and toasts are independent. It is also unclear that if there are tiles and toasts, the first tile and first toast always go together etc.

We could emit the chunk, tiles and toasts individually and add the chunk id to the latter two, but then you'd have to tie everything back together again at the receiving end.

dissect/target/filesystem.py
```diff
@@ -960,6 +961,21 @@ def lattr(self) -> dict[str, bytes]:
         return fsutil.fs_attrs(self.entry, follow_symlinks=False)
 
 
+class MappedCompressedFile(MappedFile):
+    """Virtual file backed by a compressed file on the host machine."""
+
+    _compressors = {"gzip": gzip}
+
+    def __init__(self, fs: Filesystem, path: str, entry: Any, algo: str = "gzip") -> None:
```
Schamper, line 969:
```suggestion
    def __init__(self, fs: Filesystem, path: str, entry: Any, algo: str = "gzip"):
```

Schamper (changes requested):

dissect/target/plugins/os/windows/notifications.py
Schamper: Can you also still apply this?

pyrco: I thought I did, or did I miss something (modified it by hand)?

tests/test_plugins_os_windows_notifications.py
```diff
@@ -1,13 +1,87 @@
...
+    NOTIFICATIONS_DIR,
+    NotificationsPlugin,
+)
 
 from ._utils import absolute_path
 
+USER_DIR = "Users\\John"
+
+
+def test_notifications_appdb(target_win_users, fs_win):
```
cecinestpasunepipe, line 14: types?

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/windows/notifications.py
Schamper: I also indented the structures by 4 spaces, changed the location of the brackets (style choice), and the alignment of the final `//size:` comment.

dissect/target/plugins/os/windows/notifications.py
pyrco: I added a commit to normalize the records: separate push,badge, tile and toast records. This also allows the timestamps to come first. I think this makes it at least easier to answer the question what was shown to a user when.

dissect/target/filesystem.py
```diff
@@ -1115,13 +1131,18 @@ def map_dir(self, vfspath: str, realpath: str) -> None:
                 real_file_path = os.path.join(root, file_)
                 directory.add(file_, MappedFile(self, vfs_file_path, real_file_path))
 
-    def map_file(self, vfspath: str, realpath: str) -> None:
+    def map_file(self, vfspath: str, realpath: str, compression: str = "") -> None:
```
Schamper, line 1134: Very nitpicky once again, but would a default of `compression = None` not make more sense?

Schamper (changes requested):

dissect/target/filesystem.py
pyrco: Hmm, that should either read "gzip"  or not be a keyword arg. With `None` or `""` the default constructor will always raise a `ValueError`, which is not correct API behaviour.

pyrco: Ah wait, this is `map_file()`, nvm, I was thinking about the `__init__()` of `MappedCompressedFile`. None would be nicer here as a signaling value.

dissect/target/plugins/os/windows/notifications.py
```diff
@@ -1,15 +1,160 @@
...
+    WORD     NextToastWriteIndex;      // ??
+    BYTE     Flags[2];
+} ChunkInfo;
+
+typedef struct {
+    QWORD    Timestamp1;               // ??
+    QWORD    Timestamp2;               // Is this time to refresh?
+    char     Uri[1024];                // Is this the correct size?
+    char     Padding[0x818 - 0x410];
+} PushDescriptor;                    // size: 0x818
```
Schamper, line 38:
```suggestion
} PushDescriptor;                      // size: 0x818
```
I'm so sorry :smile:

Schamper (changes requested):

dissect/target/plugins/os/windows/notifications.py
pyrco: gah!

Schamper (approved):

PR #241, opened 2023-04-25
Add plugins for volatile Linux artefacts

Horofic:
(DIS-1443)

PR #395, opened 2023-09-13
Fix logging of incompatible plugins

Schamper:
Blocks #241.

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -955,6 +955,7 @@ class InternalPlugin(Plugin):
 @dataclass(frozen=True, eq=True)
 class PluginFunction:
     name: str
+    path: str
     output_type: str
     class_object: type[Plugin]
     method_name: str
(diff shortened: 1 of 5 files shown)


dissect/target/plugin.py
```diff
@@ -1074,7 +1075,8 @@ def add_to_result(func: PluginFunction) -> None:
                 matches = True
                 add_to_result(
                     PluginFunction(
-                        name=index_name,
+                        name=f"{func['namespace']}.{method_name}" if func["namespace"] else method_name,
```
cecinestpasunepipe, line 1078: This might cause unwanted behaviour in query.py, because the names are now no longer unique unlike the paths?
```

if args.targets:
            for target in args.targets:
               ....
                for func in funcs:
                    collected_plugins[func.name] = func.plugin_desc
        else:
           ....
            for func in funcs:
                collected_plugins[func.name] = func.plugin_desc
```

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #397, opened 2023-09-15
Support duplicate volume names in target-mount

### LINKED ISSUE #361: Target mount doesn't expose all volumes
using the latest release (dissect.target 3.11.1), mounting a .E01 disk image with multiple partitions yields the following results when using:

target-mount <source> /mnt/tm
Screenshot 2023-08-17 at 08 42 51

As shown in the last fdisk -l command, not all partitions are exposed under /mnt/tm/filesystems and /mnt/tm/volumes.

According to @Schamper this is due to the following:
https://github.com/fox-it/dissect.target/blob/main/dissect/target/tools/mount.py#L55
and
https://github.com/fox-it/dissect.target/blob/main/dissect/target/tools/mount.py#L59

Schamper:
Fixes #361 

Duplicate names will have a `_{n}` appended to the volume name. Duplicate volume names will be consistent between the `volumes/` and `filesystems/` listings.


dissect/target/tools/mount.py
```diff
@@ -51,12 +51,20 @@ def main():
...
         vfs.map_file_fh(fname, d)
 
+    vnames = {}
     for i, v in enumerate(t.volumes):
-        fname = f"volumes/{v.name or f'volume_{i}'}"
-        vfs.map_file_fh(fname, v)
+        basename = f"{v.name or f'volume_{i}'}"
+        fname = basename
+
+        if fname in vnames:
```
cecinestpasunepipe, line 59: In theory this might fail with volumes like: 

```
- a
- a_1
- a

```
Because the 2nd a will try to claim a_1 without checking for that volume name right? 

Simple test code:
```

vnames = {}

class Vol:
        def __init__(self,name):
                self.name = name

volumes = [Vol("la"),Vol("la_1"),Vol("la")]
for i, v in enumerate(volumes):
        basename = f"{v.name}"
        fname = basename
        if fname in vnames:
                fname = f"{basename}_{len(vnames[basename])}"

        vnames[v] = fname
        vnames.setdefault(basename, []).append(v)
        print(fname)
```

gives

```
la
la_1
la_1
```

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #398, opened 2023-09-15
Allow DefaultOS to access all plugins

cecinestpasunepipe:
This allows for instance the log/file loader to access plugins even though the target does not have the necessary OS.

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -19,6 +19,7 @@ from typing import TYPE_CHECKING, Any, Callable, Iterator, Optional, Type
 
 from flow.record import Record, RecordDescriptor
 
+import dissect.target.plugins.general as general
 from dissect.target.exceptions import PluginError, UnsupportedPluginError
 from dissect.target.helpers import cache
 from dissect.target.helpers.record import EmptyRecord
@@ -527,7 +528,7 @@ def arg(*args, **kwargs) -> Callable:
     return decorator
 
 
-def plugins(osfilter: str = None) -> Iterator[PluginDescriptor]:
+def plugins(osfilter: Type[OSPlugin] = None) -> Iterator[PluginDescriptor]:
     """Retrieve all plugin descriptors.
 
     Args:
@@ -553,7 +554,7 @@ def plugins(osfilter: str = None) -> Iterator[PluginDescriptor]:
 
     if (
         osfilter
-        and not hasattr(osfilter, "__is_default__")
+        and not isinstance(osfilter, general.default.DefaultPlugin)
         and isinstance(osfilter, type)
         and issubclass(osfilter, OSPlugin)
         and osfilter.__module__.startswith(MODULE_PATH)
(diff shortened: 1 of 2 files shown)


dissect/target/plugin.py
```diff
@@ -553,6 +553,7 @@ def _walk(osfilter: str = None, root: dict = None) -> Iterator[PluginDescriptor]
 
     if (
         osfilter
+        and not hasattr(osfilter, "__is_default__")
```
pyrco, line 556:
```suggestion
        and not osfilter is DefaultPlugin
```

Maybe also fix the typing of osfilter, as `type[OSPlugin]` it is a OSPlugin subclass.

dissect/target/plugins/general/default.py
```diff
@@ -4,6 +4,7 @@
 
 class DefaultPlugin(OSPlugin):
     __skip__ = True
+    __is_default__ = True
```
pyrco, line 7:
```suggestion
```

pyrco (changes requested):

dissect/target/plugin.py
cecinestpasunepipe: This won't work due to cyclic imports.

pyrco: Ah nasty

pyrco: Would it still be an idea to refactor the typing of `plugins(osfilter: str ...` to `plugins(osfilter: type[OSPlugin] ...`?

cecinestpasunepipe: Yes, that would be better, working on that...

dissect/target/plugin.py
```diff
@@ -527,7 +528,7 @@ def decorator(obj):
     return decorator
 
 
-def plugins(osfilter: str = None) -> Iterator[PluginDescriptor]:
+def plugins(osfilter: Type[OSPlugin] = None) -> Iterator[PluginDescriptor]:
```
Schamper, line 531:
```suggestion
def plugins(osfilter: type[OSPlugin] = None) -> Iterator[PluginDescriptor]:
```
Or is that not allowed?

dissect/target/plugin.py
cecinestpasunepipe: Forgot there are some Python tricks for this. Implemented.

pyrco (dismissed):

dissect/target/plugin.py
```diff
@@ -527,7 +528,7 @@ def decorator(obj):
     return decorator
 
 
-def plugins(osfilter: str = None) -> Iterator[PluginDescriptor]:
+def plugins(osfilter: type[OSPlugin] = None) -> Iterator[PluginDescriptor]:
```
Schamper, line 531:
```suggestion
def plugins(osfilter: Optional[type[OSPlugin]] = None) -> Iterator[PluginDescriptor]:
```

Does this not also break other APIs? This function is used in a number of places if I recall correctly.

Schamper (changes requested):

Schamper (approved):

pyrco (approved):

PR #401, opened 2023-09-18
Get scheduled task display name with filename 

0x49736b:
If Principals/Principal/DisplayName is empty, use the filename as the display_name.

I noticed that the field "Principals/Principal/DisplayName" is not used a lot, resulting in the Display Name being null for most of the records.

Instead, let's use the filename as Display Name (Like in the Task Scheduler GUI)

### DIFF
--- a/dissect/target/plugins/os/windows/task_helpers/tasks_xml.py
+++ b/dissect/target/plugins/os/windows/task_helpers/tasks_xml.py
@@ -100,9 +100,7 @@ class XmlTask:
         self.user_id = self.get_element("Principals/Principal/UserId")
         self.logon_type = self.get_element("Principals/Principal/LogonType")
         self.group_id = self.get_element("Principals/Principal/GroupId")
-        self.display_name = self.get_element("Principals/Principal/DisplayName")
-        if self.display_name == None:
-            self.display_name = str(task_path)[str(task_path).rfind("/")+1:]
+        self.display_name = self.get_element("Principals/Principal/DisplayName") or task_path.name
         self.run_level = self.get_element("Principals/Principal/RunLevel")
         self.process_token_sid_type = self.get_element("Principals/Principal/ProcessTokenSidType")
         self.required_privileges = self.get_element("Principals/Principal/RequiredPrivileges")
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/os/windows/task_helpers/tasks_xml.py
```diff
@@ -101,6 +101,8 @@ def __init__(self, task_element: Element, task_path: TargetPath):
         self.logon_type = self.get_element("Principals/Principal/LogonType")
         self.group_id = self.get_element("Principals/Principal/GroupId")
         self.display_name = self.get_element("Principals/Principal/DisplayName")
+        if self.display_name == None:
+            self.display_name = str(task_path)[str(task_path).rfind("/")+1:]
```
Schamper, line 105:
```suggestion
        self.display_name = self.get_element("Principals/Principal/DisplayName") or task_path.name
```
I believe this achieves the same result.

Schamper (changes requested):

0x49736b:
@DissectBot agree

Schamper:
Could you adjust the unit tests with the newly filled in field?

0x49736b:
> Could you adjust the unit tests with the newly filled in field?

@Schamper , done here https://github.com/fox-it/dissect.target/pull/401/commits/e75a335467853d7643c795f62f5ccc5b307611fe

tests/test_plugins_os_windows_tasks.py
```diff
@@ -40,7 +40,7 @@ def assert_xml_task_properties(xml_task):
     assert xml_task.user_id is None
     assert xml_task.logon_type is None
     assert xml_task.group_id == "S-1-5-4"
-    assert xml_task.display_name is None
+    assert xml_task.display_name is "test_xml.xml"
```
Schamper, line 43:
```suggestion
    assert xml_task.display_name == "test_xml.xml"
```
Usually you only use a `is` comparison on things like `None` or class types. Any other type of comparison should be done with `==`.

Schamper (changes requested):

Schamper (approved):

PR #403, opened 2023-09-29
Add DDF volume system support

Schamper:
I tried to merge it with the MD volume system but didn't really like the way it ended up, so opted to just make it a separate one for now.

There's a lot of reusable code in the whole volume system stuff anyway, something for a future cleanup perhaps.


dissect/target/volume.py
```diff
@@ -22,6 +22,8 @@
 """A lazy import of :mod:`dissect.target.volumes.vmfs`."""
 md = import_lazy("dissect.target.volumes.md")
 """A lazy import of :mod:`dissect.target.volumes.md`."""
+ddf = import_lazy("dissect.target.volumes.ddf")
```
cecinestpasunepipe, line 25: I tried to open a RAID-0 target consisting of 2 DDF RAID-0 disks with EXT4 trough the multirawloader but it does not seem to work? Is there some wiring missing somewhere?

`target-query s1.img+s2.img -f users`

```
s1.img+s2.img: Attempting to use loader: <lazyattr dissect.target.loaders.multiraw.MultiRawLoader loaded=True> [dissect.target.target] 
2023-10-18T14:42:59.404293Z [debug    ] <Target s1.img+s2.img>: LVM volumes found: [<Volume name=None size=10737418240 fs=None>, <Volume name=None size=10737418240 fs=None>] [dissect.target.target] 
2023-10-18T14:42:59.404357Z [debug    ] <Target s1.img+s2.img>: Encrypted volumes found: [] [dissect.target.target] 
2023-10-18T14:42:59.411051Z [debug    ] <Target s1.img+s2.img>: Opened LVM: <DdfVolumeSystem serial=None> [dissect.target.target] 
2023-10-18T14:42:59.471100Z [warning  ] <Target s1.img+s2.img>: Can't identify filesystem: <Volume name='0' size=21407727616 fs=None> [dissect.target.target]
```

cecinestpasunepipe (changes requested):

dissect/target/volume.py
Schamper: Can you share these files with me for debugging?

dissect/target/target.py
```diff
@@ -730,7 +736,25 @@ def apply(self) -> None:
...
                     encrypted_volumes.append(vol)
                 else:
-                    self.open(vol)
+                    # We could be getting "regular" volume systems out of LVM or encrypted volumes
+                    # Try to open each volume as a regular volume system, or add as a filesystem if it fails
+                    try:
+                        vs = volume.open(vol)
+                    except Exception:
+                        # If opening a volume system fails, there's likely none, so open as a filesystem instead
+                        self.open(vol)
```
cecinestpasunepipe, line 745: I would recommend to add a log message here, for diagnostic purposes.

cecinestpasunepipe (changes requested):

dissect/target/target.py
Schamper: Technically this is the happy path though (it's relatively rare that a volume has another default volume system), so this will result in a lot of unnecessary log messages. Would you still want a log message then?

cecinestpasunepipe: Should the code then not be restructured to make it reflect this fact better?

cecinestpasunepipe (changes requested):

dissect/target/target.py
Schamper: Do you have a suggestion for doing so? Other than adding a comment.

cecinestpasunepipe: Did some research, I see that it is not possible currently.

cecinestpasunepipe (approved):

PR #404, opened 2023-09-29
Add LUKS volume support

Schamper:


dissect/target/volumes/luks.py
```diff
@@ -0,0 +1,116 @@
...
+        stream = self.unlock_volume()
+        yield Volume(
+            fh=stream,
+            size=stream.size,
+            raw=self.fh,
+            vs=self,
+            **volume_details,
+        )
+
+    def unlock_with_volume_encryption_key(self, key: bytes, keyslot: Optional[int] = None) -> None:
```
cecinestpasunepipe, line 55: Does not seem to work?

`self.unlock_with_passphrase("luks", 0)`

ValueError: No valid keyslot found.
Tested on latest Ubuntu. Some usage documentation might be handy.
Also, is there a mechanism to provide a key through an ENV var or option?

cecinestpasunepipe (changes requested):

dissect/target/volumes/luks.py
Schamper: Works for me, are you using the correct passphrase and keyslot?

Also, the intended way is to use the keychain, for which you can either provide a csv file using`-K <path>` or a value using `-Kv <value>`. For example:

```bash

❯ target-shell -p ../dissect.fve/tests/data/luks_aes-cbc-essiv.bin -Kv password -q
Python 3.11.5 (main, Aug 24 2023, 12:23:19) [Clang 15.0.0 (clang-1500.0.40.1)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.13.2 -- An enhanced Interactive Python. Type '?' for help.


Loaded targets in 'targets' variable. First target is in 't'.

In [1]: t.volumes[1].vs.luks.unlocked
Out[1]: True
```

cecinestpasunepipe: I can't get it to work with the -Kv parameter, let's say I make an empty luks volume like this:

```
dd if=/dev/zero of=/path/to/lucky_luks.img bs=1M count=512
cryptsetup -vy luksFormat /path/to/lucky_luks.img
sudo cryptsetup luksOpen /path/to/lucky_luks.img lucky_luks_volume
sudo mkfs.ext4 /dev/mapper/lucky_luks_volume
```

But opening it with:

`target-shell /path/to/lucky_luks.img  -Kv luks`

gives an error (Error: Group descriptor block locations exceed last block), yet:

```
dd if=/dev/zero of=/path/to/noluks.img bs=1M count=512
mkfs.ext4 -F /path/to/noluks.img
target-shell /path/to/noluks.img
```

works as expected?
What do I miss?

Schamper: For documentation purposes: this was picked up offline and turned out to be a two-fold problem:
- LUKS1 is not yet supported, which it turns out is what cryptsetup formatted it as by default
- There was a bug in dissect.fve that resulted in LUKS volumes with a large sector size (4k) being decrypted incorrectly

cecinestpasunepipe: Confirmed, seems to work now.

cecinestpasunepipe: I would add a warning or error if the LUKS version is 1 because now I can still open it but I will only see an empty disk.

dissect/target/volumes/luks.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            if keyslot is None:
+                for keyslot in self.luks.keyslots.keys():
+                    try:
+                        self.luks.unlock(key, keyslot)
+                        break
+                    except ValueError:
+                        continue
+                else:
+                    raise ValueError("Failed to find matching keyslot for provided volume encryption key")
```
cecinestpasunepipe, line 65: Can you also add an error if the LUKS version is 1?

cecinestpasunepipe (changes requested):

dissect/target/volumes/luks.py
Schamper: I improved the error logging in volume.py to make the existing error more visible.

dissect/target/volumes/luks.py
Schamper: I improved the error logging in volume.py to make the existing error more visible.

cecinestpasunepipe (approved):

Issue #256, opened 2023-05-16
Plugin errors for domain and datetime


pemontto:
I've installed dissect via pip a few different ways on different python version 3.9 through 3.11 and used the ghcr.io/fox-it/dissect:3.5 docker image. In all cases I get errors running target-info on both amd64 and aarch64.

Testing with some public E01 and vmdk files I'm getting plugin errors related to domain and datetime. Most other functions/commands seem to work however.

## [2020JimmyWilson.E01](https://drive.google.com/file/d/1Fd1pX1r4waRkD6Z2O8J5cRZyeSNU5-SY/view)
```bash
# target-shell 2020JimmyWilson.E01
2023-05-16T14:01:23.101142Z [warning  ] Failed to import <lazyattr dissect.target.loaders.asdf.AsdfLoader loaded=False> [dissect.target.loader]
2023-05-16T14:01:23.163181Z [warning  ] Failed to import <lazyattr dissect.target.volumes.bde.BitlockerVolumeSystem loaded=False> [dissect.target.volume]
2023-05-16T14:01:23.201831Z [warning  ] <Target 2020JimmyWilson.E01>: Can't identify filesystem: <Volume name='Microsoft reserved partition' size=33553920 fs=None> [dissect.target.target]
2023-05-16T14:01:23.202738Z [warning  ] Failed to import <lazyattr dissect.target.volumes.bde.BitlockerVolumeSystem loaded=False> [dissect.target.volume]
2023-05-16T14:01:23.492200Z [warning  ] <Target 2020JimmyWilson.E01>: Empty hive: sysvol/windows/system32/config/RegBack/SYSTEM [dissect.target.target]
IACIS-HDD-2014 /> domain
None
IACIS-HDD-2014 /> architecture
amd64-win64
IACIS-HDD-2014 /> hostname
IACIS-HDD-2014
IACIS-HDD-2014 /> os
windows
IACIS-HDD-2014 /> ips
['192.168.0.170']
IACIS-HDD-2014 /> install_date
2014-01-25 01:27:46+00:00
IACIS-HDD-2014 /> activity
2015-05-26 13:05:11.623703+00:00



# target-info 2020JimmyWilson.E01
2023-05-16T13:29:10.110702Z [warning  ] Failed to import <lazyattr dissect.target.loaders.asdf.AsdfLoader loaded=False> [dissect.target.loader]
[...]

PR #405, opened 2023-10-02
Improve robustness of target-info

JSCU-CNI:
This PR aims to improve the stability of target-info when used on targets without optional `_os.py` functions.

### DIFF
--- a/dissect/target/tools/info.py
+++ b/dissect/target/tools/info.py
@@ -90,29 +90,26 @@ def main():
 
 
 def get_target_info(target: Target) -> dict[str, Union[str, list[str]]]:
-    info = {
+    return {
         "disks": get_disks_info(target),
         "volumes": get_volumes_info(target),
         "children": get_children_info(target),
         "hostname": target.hostname,
-        "domain": None,
+        "domain": get_optional_func(target, "domain"),
         "ips": target.ips,
         "os_family": target.os,
         "os_version": target.version,
         "architecture": target.architecture,
-        "language": None,
-        "timezone": None,
-        "install_date": None,
-        "activity": None,
+        "language": get_optional_func(target, "language"),
+        "timezone": get_optional_func(target, "timezone"),
+        "install_date": get_optional_func(target, "install_date"),
+        "last_activity": get_optional_func(target, "activity"),
     }
 
-    optional = ["domain", "language", "timezone", "install_date", "activity"]
 
-    for func in optional:
-        if target.has_function(func):
-            info[func] = getattr(target, func)
-
-    return info
+def get_optional_func(target: Target, func: str) -> Union[str, None]:
+    if target.has_function(func):
+        return getattr(target, func)
 
 
 def print_target_info(target: Target) -> None:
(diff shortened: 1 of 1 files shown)


dissect/target/tools/info.py
```diff
@@ -90,49 +90,51 @@ def main():
...
+        "timezone": None,
+        "install_date": None,
+        "activity": None,
     }
 
+    optional = ["domain", "language", "timezone", "install_date", "activity"]
+
+    for func in optional:
+        if target.has_function(func):
+            info[func] = getattr(target, func)
```
pyrco, line 113:
```suggestion
        info[func] = getattr(target, func, None)
```

Would this work?
In this form you could also assign it immediately in the dict without the need to loop over a list of optional keys.

dissect/target/tools/info.py
```diff
@@ -90,49 +90,51 @@ def main():
...
-    print(f"Last activity : {target.activity}")
+
+    for name, value in get_target_info(target).items():
+        if name in ["disks", "volumes", "children"]:
+            if not any(value):
+                continue
+            print(f"\n{name.capitalize()}")
+            for i in value:
+                values = " ".join([f'{k}="{v}"' for k, v in i.items()])
+                print(f"- <{name.capitalize()} {values}>")
```
pyrco, line 128: If you're special casing anyway for these three keys, maybe strip the "s" from Disks and Volumes and "ren" from Children

pyrco (changes requested):

dissect/target/tools/info.py
```diff
@@ -90,49 +90,51 @@ def main():
...
+            continue
+
+        if isinstance(value, list):
+            value = ", ".join(value)
+
+        if name == "hostname":
+            print()
+
+        print(f"{name.capitalize().replace('_', ' ')}" + (14 - len(name)) * " " + f" : {value}")
```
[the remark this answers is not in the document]
JSCU-CNI: Your suggestion would break backwards compatibility with the json and flow record output. We do use those output formats extensively in an automated fashion.

dissect/target/tools/info.py
JSCU-CNI: Unfortunately this would not work as the `getattr` would fail when calling an incompatible or non-existing plugin for the current target. Which is why the `has_function` method exists.

JSCU-CNI: I've added a small improvement regarding your suggestion in [1cc7663](https://github.com/fox-it/dissect.target/pull/405/commits/1cc7663a82c8a520771a0357eaa3c4ba745a9fba).

dissect/target/tools/info.py
JSCU-CNI: Good catch, added in [b3e4af0](https://github.com/fox-it/dissect.target/pull/405/commits/b3e4af017484478f0d290b1a1f843ca82484edc6)

dissect/target/tools/info.py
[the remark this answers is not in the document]
pyrco: Ah, check, I missed that part.

pyrco (approved):
Once the other MRs linked to #256 are fixed, the only thing left is to make the `architecture` function an official OSPlugin function. We have an internal ticket for this. So this MR is good to go.

PR #408, opened 2023-10-05
Fix plugin lookups for the DefaultPlugin

pyrco:
(DIS-2357)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -532,10 +532,10 @@ def plugins(osfilter: Optional[type[OSPlugin]] = None) -> Iterator[PluginDescrip
     """Retrieve all plugin descriptors.
 
     Args:
-        osfilter: The OS module path the plugin should be from.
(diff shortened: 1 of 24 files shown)


dissect/target/plugin.py
```diff
@@ -528,14 +528,14 @@ def decorator(obj):
     return decorator
 
 
-def plugins(osfilter: Optional[type[OSPlugin]] = None) -> Iterator[PluginDescriptor]:
+def plugins(osfilter: Optional[OSPlugin] = None) -> Iterator[PluginDescriptor]:
```
Schamper, line 531: Nice, ran into this yesterday as well but didn't push the fix yet.

Isn't the type-hint still correct though?

pyrco: Yeah, that was a brain fart on my part. The check is wrong, but the typing is correct. I see a lot of testcases fail though, apparently I only tested the newly created test case :raised_eyebrow:

Schamper (dismissed):

Schamper (approved):

PR #414, opened 2023-10-12
Add the architecture function to OSPlugin

pyrco:
This function is now mandatory to be implemented by all OSPlugin subclasses.

(DIS-2386)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -377,10 +377,22 @@ class OSPlugin(Plugin):
     def os(self) -> str:
         """Required OS function.
 
-        Implementations must be decorated with ``@export(property=True)``
+        Implementations must be decorated with ``@export(property=True)``.
 
         Returns:
             A slug of the OS name, e.g. 'windows' or 'linux'.
+        """
+        raise NotImplementedError
+
+    @export(property=True)
+    def architecture(self) -> Optional[str]:
+        """Required OS function.
+
+        Implementations must be decorated with ``@export(property=True)``.
+
+        Returns:
+            A slug of the OS architecture, e.g. 'x86_32-unix', 'MIPS-linux' or
+            'AMD64-win32', or 'unknown' if the architecture is unknown.
         """
         raise NotImplementedError
 
--- a/dissect/target/plugins/general/default.py
+++ b/dissect/target/plugins/general/default.py
@@ -1,3 +1,13 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from flow.record import Record
+
+    from dissect.target.target import Target
+    from dissect.target.filesystem import Filesystem
+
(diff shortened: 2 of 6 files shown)


dissect/target/plugin.py
```diff
@@ -384,6 +384,18 @@ def os(self) -> str:
         """
         raise NotImplementedError
 
+    @export(property=True)
+    def architecture(self) -> Optional[str]:
+        """Required OS function.
+
+        Implementations must be decorated with ``@export(property=True)``
```
Schamper, line 391:
```suggestion
        Implementations must be decorated with ``@export(property=True)``.
```

dissect/target/plugin.py
```diff
@@ -384,6 +384,18 @@ def os(self) -> str:
...
+
+        Returns:
+            A slug of the OS architecture, e.g. 'x86_32-unix', 'MIPS-linux' or
```
Schamper, line 394:
```suggestion
            A slug of the OS architecture, e.g. 'x86_32-unix', 'MIPS-linux' or
```
Should we standardize this to target triplets? https://wiki.osdev.org/Target_Triplet

dissect/target/plugins/os/unix/bsd/osx/_os.py
```diff
@@ -97,3 +97,9 @@ def users(self) -> Iterator[UnixUserRecord]:
     @export(property=True)
     def os(self) -> str:
         return OperatingSystem.OSX.value
+
+    @export(property=True)
+    def architecture(self) -> str:
```
Schamper, line 102:
```suggestion
    def architecture(self) -> Optional[str]:
```

Schamper (changes requested):

dissect/target/plugin.py
pyrco: That would be nice yes.

dissect/target/plugins/general/default.py
```diff
@@ -14,31 +24,35 @@ def __init__(self, target):
                 target.fs.mount(f"fs{i}", fs)
 
     @classmethod
-    def detect(cls, target):
+    def detect(cls, target: Target):
```
Schamper, line 27:
```suggestion
    def detect(cls, target: Target) -> Optional[Filesystem]:
```

dissect/target/plugins/general/default.py
```diff
@@ -14,31 +24,35 @@ def __init__(self, target):
...
         pass
 
     @classmethod
-    def create(cls, target, sysvol):
+    def create(cls, target: Target, sysvol: Filesystem):
```
Schamper, line 31:
```suggestion
    def create(cls, target: Target, sysvol: Filesystem) -> DefaultPlugin:
```

Schamper (changes requested):

Schamper (approved):

PR #417, opened 2023-10-16
Add JFFS2 support to Dissect

JSCU-CNI:
This PR adds support for the Journalling Flash File System version 2 to Dissect.

Depends on https://github.com/fox-it/dissect.jffs/pull/1


dissect/target/containers/vhd.py
```diff
@@ -18,12 +18,14 @@ def __init__(self, fh: Union[BinaryIO, Path], *args, **kwargs):
...
-
-        return b"conectix" in magic
+        try:
+            offset = fh.tell()
+            fh.seek(-512, io.SEEK_END)
+            magic = fh.read(9)
+            fh.seek(offset)
+            return b"conectix" in magic
+        except Exception:
+            return False
```
JSCU-CNI, line 28: JFFS2 filesystems can be [smaller than 512 bytes](https://github.com/fox-it/dissect.jffs/pull/1/files#diff-b86fab974720cbe31570915ef800d0a155a2bf807fd3bbca378855bdf93d0a43). Please let us know if a different fix in `vhd.py` is desired.

Schamper: The volume and filesystem detect methods have a small wrapper that deals with errors like these. We'll need to add that to container down the line too. For now, this will do. Could you change the `Exception` to the specific error that is raised (`OSError` IIRC)?

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+            raise NotASymlinkError(path, cause=e)
+        except jffs2.Error as e:
+            raise FileNotFoundError(path, cause=e)
+
+
+class JFFSFilesystemEntry(FilesystemEntry):
+    def _resolve(self) -> FilesystemEntry:
+        if self.is_symlink():
+            return self.readlink_ext()
+        return self
```
Schamper, line 53: This is now handled by `FilesystemEntry` so this can be removed.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+            raise IsADirectoryError(self.path)
+        return self._resolve().entry.open()
+
+    def iterdir(self) -> Iterator[str]:
+        if not self.is_dir():
+            raise NotADirectoryError(self.path)
+
+        if self.is_symlink():
+            for name, entry in self.readlink_ext().iterdir():
+                yield name, entry
```
Schamper, line 71: `iterdir()` should only yield the names.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            for name, entry in self.entry.iterdir():
+                yield name, entry
+
+    def scandir(self) -> Iterator[FilesystemEntry]:
+        for name, entry in self.iterdir():
```
Schamper, line 77: Which means you also have to change this around a bit.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+            entry_path = fsutil.join(self.path, name, alt_separator=self.fs.alt_separator)
+            yield JFFSFilesystemEntry(self.fs, entry_path, entry)
+
+    def is_dir(self, follow_symlinks: bool = False) -> bool:
+        try:
+            return self._resolve().entry.is_dir()
```
Schamper, line 83:
```suggestion
            return self._resolve(follow_symlinks).entry.is_dir()
```
If you remove the `_resolve()` in this file.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+        except FilesystemError:
+            return False
+
+    def is_file(self, follow_symlinks: bool = False) -> bool:
+        try:
+            return self._resolve().entry.is_file()
```
Schamper, line 89:
```suggestion
            return self._resolve(follow_symlinks).entry.is_file()
```
Same here.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+        return self.entry.is_symlink()
+
+    def readlink(self) -> str:
+        if not self.is_symlink():
+            raise NotASymlinkError()
+
+        return self.entry.link
+
+    def stat(self, follow_symlinks: bool = False) -> fsutil.stat_result:
+        return self._resolve().lstat()
```
Schamper, line 103:
```suggestion
        return self._resolve(follow_symlinks).lstat()
```
Same here.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def lstat(self) -> fsutil.stat_result:
+        node = self.entry.inode
+
+        # mode, ino, dev, nlink, uid, gid, size, atime, mtime, ctime
+        st_info = fsutil.stat_result(
+            [
+                self.entry.mode,
+                self.entry.inum,
+                0,
```
Schamper, line 113:
```suggestion
                id(self.fs),
```

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,124 @@
...
+                0,
```
Schamper, line 114:
```suggestion
                1,
```
You could calculate a proper `nlink` number in dissect.jffs by keeping track of `dirent.ino` in a new dict and resetting the number to `0` when there's an unlink. Then expose that value in a new `nlink` property on the `INode` class.

I think it's also fine to "lie" for now and just say it's 1. If you don't decide to implement this properly, can you add a comment that it could be done in the future? Then again it should be a small addition to dissect.jffs which is still in review anyway.

Schamper (changes requested):

dissect/target/containers/vhd.py
JSCU-CNI: Added in 4729cde6cdfde5a5368f1adbad6daa71d9fa9732

dissect/target/filesystems/jffs.py
JSCU-CNI: Removed in 4729cde6cdfde5a5368f1adbad6daa71d9fa9732.

dissect/target/filesystems/jffs.py
JSCU-CNI: Changed to `1` for now until we implement the proper `nlink` counting in `dissect.jffs`.

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,120 @@
...
+    def lstat(self) -> fsutil.stat_result:
+        node = self.entry.inode
+
+        # mode, ino, dev, nlink, uid, gid, size, atime, mtime, ctime
+        st_info = fsutil.stat_result(
+            [
+                self.entry.mode,
+                self.entry.inum,
+                id(self.fs),
+                1, # TODO: properly calculate nlink in dissect.jffs
```
Schamper, line 110:
```suggestion
                1,  # TODO: properly calculate nlink in dissect.jffs
```

Schamper (changes requested):

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,120 @@
...
+
+        return magic in (
+            c_jffs2.c_jffs2.JFFS2_MAGIC_BITMASK,
+            c_jffs2.c_jffs2.JFFS2_OLD_MAGIC_BITMASK,
+        )
+
+    def get(self, path: str) -> FilesystemEntry:
+        return JFFSFilesystemEntry(self, path, self._get_node(path))
+
+    def _get_node(self, path: str, node: Optional[jffs2.INode] = None):
```
Schamper, line 36:
```suggestion
    def _get_node(self, path: str, node: Optional[jffs2.INode] = None) -> jffs2.INode:
```

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,120 @@
...
+            for name, entry in self.readlink_ext().iterdir():
+                yield name
+        else:
+            for name, entry in self.entry.iterdir():
+                yield name
+
+    def scandir(self) -> Iterator[FilesystemEntry]:
+        for name in self.iterdir():
+            entry_path = fsutil.join(self.path, name, alt_separator=self.fs.alt_separator)
+            entry = self.fs.get(entry_path).entry
```
Schamper, line 74: Can you tweak the logic of `iterdir()` and `scandir()` to be closer to this:
https://github.com/fox-it/dissect.target/blob/main/dissect/target/filesystems/ffs.py#L62-L83

The logic here being that by iterating over the `INode.iterdir()` items, you already have the `INode` and don't need to do another lookup with `self.fs.get()` and can instead re-use that object.

Schamper (changes requested):

dissect/target/filesystems/jffs.py
JSCU-CNI: That makes sense, added in [80c5706](https://github.com/fox-it/dissect.target/pull/417/commits/80c570630c5e6db80c84255b409e96d47c9e2197).

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,123 @@
...
+            raise IsADirectoryError(self.path)
+        return self._resolve().entry.open()
+
+    def _iterdir(self) -> Iterator[tuple[str, jffs2.INode]]:
+        if not self.is_dir():
+            raise NotADirectoryError(self.path)
+
+        if self.is_symlink():
+            for name, entry in self.readlink_ext().iterdir():
+                yield name, entry
```
Schamper, line 66:
```suggestion
            yield from self.readlink_ext().iterdir()
```
Last small suggestion I see for now, seems ready to go until dissect.jffs!

dissect/target/filesystems/jffs.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            for name, entry in self.entry.iterdir():
+                yield name, entry
```
Schamper, line 69:
```suggestion
            yield from self.entry.iterdir()
```

Schamper (dismissed):

Schamper (approved):

PR #419, opened 2023-10-16
Add support for Android Backups

Schamper:
Also some miscellaneous cleanup.

Depends on https://github.com/fox-it/dissect.util/pull/33.


dissect/target/plugins/os/unix/linux/android/_os.py
```diff
@@ -66,5 +55,19 @@ def version(self) -> str:
     def os(self) -> str:
         return OperatingSystem.ANDROID.value
 
-    def users(self) -> Iterator[UnixUserRecord]:
-        raise NotImplementedError()
+    @export(record=EmptyRecord)
```
JSCU-CNI, line 58: This addition would temporarily fix https://github.com/fox-it/dissect.target/issues/421.

dissect/target/loaders/ab.py
```diff
@@ -0,0 +1,369 @@
...
+                raise ValueError("Missing password for encrypted Android Backup")
+            fh = CipherStream(fh, self._mk, self._iv, self.size)
+
+        if self.compressed:
+            fh = ZlibStream(fh)
+
+        return fh
+
+
+class ZlibStream(AlignedStream):
```
cecinestpasunepipe, line 198: Would it not be better to move this to the dissect.util streams?

dissect/target/loaders/ab.py
```diff
@@ -0,0 +1,369 @@
...
+
+        return data
+
+
+def _encode_bytes(buf: bytes) -> bytes:
+    # Emulate byte[] -> char[] -> utf8 byte[] casting
+    return struct.pack(">32h", *struct.unpack(">32b", buf)).decode("utf-16-be").encode("utf-8")
+
+
+def main() -> None:
```
cecinestpasunepipe, line 325: Maybe we could convert this into loader arguments?

cecinestpasunepipe (changes requested):

dissect/target/loaders/ab.py
Schamper: That's possible, but I'm not sure if it improves the usability. Then you suddenly have a loader that acts like a separate utility with the right combination of arguments, instead of just acting like a loader. What do you think?

dissect/target/loaders/ab.py
Schamper: https://github.com/fox-it/dissect.util/pull/38

dissect/target/plugins/os/unix/linux/android/_os.py
```diff
@@ -66,5 +55,21 @@ def version(self) -> str:
...
+    for line in fh:
+        line = line.strip()
+
+        if not line or line.startswith("#"):
+            continue
+
+        k, v = line.split("=")
+        result[k] = v
+
+    return result
```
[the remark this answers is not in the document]
Schamper: FIxed

dissect/target/plugins/apps/browser/firefox.py
```diff
@@ -9,7 +9,7 @@
 from dissect.target.exceptions import FileNotFoundError, UnsupportedPluginError
```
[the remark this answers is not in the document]
Schamper: Done: #752

dissect/target/loaders/ab.py
```diff
@@ -0,0 +1,287 @@
...
+        target.fs.mount("/", vfs)
+        target._os_plugin = AndroidPlugin(target)
+
+
+class AndroidBackup:
+    def __init__(self, fh: BinaryIO):
+        self.fh = fh
+
+        size = fh.seek(0, io.SEEK_END)
+        fh.seek(0)
```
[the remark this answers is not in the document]
Schamper: No, that's an "error" we often made in older parser code. It's better to assume that the file starts at offset 0. If you really need a relative Android Backup, use `RelativeStream`.

dissect/target/loaders/ab.py
```diff
@@ -0,0 +1,287 @@
...
+                raise ValueError("Missing password for encrypted Android Backup")
+            fh = CipherStream(fh, self._mk, self._iv, self.size)
+
+        if self.compressed:
+            fh = ZlibStream(fh)
+
+        return fh
+
+
+class CipherStream(AlignedStream):
```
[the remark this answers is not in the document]
Schamper: Only if it turns out it's 1:1 compatible with other encrypted files/streams. I'd say leave it here for now and only when that is the case, move it there.

PR #426, opened 2023-10-26
Restructure tests and some plugins

Schamper:


pyrco (commented):
LGTM. Be aware when rebasing that MANIFEST.in also needs to be updated.

Do we need to communicate at the next release the fact that certain plugins will have a different long path name when called?

Schamper:
I've rebased, everything should be up to date now. I've also updated the MANIFEST.in with the new data path.

> Do we need to communicate at the next release the fact that certain plugins will have a different long path name when called?

The canonical name doesn't change, but it would be nice to mention it regardless.

pyrco (approved):

PR #429, opened 2023-10-31
Add Tab URL's and MIME type to download record

Zawadidone:
In some cases the download URL isn't a valid link but a GUID `blob:null/<GUID>`, while the other tab URL related fields contain useful information about the origin of the downloaded file.

### DIFF
--- a/dissect/target/plugins/browsers/browser.py
+++ b/dissect/target/plugins/browsers/browser.py
@@ -9,7 +9,10 @@ GENERIC_DOWNLOAD_RECORD_FIELDS = [
     ("varint", "id"),
     ("path", "path"),
     ("uri", "url"),
+    ("uri", "tab_url"),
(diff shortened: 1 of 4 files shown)


Zawadidone:
Is it better to add the fields to the `ChromiumMixin` class?

pyrco:
If these fields are only applicable to all chromium based browsers (Edge also does it this way?), then I'd add them there. If something like `mime_type` also becomes available for Firefox, we can always migrate that property to the generic record fields.

Zawadidone:
Okay I will add the fields to the record (https://github.com/Zawadidone/dissect.target/blob/feature/chromium_tab_urls/dissect/target/plugins/browsers/chromium.py#L31), yes Edge also uses these fields.

pyrco (approved):

PR #430, opened 2023-10-31
Change the RunKeyRecord path field to string

pyrco:
The "path" value of a run key is actually a full command. If this command does not have just a path but also things like command parameters, this can not be parsed as a path type.

The field is now treated as a normal string to prevent parse errors and in anticipation of a proper command field type.

(DIS-2563)

### DIFF
--- a/dissect/target/plugins/os/windows/regf/runkeys.py
+++ b/dissect/target/plugins/os/windows/regf/runkeys.py
@@ -1,5 +1,3 @@
-from flow.record.fieldtypes import path
-
 from dissect.target.exceptions import UnsupportedPluginError
 from dissect.target.helpers.descriptor_extensions import (
     RegistryRecordDescriptorExtension,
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/regf/runkeys.py
```diff
@@ -65,7 +63,7 @@ def runkeys(self):
             domain (string): The target domain.
             ts (datetime): The registry key last modified timestamp.
             name (string): The run key name.
-            path (string): The run key path.
+            path (string): The run key command.
```
Poeloe, line 66: I would suggest to then also change the record name. But maybe it's better to do that with the full fix.

So long story short, I would suggest to keep it path here.

Poeloe (changes requested):

Poeloe (approved):
LGTM!

Issue #413, opened 2023-10-11
target-query --hash function broken


JSCU-CNI:
The `--hash` function of `target-query` seems to be broken.

Example stacktrace (this also occurs with `walkfs` and `mft` plugins, both on Windows and GNU/Linux):

```
$ target-query -f lnk /tmp/windows10.vmdk --limit 1 --hash

Traceback (most recent call last):
  File "/tmp/dissect/.venv/bin/target-query", line 8, in <module>
    sys.exit(main())
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/tools/utils.py", line 250, in wrapper
    return func(*args, **kwargs)
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/tools/query.py", line 338, in main
    raise e
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/tools/query.py", line 325, in main
    rs.write(hashutil.hash_uri_records(target, record_entries))
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/helpers/hashutil.py", line 70, in hash_uri_records
    uri_record = hash_uri(target, getattr(record, name))
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/helpers/hashutil.py", line 95, in hash_uri
    path = target.resolve(path)
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/plugins/filesystem/resolver.py", line 38, in resolve
    return self.resolve_windows(path, user_sid=user)
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/plugins/filesystem/resolver.py", line 44, in resolve_windows
    path = fsutil.normalize(path, alt_separator=self.target.fs.alt_separator)
  File "/tmp/dissect/.venv/lib/python3.10/site-packages/dissect/target/helpers/fsutil.py", line 46, in normalize
    return re_normalize_sbs_path.sub("/", path)
TypeError: expected string or bytes-like object
```

It seems that the `re_normalize_sbs_path.sub("/", path)` call expects `path` to be a string, while in reality it is a `flow.record.fieldtypes.windows_path` instance. Casting `path` to a string seems to fix the issue.

```python
def normalize(path: str, alt_separator: str = "") -> str:
    if alt_separator == "\\":
        return re_normalize_sbs_path.sub("/", str(path))
    else:
[...]

PR #432, opened 2023-11-01
Fix the hashutil functions to properly deal with path record field types

pyrco:
(DIS-2526)

### DIFF
--- a/dissect/target/helpers/hashutil.py
+++ b/dissect/target/helpers/hashutil.py
@@ -1,14 +1,16 @@
 from __future__ import annotations
 
 import hashlib
-from typing import TYPE_CHECKING
+import warnings
+from typing import TYPE_CHECKING, BinaryIO, Union
 
-from flow.record import GroupedRecord, Record, RecordDescriptor
-from flow.record.fieldtypes import path, posix_path, windows_path
+from flow.record import GroupedRecord, Record, RecordDescriptor, fieldtypes
 
 from dissect.target.exceptions import FileNotFoundError, IsADirectoryError
 
 if TYPE_CHECKING:
+    from hashlib._hashlib import HASH
+
     from dissect.target.target import Target
 
 BUFFER_SIZE = 32768
@@ -22,7 +24,7 @@ HashRecord = RecordDescriptor(
 )
 
 
-def _hash(fh, ctx):
+def _hash(fh: BinaryIO, ctx: Union[HASH, list[HASH]]) -> tuple[str]:
     if not isinstance(ctx, list):
         ctx = [ctx]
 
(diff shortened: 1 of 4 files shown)


pyrco:
This should fix #413

pyrco:
I made `hash_uri()` and `hash_uri_records()` deprecated and let them point to the new functions for now. Any thoughts on removing them completely already in this PR @Schamper, @Miauwkeru?

dissect/target/helpers/hashutil.py
```diff
@@ -60,36 +62,42 @@ def custom(fh, algos):
...
+
+
+def hash_path(target: Target, path: fieldtypes.path) -> Record:
     """Hash the target path."""
     if path is None:
         raise FileNotFoundError()
 
-    path = target.resolve(path)
-    return (path, target.fs.hash(path))
+    return (path, target.fs.hash(str(path)))
```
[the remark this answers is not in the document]
pyrco: `path` type entries should be proper paths, fully resolved if possible, so I don't think it is the responsibility of this function to do that.

dissect/target/helpers/hashutil.py
```diff
@@ -60,34 +62,73 @@ def custom(fh, algos):
...
+                resolved_path = target.resolve(path)
+                path_hash = target.fs.hash(resolved_path)
+            except (FileNotFoundError, IsADirectoryError):
+                pass
+            else:
+                if path != resolved_path:
+                    warnings.warn(
+                        (
+                            "The function passing in a path field did not resolve that path. ",
+                            "The autmatic resolving of path fields will be removed in the future.",
```
Schamper, line 105:
```suggestion
                            "The automatic resolving of path fields will be removed in the future.",
```

Schamper: Why, by the way?

Schamper (changes requested):

dissect/target/helpers/hashutil.py
pyrco: Because this part of the code should not have any knowledge about what the meaning of the content of a field is. That is the responsibility of the plugin creating the record. The plugin knows whether it is useful to resolve a path or not. It's correct separation of responsibilities.

Schamper: What about when for correctness I want to put the original path from e.g. the registry in the record which might include an environment variable, but want to hash that file? Should I create a second field like `real_path`?

pyrco: That is a good point. You do want to preserve original values. Though putting unresolved paths with e.g. env variables feels problematic. In my mind a (pathlib style) path type means a literal path, where percentages, dollars and all are part of the actual path, not of potential env variables, otherwise you would still be guessing if it is an actual path or if it still needs to be resolved.

In an ideal world I think you would have 2 values, a literal string which contains whatever was parsed from the artifact and a path which is resolved and present on the target. You would still have the side cases of: 1. what if the path correctly resolves but is not present in the target and 2. what if the path does not or partially resolve how can you be sure the dollar/percentage is part of the path or an unresolved env variable.

Although the last side case won't be solvable in practice, I do think we need to move to a system where a path is as fully resolved as possible and if the literal value, is of value it should be present in the record as a second (string) value. Also because people tend to want to put paths that come out of target-query into something like target-fs to get to the file, which will fail if it is not resolved (we could of course change target-fs to (have the option to) resolve paths).

But given that this is more a policy discussion then something that should be realized by this PR, I'll revert the deprecation warning on unresolved paths for now.

PR #439, opened 2023-11-10
Do not package test data in the Python source distribution

pyrco:
The amount of test data is such that we fairly quickly run out of storage on pypi.org.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,6 +22,18 @@ from dissect.target.target import Target
 
 from ._utils import absolute_path
 
+# Test if the data/ directory is present and if not, as is the case in Python
+# source distributions of dissect.target, we give an error
+data_dir = absolute_path("data")
+if not pathlib.Path(data_dir).is_dir():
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -22,6 +22,18 @@
...
 from ._utils import absolute_path
 
+# Test if the data/ directory is present and if not, as is the case in Python
+# source distributions of dissect.target, we give an error
+data_dir = absolute_path("data")
+if not pathlib.Path(data_dir).is_dir():
+    raise pytest.PytestConfigWarning(
+        f"No test data directory {data_dir} found.\n"
+        "This can happen when you have downloaded the source distribution\n"
+        "of dissect.target from pypi.org. If so, retreive the test data from\n"
```
Schamper, line 32:
```suggestion
        "of dissect.target from pypi.org. If so, retrieve the test data from\n"
```

Schamper (changes requested):

Schamper (approved):

PR #436, opened 2023-11-07
Fix support for unicode paths in target-shell

JSCU-CNI:
This PR adds [`lexer.whitespace_split = True`](https://docs.python.org/3/library/shlex.html#shlex.shlex.whitespace_split) to `shlex` in `target-shell` which improves handling unicode arguments for argparse.

This fixes `ls` and `cat` in `target-shell` when paths contain unicode characters.

PR #440, opened 2023-11-13
Fix support for unicode paths in target-shell (copy)

pyrco:
This PR adds [lexer.whitespace_split = True](https://docs.python.org/3/library/shlex.html#shlex.shlex.whitespace_split) to shlex in target-shell which improves handling unicode arguments for argparse.

This fixes ls and cat in target-shell when paths contain unicode characters.


pyrco:
I could not update the original branch in #436, so I duplicated the PR here so it can be merged

PR #441, opened 2023-11-13
Add support for docker containers as children

JSCU-CNI:
This PR adds child target support to Dissect for docker containers and depends on #507.

To demonstrate:
```bash
$ target-query -f os,version --children example.img
<Target example.img> linux Ubuntu 22.04.3 LTS (Jammy Jellyfish)
<Target /var/lib/docker/image/overlay2/layerdb/mounts/852d704264f1078107bd593e261b79c7f5f3b53243be2f3b26f9189c7b8cbf75> linux Alpine Linux 3.18.4
<Target /var/lib/docker/image/overlay2/layerdb/mounts/53ad2edfc7474c3122e601b9f23fca705eae85b405c7c52b9b53d400618a9bd4> linux Debian GNU/Linux 12 (bookworm)
<Target /var/lib/docker/image/overlay2/layerdb/mounts/96171c133b33b6db82fe59ca91d45a5dc2803bba19501f00e06d243514da1f76> linux CentOS Linux 8
```

Commands such as `target-shell`'s `enter` also work when provided the path to the overlay2 mount folder.

Uses the `LayerFilesystem` functionality of #575 to map every overlay2 layer on top of each other in a single `Overlay2Filesystem`. Deleted or moved files will still exist in the final reconstructed filesystem.


Horofic (commented):
Sorry for the delayed response. Internally we are looking at creating a separate `LayerFilesystem` (name pending :)) that can be used independent of the `RootFilesystem`. Which is a different direction of what this PR is trying to achieve, since it builds on top of the `RootFilesystem`.
 
The benefit of having a separate `LayerFilesystem` is that it would give us the ability to more cleanly map files to different paths in loaders, instead of how it is done currently. 
 
In order to accomplish this, we need to do some refactoring of the current `VirtualFilesystem` implementation, too able to use layers. After that we are going to look and see if that approach is what we want.
 
With that said, I think the best course of action would be to keep this PR open (unmerged). So that improvements can still be made and be used by other users, until we decided on how to move forward with this.

JSCU-CNI:
Thanks for the consideration @Horofic. In the meantime, could you review the new log parsing functions in this PR? I can split the children functionality in a separate PR if needed.

Schamper:
Can you split that into a different PR, please? That would aid the review, and likely get that merged in sooner!

JSCU-CNI:
@Schamper. Done, see #507.

Schamper:
In light of #575 it might be nice to base this on top of that. Additionally, I think any logic that is currently in the loader should be transferred to a `overlay2` filesystem implementation. The logic for it should be contained within the filesystem implementation as much as possible.

Other thought I had was if our current "overlay filesystem" behaves 1:1 with overlay2? Is there no additional logic necessary to deal with edge cases from overlay2? Something like deleted files, for example. I'm not sure if I ever wrote this on GitHub before, but we plan to migrate to a "DirectoryEntry" kind of approach in the future, where concepts like deleted files can more easily be integrated.

Schamper:
Yeah don't worry about the deleted files - that was more a random thought, I didn't expect it to be supported right away. My question was mostly if your implementation behaves exactly as `overlay2` in most other regards. For example, in the loader there's currently a note that says "Older files are 'overwritten' by newer versions.". It reads as if that's a warning/unintended side-effect, but is that not the correct behaviour?

As for implementation, I had something like this in mind:
- `Overlay2Loader`
  - We need this because filesystem detection only supports file-like objects, but this should primarily be just detection and then a map that basically boils down to `target.filesystems.add(Overlay2Filesystem(self.path))`
- `Overlay2Filesystem`
  - Subclass of `LayerFilesystem` that accepts a `Path` object as argument (similar to `DirFilesystem`)
  - In the `__init__`, you do the logic of traversing the parents and adding the layers (`self.add_fs_layer`)
  - Can the actual layers just be a straight up `DirectoryFilesystem`? I don't think you actually need a `Overlay2LayerFilesystem` (and `...Entry` class)

Now I (again) have to admit that I have not fully read up yet on overlay2. But my reasoning here is that with the above layout, it's easier to re-use the `Overlay2Filesystem` as a standalone filesystem. E.g.:

```python
>>> from pathlib import Path
>>> from dissect.target.filesystems.overlay2 import Overlay2Filesystem
>>> fs = Overlay2Filesystem(Path("/imagine/that/this/is/a/valid/path/to/a//overlay2/root"))
>>> fs.listdir("/")
```

So, having not actually read up on the details, I hope my assumptions hold up and this is actually a good idea. Inb4 I get smack-downed and thrown the book at me.

Schamper (changes requested):

Schamper:
> Only issue I ran into was a Python 3.7 `os.readlink` call in `DirectoryFilesystemEntry`. I assume we can now safely use `self.entry.readlink()` there, but please correct me if I'm wrong.

Yup, we only target >= Python 3.9 now, so that was fine to remove!

dissect/target/filesystems/overlay.py
Schamper, line 1: Can you rename this to `overlay2.py`?

Schamper:
Can you rebase on top of main?

dissect/target/filesystems/overlay.py
```diff
@@ -0,0 +1,86 @@
...
+            for mount in config.get("MountPoints").values():
+                if not mount["Type"] in ["volume", "bind"]:
+                    path._fs.target.log.warning(
+                        "Encountered unsupported mount type %s in container %s", mount["Type"], path.name
+                    )
+                    continue
+
+                if not mount["Source"] and mount["Name"]:
+                    # anonymous volumes do not have a Source but a volume id
+                    layer_fs = DirectoryFilesystem(root / "volumes" / mount["Name"] / "_data")
```
Schamper, line 75:
```suggestion
                    layer_fs = DirectoryFilesystem(root .joinpath("volumes", mount["Name"], "_data"))
```

dissect/target/plugins/child/docker.py
```diff
@@ -0,0 +1,30 @@
...
+    """Child target plugin that yields from Docker overlay2fs containers."""
+
+    __type__ = "docker"
+
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self.data_roots = set(find_installs(target))
+
+    def check_compatible(self) -> None:
+        if not any(self.data_roots):
```
Schamper, line 18:
```suggestion
        if not self.data_roots:
```

dissect/target/plugins/child/docker.py
```diff
@@ -0,0 +1,30 @@
...
+            raise UnsupportedPluginError("No Docker data root folder(s) found!")
+
+    def list_children(self):
+        for data_root in self.data_roots:
+            mount_folder = data_root.joinpath("image/overlay2/layerdb/mounts")
+
+            for container in self.target.docker.containers():
```
Schamper, line 25: `docker.containers()` also iterates over the installs, wouldn't you generate invalid child records with containers that don't exist in that install (aside from generating an exponential amount)? Perhaps you need to add a field to the `DockerContainerRecord` to fix this cleanly.

tests/loaders/test_overlay.py
```diff
@@ -0,0 +1,22 @@
...
+from dissect.target.filesystem import VirtualFilesystem
+from dissect.target.loaders.overlay import Overlay2Loader
+from tests.plugins.apps.container.test_docker import (  # noqa: F401
+    fs_docker,
+    target_linux_docker,
+)
+
+
+def test_overlay_loader_docker_container(
+    target_linux_docker: Target, fs_docker: VirtualFilesystem  # noqa: F811
```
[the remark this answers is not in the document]
Schamper: I took a closer look, you're using pytest fixtures not quite how they're supposed to be used. You don't have to manually import them, pytest will insert the correct fixture by name of the argument. You also don't directly use the `fs_docker` fixture in your test, so there's no need to reference it, it's already implicitly included from `target_linux_docker`. Only if you need to interact with the fixture in your own test should you include it.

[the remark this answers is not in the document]
Schamper: Ah I missed that, apologies. I didn't look close enough ;) Let me come back to you on this.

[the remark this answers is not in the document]
Schamper: In the previous situation, the fixture was only used in that test file. In that case, it's indeed recommended to only place it in that file. However, since you're now using the fixture in different test files, it's better to move it to the global `conftest.py`, rather than doing an importing dance. `conftest.py` is also directory scoped, but unfortunately that doesn't help us in this case.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -191,6 +192,7 @@ def containers(self) -> Iterator[DockerContainerRecord]:
                 yield DockerContainerRecord(
                     container_id=container_id,
                     image=config.get("Config").get("Image"),
+                    image_id=config.get("Image").replace("sha256:", ""),
```
Schamper, line 195: Would a split on `:` not be a little more future proof?

Schamper (changes requested):

Schamper (changes requested):
Can you check why the tests are failing? I think it may have to do with the change in `dir.py`.

Schamper:
> Perhaps we need an `Overlay2FilesystemEntry` subclassed from `DirectoryFilesystemEntry` after all :)

Unfortunately that doesn't solve the underlying problem that `DirectoryFilesystem` should work on both `pathlib.Path` and `TargetPath`.  I've made it a bit more explicit right now.

Schamper:
Nice, that was indeed it. I've backported a fix for 3.9.

Schamper (dismissed):

PR #442, opened 2023-11-13
Improve support for alpine linux

JSCU-CNI:
This PR adds improved support for os version detection on alpine linux systems.

### DIFF
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -230,9 +230,6 @@ class UnixPlugin(OSPlugin):
         os_release = {}
 
         for path in self.target.fs.glob(glob):
-            if len(os_release) > 1:
-                break
-
             if self.target.fs.path(path).is_file():
                 with self.target.fs.path(path).open("rt") as release_file:
                     for line in release_file:
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -230,6 +230,9 @@ class UnixPlugin(OSPlugin):
         os_release = {}
 
         for path in self.target.fs.glob(glob):
+            if len(os_release) > 1:
+                break
+
             if self.target.fs.path(path).is_file():
                 with self.target.fs.path(path).open("rt") as release_file:
                     for line in release_file:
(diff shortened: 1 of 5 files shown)


dissect/target/plugins/os/unix/_os.py
```diff
@@ -230,6 +230,9 @@ def _parse_os_release(self, glob: Optional[str] = None) -> dict[str, str]:
         os_release = {}
 
         for path in self.target.fs.glob(glob):
+            if len(os_release) > 1:
+                break
+
```
pyrco, line 235: I don't think this is meant to be? The idea of this function seems to be to parse all available information. With this change it depends on which file is returned first by the filesystem which information will be present in `os_release`.

pyrco (changes requested):

dissect/target/plugins/os/unix/_os.py
JSCU-CNI: Yes this is intentional. We want to prevent only returning information from a sparse release file, if we also have a better os-release file.

Consider the following filesystem (`docker run -it alpine:latest`):
```
/ # cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.18.4
PRETTY_NAME="Alpine Linux v3.18"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"

/ # cat /etc/alpine-release 
3.18.4
```

When the parser would eventually encounter the `/etc/alpine-release` file (either first or last, does not matter), the following would be evaluated as true:

```python
# _parse_os_release @ dissect/target/plugins/os/unix/_os.py
elif "=" not in line:
    os_release["DISTRIB_DESCRIPTION"] = line.strip()
```

Which would cause the following to run:

```python
# version @ dissect/target/plugins/os/unix/linux/_os.py
if distrib_description := self._os_release.get("DISTRIB_DESCRIPTION"):
    return distrib_description
```

Which in turn would cause the output of the `version` function to be `3.18.4` instead of `Alpine Linux 3.18.4`. 

By adding `if len(os_release) > 1: break` we prevent the parsing of the `/etc/alpine-release` file if the `/etc/os-release` was already parsed. The length of `1` makes sure that the `/etc/os-release` file would still be parsed if the `/etc/alpine-release` file was parsed first.

PR #443, opened 2023-11-14
Fix handling of volumes at offset 0

Schamper:
As part of this fix I've introduced a `__type__` attribute to the `VolumeSystem` class. It's purpose is similar to `Filesystem.__fstype__` - being able to do some sort of type-check if you can't do a `isinstance()` check due to lazy importing.

In fact, I probably want to add it to `Container` too, and perhaps rename `Filesystem.__fstype__` to `Filesystem.__type__` - of course with the necessary deprecation warning and period.

### DIFF
--- a/dissect/target/volume.py
+++ b/dissect/target/volume.py
@@ -57,6 +57,8 @@ class VolumeSystem:
         serial: Serial number of the volume system, if any.
     """
 
+    # Due to lazy importing we generally can't use isinstance(), so we add a short identifying string to each class
+    # This has the added benefit of having a readily available "pretty name" for each implementation
     __type__: str = None
     """A short string identifying the type of volume system."""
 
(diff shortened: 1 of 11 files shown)


dissect/target/volume.py
```diff
@@ -57,6 +57,9 @@ class VolumeSystem:
         serial: Serial number of the volume system, if any.
     """
 
+    __type__: str = None
+    """A short string identifying the type of volume system."""
```
pyrco, line 61: Would it be an idea to add a comment (or extend the docstring) with some usecases for using `__type__` instead of `isinstance()`?

pyrco (changes requested):

dissect/target/volume.py
Schamper: Added one

pyrco (approved):

PR #444, opened 2023-11-14
Unify __type__ on filesystems and containers

Schamper:
Depends on #443 for doing the same with volume systems.

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -76,7 +76,8 @@ class Filesystem:
     @property
     def __fstype__(cls) -> str:
         warnings.warn(
-            "The __fstype__ attribute is deprecated and will be removed in dissect.target 3.15. Use __type__ instead"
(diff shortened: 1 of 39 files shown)


dissect/target/filesystem.py
```diff
@@ -57,17 +60,26 @@ def __init__(
...
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}>"
 
+    @classmethod
+    @property
+    def __fstype__(cls) -> str:
+        warnings.warn(
+            "The __fstype__ attribute is deprecated and will be removed in dissect.target 3.15. Use __type__ instead"
+        )
```
pyrco, line 80:
```suggestion
        warnings.warn(
            "The __fstype__ attribute is deprecated and will be removed in dissect.target 3.15. Use __type__ instead",
            DeprecationWarning,
        )
```

pyrco (changes requested):

pyrco (approved):

PR #447, opened 2023-11-17
Put plugins in the proper unix or linux namespace

pyrco:
(DIS-1996)


dissect/target/plugins/os/unix/linux/packagemanager.py
Schamper, line 1: This doesn't necessarily has to be moved, has it? BSD-likes also have a package manager? We don't have any parsers for those currently, but that doesn't mean we won't want to add them to here.

Schamper (changes requested):

dissect/target/plugins/os/unix/linux/packagemanager.py
pyrco: This plugin is for installation logs only, the pacakge manager specific database plugins are in the various subdirectories. OpenBSD does not log any package installation actions and FreeBSD logs to syslog (it doesn't have a default log file for package install logs only possibly site specific through syslog.conf), which means most install logs are probably rotated away and cleaned up. So support for the BSDs felt rather hypothetically.But if you think the FreeBSD case should be supported, Ill move it back.

Schamper: Ah, that's a fair comment. However, I still think it belongs in Unix because we might add package managers down the line that aren't exclusively Linux.

Speaking of, I don't think `esxi` is actually a Linux derivative. Would you also move that with this PR?

dissect/target/plugins/os/unix/packagemanager.py
```diff
@@ -77,5 +77,5 @@ def _func(self, f: str) -> Iterator[PackageManagerLogRecord]:
 
     @export(record=PackageManagerLogRecord)
     def logs(self) -> Iterator[PackageManagerLogRecord]:
-        """Returns logs from apt, yum and zypper package managers."""
+        """Returns logs from all available unix package managers."""
```
Schamper, line 80:
```suggestion
        """Returns logs from all available Unix package managers."""
```

Schamper (approved):

PR #453, opened 2023-11-21
Initial cookies implementation for Firefox and Chromium

YoeriNijs:
Initial cookies support for Firefox and Chromium. Missing iexplore since I do not run Windows atm. Unit tests may follow when I have some time remaining.


YoeriNijs:
@DissectBot agree [company="default"]

Horofic:
Thank you for your contribution. I will be doing the review in the coming days. So stay tuned!

YoeriNijs:
@Horofic Thanks! I also refactored some tests today in order to obey the boyscout rule. I think we can increase the test quality by checking the record values as well, but this is future work ;)

tests/plugins/apps/browser/test_iexplore_plugin.py
```diff
@@ -0,0 +1,51 @@
...
+
+    records = list(target_win.iexplore.downloads())
+    assert len(records) == 1
+    assert records[0].path == "C:\\Users\\John\\Downloads\\archlinux-2023.02.01-x86_64.iso"
+    assert records[0].url == "https://mirror.cj2.nl/archlinux/iso/2023.02.01/archlinux-2023.02.01-x86_64.iso"
+
+    records = list(target_win.browser.downloads())
+    assert len(records) == 1
+    assert records[0].path == "C:\\Users\\John\\Downloads\\archlinux-2023.02.01-x86_64.iso"
+    assert records[0].url == "https://mirror.cj2.nl/archlinux/iso/2023.02.01/archlinux-2023.02.01-x86_64.iso"
```
Horofic, line 29:
```suggestion
```

tests/plugins/apps/browser/test_chromium_plugin.py
```diff
@@ -0,0 +1,52 @@
+from dissect.target.plugins.apps.browser import chromium
+from tests._utils import absolute_path
+
+
+def test_chromium_downloads(target_win, fs_win, tmp_path, target_win_users):
```
Horofic, line 5: Please add typehints to the tests as well.

tests/plugins/apps/browser/test_iexplore_plugin.py
```diff
@@ -0,0 +1,51 @@
...
+def test_iexplore_downloads(target_win, fs_win, tmp_path, target_win_users):
+    __setup(target_win, fs_win, tmp_path, target_win_users)
+
+    records = list(target_win.iexplore.downloads())
+    assert len(records) == 1
+    assert records[0].path == "C:\\Users\\John\\Downloads\\archlinux-2023.02.01-x86_64.iso"
+    assert records[0].url == "https://mirror.cj2.nl/archlinux/iso/2023.02.01/archlinux-2023.02.01-x86_64.iso"
+
+    records = list(target_win.browser.downloads())
+    assert len(records) == 1
```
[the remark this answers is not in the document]
YoeriNijs: Not sure since it is possible that there a more than 1 downloads. We do not check on this later, so it seems to me that this check is useful.

tests/plugins/apps/browser/test_chrome_plugin.py
```diff
@@ -0,0 +1,34 @@
+from dissect.target.plugins.apps.browser import chrome
+from tests._utils import absolute_path
+
+
+def test_chrome(target_win, fs_win, tmp_path, target_win_users):
+    chrome_db = absolute_path("_data/plugins/apps/browser/chrome/History.sqlite")
```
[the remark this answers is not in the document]
YoeriNijs: Personally, I don't like the use of fixtures in this way. Yes, it cleans up the test, but it may also be very confusing how the test is constructed (for me, it was). Hence, I would leave the test logic all together.

I you prefer to move this setup logic nonetheless, I think you can do two things:
- create a fixture file that belongs only to the browsers tests, and don't add the fixture in one god file that holds all fixtures. Place this fixture file in the same directory as the tests, and eventually add a prefix to indictate it is a fixture (e.g. `fixture_chrome.py`);
- refactor the tests so we can use the native setUp hook: https://docs.python.org/3/library/unittest.html. I believe this is out of scope for this pr, however.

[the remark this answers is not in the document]
Horofic: I think your second point is indeed out of scope for this PR as it would require significant refactoring, since you are essentially switching from one testing framework to another.

I do get your first point, so i'd like to propose moving the contents of `__setup()` to a `@pytest.fixture` decorated function within the respective test file itself. This avoids putting more, not globally used fixtures in `conftest.py`, is in line with how we currently handle these specifics, arguably as readable as the current `__setup()` way and keeps the test itself clean. 

To accomplish this, you could use one of the examples above or take some inspiration from other test files containing fixtures.

dissect/target/plugins/apps/browser/chrome.py
```diff
@@ -29,6 +30,10 @@ class ChromePlugin(ChromiumMixin, BrowserPlugin):
         # Macos
         "Library/Application Support/Google/Chrome/Default",
     ]
+    BrowserCookieRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/edge/cookie",
```
Horofic, line 34:
```suggestion
        "browser/chrome/cookie",
```

dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -170,6 +174,46 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
             except SQLError as e:
                 self.target.log.warning("Error processing history file: %s", db_file, exc_info=e)
 
+    def cookies(self, browser_name: str = None) -> Iterator[BrowserCookieRecord]:
+        """Return browser cookies records from supported Chromium-based browsers.
```
Horofic, line 178:
```suggestion
        """Return browser cookie records from supported Chromium-based browsers.
```

dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -170,6 +174,46 @@ def downloads(self, browser_name: str = None) -> Iterator[BrowserDownloadRecord]
...
+        Args:
+            browser_name: The name of the browser as a string.
+        Yields:
+            Records with the following fields:
+                ts_created (datetime): Download start timestamp.
+                ts_last_accessed (datetime): Download end timestamp.
```
Horofic, line 184: Is this the correct description? Should this not be related to the creation and last access of the cookie?

dissect/target/plugins/apps/browser/firefox.py
```diff
@@ -74,6 +78,47 @@ def _iter_db(self, filename: str) -> Iterator[SQLite3]:
...
 
+    @export(record=BrowserCookieRecord)
+    def cookies(self) -> Iterator[BrowserCookieRecord]:
+        """Return browser cookies records from Firefox.
+        Args:
+            browser_name: The name of the browser as a string.
+        Yields:
+            Records with the following fields:
+                ts_created (datetime): Download start timestamp.
+                ts_last_accessed (datetime): Download end timestamp.
```
Horofic, line 89: Same here. Is this the correct description? Should this not be related to the creation and last access of the cookie?

dissect/target/plugins/apps/browser/iexplore.py
```diff
@@ -69,6 +69,9 @@ def downloads(self) -> Iterator[record.Record]:
         """Yield records from the iedownload webcache container."""
         yield from self._iter_records("iedownload")
 
+    def cookies(self) -> Iterator[record.Record]:
+        raise NotImplementedError("Cookies function is not implemented for Internet Explorer yet")
```
Horofic, line 73:
```suggestion
    def cookies(self) -> None:
        raise NotImplementedError("Cookies plugin is not implemented for Internet Explorer yet")
```

Horofic (changes requested):

Schamper (changes requested):
I've also pushed some small style changes, and fixed the tests!

tests/_data/plugins/apps/browser/chromium/Cookies.sqlite
Schamper, line 1: These files are added as LFS files in `.gitattributes` but not actually committed/uploaded as LFS files. Could you add those as LFS?

Schamper (changes requested):

tests/_data/plugins/apps/browser/chromium/Cookies.sqlite
YoeriNijs: Any idea how I am actually doing this? Branch is up to date atm, according to Git. I've also added the Cookies file to Git already.

Schamper: It's a bit tricky, and unfortunately I can't do it from my end since I can't interact with the LFS on your fork. We're now in a situation where the `.gitattributes` file says that the file should be in LFS but the file is actually included directly in the repo.

However, what you could try to do it to go back to a state where there are "no files", and then add them as LFS files:
- First off, ensure that you have `git-lfs` installed. Steps may vary depending on your operating system: https://git-lfs.com/
- Remove the paths from `.gitattributes`
- Delete the file from git with `git rm --cached tests/_data/plugins/apps/browser/chromium/Cookies.sqlite`
- Track the file in LFS with `git lfs track tests/_data/plugins/apps/browser/chromium/Cookies.sqlite`
- Stage both the `.gitattributes` and the test file
- Commit and push

You may need to commit in between these steps to get everything to play nicely. The most important thing is probably to have `git-lfs` installed. Maybe doing that and `git rm`'ing and re-adding the file is already enough. You'll have to play around with that a little bit.

Schamper:
image

Nice! Can you also do it for the Firefox file?

Schamper (approved):

Horofic (approved):

PR #457, opened 2023-11-24
Improve hash functionality

Miauwkeru:
(DIS-2623)

### DIFF
--- a/dissect/target/helpers/hashutil.py
+++ b/dissect/target/helpers/hashutil.py
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
 BUFFER_SIZE = 32768
 
 RECORD_NAME = "filesystem/file/digest"
-NAME_SUFFIXES = ["_resolved_path", "_digest"]
+NAME_SUFFIXES = ["_resolved", "_digest"]
 RECORD_TYPES = ["path", "digest"]
 
 
--- a/tests/helpers/test_hashutil.py
+++ b/tests/helpers/test_hashutil.py
@@ -70,7 +70,7 @@ def test_hash_path_records_with_paths(
     _record = hashed_record.records[1]
 
     for name, _record in zip(record_names, hashed_record.records[1:]):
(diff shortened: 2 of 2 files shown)


dissect/target/helpers/hashutil.py
```diff
@@ -15,13 +15,9 @@
...
-HashRecord = RecordDescriptor(
-    "filesystem/file/digest",
-    [
-        ("path[]", "paths"),
-        ("digest[]", "digests"),
-    ],
-)
+RECORD_NAME = "filesystem/file/digest"
+NAME_SUFFIXES = ["_path", "_resolved_path", "_digest"]
+RECORD_TYPES = ["path", "path", "digest"]
```
pyrco, line 20:
```suggestion
NAME_SUFFIXES = ["_resolved", "_digest"]
RECORD_TYPES = ["path", "digest"]
```
I don't think we need to duplicate the original path value. As we prepend the field names with the original fieldname, it is easy to see what the original path would have been.

pyrco (changes requested):

JSCU-CNI:
Thanks for the heads-up @Miauwkeru.

dissect/target/helpers/hashutil.py
```diff
@@ -15,13 +15,9 @@
...
 
-HashRecord = RecordDescriptor(
-    "filesystem/file/digest",
-    [
-        ("path[]", "paths"),
-        ("digest[]", "digests"),
-    ],
-)
+RECORD_NAME = "filesystem/file/digest"
+NAME_SUFFIXES = ["_resolved_path", "_digest"]
```
pyrco, line 19:
```suggestion
NAME_SUFFIXES = ["_resolved", "_digest"]
```

I'd still rename this, as a lot of filed names are or end with `"path"`, you'd get `path_resolved_path`, which doesn't look nice.

pyrco (changes requested):

pyrco (approved):

PR #460, opened 2023-11-27
Add systemd support to the config_parser utility

Miauwkeru:
- Force parsed_data to always be a dictionary
- Improve API for config_tree functionality

### DIFF
--- a/dissect/target/filesystems/config.py
+++ b/dissect/target/filesystems/config.py
@@ -276,7 +276,8 @@ class ConfigurationEntry(FilesystemEntry):
         return False
 
     def exists(self, key: str) -> bool:
-        """Return whether the underlying :class:`.FilesystemEntry` :attr:`entry` and supplied ``key`` exists inside this :class:`ConfigurationEntry`.
+        """Return whether the underlying :class:`.FilesystemEntry` :attr:`entry` and
+        supplied ``key`` exists inside this :class:`ConfigurationEntry`.
 
         Returns:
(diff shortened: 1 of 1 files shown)


Horofic (changes requested):

Horofic (changes requested):
I'm still torn on whether going the "Filesystem" route is the right course. A lot of functions act like a that of a file system but have some side-effect or caveat you end up needing to explain in your documentation. 

This also makes using the API not that intuitive as you would think, because you need to account for these specifics.

Going more into the route what the Registry is currently doing might  make the API more clear. Thoughts @Miauwkeru, @Schamper?

dissect/target/helpers/configutil.py
```diff
@@ -357,17 +558,18 @@ def parse(path: Union[FilesystemEntry, TargetPath], hint: Optional[str] = None,
     """Parses the content of an ``path`` or ``entry`` to a dictionary.
 
     Args:
-        file_path: An entry or targetpath that with contents to parse
-        hint: A hint to what parser should be used.
-        collapse:
-        seperator: What seperator to use for key value mapping
-        comment_prefixes: The characters that determine a comment.
+        path: The path to either a directory or file.
+        hint: What kind of parser it should use.
```
Horofic, line 562:
```suggestion
        hint: What kind of parser should be used.
```

Horofic (changes requested):

Horofic (dismissed):

PR #463, opened 2023-11-28
Fix NamespacePlugin function filtering

MaxGroot:
While working on https://github.com/fox-it/dissect.target/pull/384, I ran into some issues with the `NamespacePlugin`.

- The previous check for subplugin export functions was too strict. It did not account for subplugins reusing export methods of subplugins. 
- Also improved handling `UnsupportedPlugin` exceptions, in favour of  handling any `Exception` with a `continue`.
- Lastly, added a check for when a subplugin has an export function with the same name as the `NamespacePlugin` subclass to which it belongs.
- Added tests for the scenarios outlined above.


MaxGroot:
@DissectBot agree

dissect/target/plugin.py
```diff
@@ -978,10 +985,12 @@ def aggregator(self) -> Iterator[Record]:
...
                         try:
                             subplugin = getattr(self.target, entry)
-                            for item in getattr(subplugin, method_name)():
-                                yield item
-                        except Exception:
+                            yield from getattr(subplugin, method_name)()
+                        except UnsupportedPluginError:
                             continue
+                        except Exception as e:
+                            self.target.log.error("Subplugin %s raised an exception for %s", entry, method_name)
```
cecinestpasunepipe, line 992:
```suggestion
                            self.target.log.error("Subplugin: %s raised an exception for: %s", entry, method_name)
```

dissect/target/plugin.py
```diff
@@ -951,6 +951,10 @@ def __init_subclass_subplugin__(cls, **kwargs):
         # the direct subclass of NamespacePlugin
         cls.__nsplugin__.SUBPLUGINS.add(cls.__namespace__)
 
+        # Generate a tuple of class names for which we do not want to add subplugin functions, which is the
+        # namespaceplugin and all of its superclasses.
```
cecinestpasunepipe, line 955:
```suggestion
        # namespaceplugin and all of its superclasses (minus the base object).
```

cecinestpasunepipe (changes requested):

dissect/target/plugin.py
```diff
@@ -978,10 +985,12 @@ def aggregator(self) -> Iterator[Record]:
...
                             subplugin = getattr(self.target, entry)
-                            for item in getattr(subplugin, method_name)():
-                                yield item
-                        except Exception:
+                            yield from getattr(subplugin, method_name)()
+                        except UnsupportedPluginError:
                             continue
+                        except Exception as e:
+                            self.target.log.error("Subplugin %s raised an exception for %s", entry, method_name)
+                            self.target.log.debug("Exception: %s", e, exc_info=True)
```
Schamper, line 993:
```suggestion
                            self.target.log.debug("Exception: %s", e, exc_info=e)
```

Schamper (changes requested):

Schamper (approved):

PR #467, opened 2023-12-01
Fix most tests for windows computers

Miauwkeru:
- Check whether effective_ids is enabled for the system
- Use __type__ consistently
- Open files as UTF-8 by default
- Fix paths in test_target.py
- Change all occurrences to flow.record path to self.target.fs.path
- Fix issues regarding the resolver in trusteddocs

### DIFF
--- a/dissect/target/plugins/os/windows/startupinfo.py
+++ b/dissect/target/plugins/os/windows/startupinfo.py
@@ -1,7 +1,6 @@
 import datetime
 
 from defusedxml import ElementTree
-from flow.record.fieldtypes import path
 
 from dissect.target.exceptions import UnsupportedPluginError
 from dissect.target.helpers.record import TargetRecordDescriptor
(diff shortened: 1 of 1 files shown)


dissect/target/helpers/cache.py
```diff
@@ -142,9 +142,11 @@ def call(self, *args, **kwargs):
         elif REWRITE_CACHE:
             read_file_cache = False
 
+        effective_ids = True if os.supports_effective_ids else False
```
Schamper, line 145: Can't you just put `effective_ids=os.supports_effective_ids` in all the calls?

dissect/target/plugins/os/unix/linux/nixos/_os.py
Schamper, line 1: Accidentally added or intentional addition? Missing a `__init__.py` if the latter case.

dissect/target/plugins/os/windows/regf/trusteddocs.py
```diff
@@ -73,7 +73,7 @@ def trusteddocs(self) -> Iterator[TrustedDocumentsRecord]:
                     ts=key.ts,
                     type=value.type,
                     application=application,
-                    document_path=self.target.resolve(value.name),
+                    document_path=self.target.fs.path(self.target.resolve(value.name)),
```
Schamper, line 76: Perhaps `.resolve()` should return a path?

Schamper (changes requested):

dissect/target/plugins/os/windows/startupinfo.py
Poeloe, line 1: I guess you can remove the `from flow.record.fieldtypes import path` here as well, right?

Poeloe (changes requested):

Poeloe (approved):
LGTM

PR #469, opened 2023-12-06
Fix unix activity plugin

JSCU-CNI:
When a target system contains (broken) symlinks in `/var/log, the activity plugin could break. This is now fixed by checking if a file is a symlink or if it exists at all.

### DIFF
--- a/dissect/target/plugins/os/unix/generic.py
+++ b/dissect/target/plugins/os/unix/generic.py
@@ -20,6 +20,8 @@ class GenericPlugin(Plugin):
 
         last_seen = 0
         for f in var_log.iterdir():
+            if f.is_symlink() or not f.exists():
+                continue
             if f.stat().st_mtime > last_seen:
                 last_seen = f.stat().st_mtime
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/os/unix/generic.py
```diff
@@ -20,6 +20,8 @@ def activity(self) -> Optional[datetime]:
 
         last_seen = 0
         for f in var_log.iterdir():
+            if f.is_symlink() or not f.exists():
```
Schamper, line 23: Why outright skip symlinks? Would you not just want to check for the file existence?

JSCU-CNI: Correct, the `is_symlink` check is unnecessary. Removed in  [ffb3e36](https://github.com/fox-it/dissect.target/pull/469/commits/ffb3e367f2cc10ff79c2b78c1c4c30df85c9c681).

Schamper (approved):

PR #472, opened 2023-12-06
Improve sru plugin

JSCU-CNI:
Previously all sru plugin functions would throw an uncaught exception when the windows target does not contain a certain sru table. The plugin now gracefully exits with a warning when an sru table does not exist.


Schamper (changes requested):
We've also been meaning to change this plugin to use the information from the registry to dump available tables. But that's something for a different PR.

dissect/target/plugins/os/windows/sru.py
```diff
@@ -364,10 +366,13 @@ def check_compatible(self) -> None:
         if not self._sru:
             raise UnsupportedPluginError("No SRUDB found")
 
-    def read_records(self, table_name, record_type):
+    def read_records(
+        self, table_name: str, record_type: TargetRecordDescriptor
+    ) -> Iterator[TargetRecordDescriptor | None]:
```
Schamper, line 371:
```suggestion
    ) -> Iterator[TargetRecordDescriptor]:
```
It won't iterate values of `None`, it'd just be an empty iterator.

Also `TargetRecordDescriptor` is not quite correct. It's actually an iterator of all the listed record types above... :smile:. Maybe make a global type alias for a union of all of those called `SRURecord`?

Schamper (changes requested):

dissect/target/plugins/os/windows/sru.py
JSCU-CNI: Added in [a7e8895](https://github.com/fox-it/dissect.target/pull/472/commits/a7e8895d7b07e618cb6e690c44758f66346039ec).

dissect/target/plugins/os/windows/sru.py
```diff
@@ -322,7 +340,7 @@
 }
 
 
-def transform_app_id(value):
+def transform_app_id(value: bytes | str | None) -> Union[bytes, str, None]:
```
Schamper, line 343: You're mixing the newer `|` syntax with the older `Union[]` syntax here. I think it's best to stick with `Union[]` for now. We might migrate entirely to the `|` syntax at some point.

Schamper (changes requested):

dissect/target/plugins/os/windows/sru.py
```diff
@@ -322,7 +340,7 @@
 }
 
 
-def transform_app_id(value):
+def transform_app_id(value: Union[bytes, str, None]) -> Union[str, None]:
```
Schamper, line 343:
```suggestion
def transform_app_id(value: Optional[Union[bytes, str]]) -> Optional[str]:
```

Schamper (changes requested):

dissect/target/plugins/os/windows/sru.py
```diff
@@ -223,6 +225,22 @@
     ],
 )
 
+SRURecord: TypeAlias = Union[
```
Schamper, line 228:
```suggestion
SRURecord = Union[
```
`TypeAlias` unfortunately requires >=3.10, so this is fine IMO.

Schamper (changes requested):

dissect/target/plugins/os/windows/sru.py
JSCU-CNI: Removed in [71cb7ce](https://github.com/fox-it/dissect.target/pull/472/commits/71cb7ce665adfff82ebdcd0bf606325fbb73b718)

dissect/target/plugins/os/windows/sru.py
JSCU-CNI: Fixed in [71cb7ce](https://github.com/fox-it/dissect.target/pull/472/commits/71cb7ce665adfff82ebdcd0bf606325fbb73b718)

Schamper (changes requested):
Can you fix the unit tests?

tests/plugins/os/windows/test_sru.py
```diff
@@ -1,10 +1,14 @@
...
 import pytest
 
+from dissect.target.filesystem import VirtualFilesystem
 from dissect.target.plugins.os.windows import sru
+from dissect.target.target import Target
 from tests._utils import absolute_path
 
 
-def test_sru_plugin(target_win, fs_win):
+def test_sru_plugin(target_win: Target, fs_win: VirtualFilesystem, caplog: pytest.LogCaptureFixture):
```
Schamper, line 11:
```suggestion
def test_sru_plugin(target_win: Target, fs_win: VirtualFilesystem, caplog: pytest.LogCaptureFixture) -> None:
```

Schamper (changes requested):

Schamper:
Looks like a different unit test doesn't like the year rollover :upside_down_face: unfortunately we'll need to fix that first. The yum plugin uses the modification timestamp for the year rollover, but the CI probably changes the modification timestamp upon checkout.

.gitignore
```diff
@@ -9,3 +9,4 @@ __pycache__/
 tests/_docs/api
 tests/_docs/build
 .tox/
+.venv/
```
Schamper, line 12:
```suggestion
```
I assume this is your own venv and not tool created? This would best be placed in a personal .gitignore file.

Schamper (changes requested):

.gitignore
```diff
@@ -8,4 +8,4 @@ __pycache__/
 .pytest_cache/
 tests/_docs/api
 tests/_docs/build
-.tox/
+.tox/
```
Schamper, line 11: This now removes the final newline. Best to keep this file untouched.

Schamper (changes requested):

Schamper (approved):

PR #474, opened 2023-12-06
Fix for logloader

JSCU-CNI:
The patch below fixes the `LogLoader`. This should now work (again): 

```
target-query -f evtx log:///absolute/path/to/*.evtx

### DIFF
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -488,6 +488,10 @@ class Target:
         counter = 0
         path = "/$fs$/fs0"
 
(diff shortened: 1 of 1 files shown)


Schamper:
@Miauwkeru @cecinestpasunepipe can you comment? Why is `self.fs` ever anything other than `RootFilesystem`?

cecinestpasunepipe:
maybe it should not be, good point, have to look into that

Schamper:
#480 has been merged which should fix the originating problem for this bug, so I'm closing this one!

PR #477, opened 2023-12-09
Added sid function for windows OS plugin

fox-evv:
It returns the machine SID found in the SAM hive.


pyrco:
Hi @fox-evv , nice addition. This function should go in the os/windows/generic.py however. OSPlugins are reserved for only the methods defined in its base class.

Schamper:
Superseded by #824

PR #481, opened 2023-12-12
Only allow RootFileSystem to be assigned to target.fs

cecinestpasunepipe:


Schamper (commented):
Isn't this a little bit overkill? This will only cause unnecessary overhead for something that _should_ be caught in a code review. It wasn't this time, but I don't think it's necessary to handicap every code path that uses the filesystem for that. Now we know and can be more aware of it.

cecinestpasunepipe:
> Isn't this a little bit overkill? This will only cause unnecessary overhead for something that _should_ be caught in a code review. It wasn't this time, but I don't think it's necessary to handicap every code path that uses the filesystem for that. Now we know and can be more aware of it.
> 
> If you still want to keep the check, a more lightweight approach might be to check if `self.fs` is still a `RootFilesystem` at the end of `apply()` (or another suitable location near the end of initialization).

Ok, let's reject it then.

PR #482, opened 2023-12-12
Add support for birth timestamps in extfs / walkfs

syzzer:
The walkfs plugin would always report btime=None, which is a shame since ext4 does support birth timestamps (called crtime in ext lingo).

dissect.extfs already supported crtime, so a simple fix suffices.

### DIFF
--- a/dissect/target/filesystems/extfs.py
+++ b/dissect/target/filesystems/extfs.py
@@ -125,6 +125,10 @@ class ExtFilesystemEntry(FilesystemEntry):
             ]
         )
 
+        # Set birthtime if available
+        if self.entry.crtime:
+            st_info.st_birthtime = self.entry.crtime.timestamp()
+
         # Set the nanosecond resolution separately
         st_info.st_atime_ns = self.entry.atime_ns
         st_info.st_mtime_ns = self.entry.mtime_ns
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/filesystem/walkfs.py
```diff
@@ -57,6 +57,7 @@ def generate_record(target: Target, path: TargetPath) -> FilesystemRecord:
         atime=from_unix(stat.st_atime),
         mtime=from_unix(stat.st_mtime),
         ctime=from_unix(stat.st_ctime),
+        btime=from_unix(stat.st_birthtime),
```
pyrco, line 60: stat.st_birthtime could be `None` which makes `from_unix()` fail, so it should be guarded.

pyrco (changes requested):

dissect/target/filesystems/extfs.py
```diff
@@ -125,7 +125,8 @@ def lstat(self) -> fsutil.stat_result:
             ]
         )
 
-        # Set the nanosecond resolution separately
+        # Set birthtime and nanosecond resolution separately
+        st_info.st_birthtime = self.entry.crtime.timestamp()
```
Schamper, line 129: `self.entry.crtime` can also be `None` and should also be guarded :smile:.

dissect/target/plugins/filesystem/walkfs.py
syzzer: Gah, ofc. I promise to run the tests next time.

dissect/target/filesystems/extfs.py
syzzer: Ty, fixed.

pyrco (approved):

PR #483, opened 2023-12-12
Add default docstrings to OSPlugin subclass methods

pyrco:


dissect/target/plugin.py
```diff
@@ -292,18 +292,42 @@ class OSPlugin(Plugin):
...
+            docstring = method.__doc__
+
+            if method and not docstring:
+                if hasattr(method, "__func__"):
+                    method = method.__func__
+                method.__doc__ = os_docstring
+
     def check_compatible(self) -> bool:
-        """OSPlugin's use a different compatibility check, override the default one."""
+        """OSPlugin's use a different compatibility check, override the one from the Plugin class.
```
Schamper, line 321:
```suggestion
        """OSPlugin's use a different compatibility check, override the one from the :class:`Plugin` class.
```
I don't quite remember the syntax, but I think it was like this?

dissect/target/plugin.py
```diff
@@ -292,18 +292,42 @@ class OSPlugin(Plugin):
...
+
+        Returns:
+            This function always returns ``True``."""
```
Schamper, line 324:
```suggestion
            This function always returns ``True``.
        """
```

Schamper (changes requested):

dissect/target/plugin.py
```diff
@@ -292,18 +292,43 @@ class OSPlugin(Plugin):
     This provides a base class for certain common functions of OS's, which each OS plugin has to implement separately.
 
     For example, it provides an interface for retrieving the hostname and users of a target.
+
+    All derived classes MUST implement ALL the classmethods and exported
+    methods with the same @classmethod or @export(...) annotation.
```
Schamper, line 297:
```suggestion
    methods with the same ``@classmethod`` or ``@export(...)`` annotation.
```

Schamper (changes requested):

dissect/target/plugin.py
pyrco: Changed a few more

Schamper (approved):

PR #487, opened 2023-12-20
Fix finding namespace plugins that do not want to be __findable__

pyrco:
(DIS-2901)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1128,29 +1128,45 @@ def find_plugin_functions(
     ignore_load_errors = kwargs.get("ignore_load_errors", False)
 
     for pattern in patterns.split(","):
-        # backward compatibility fix for namespace-level plugins (i.e. chrome)
+        # Backward compatibility fix for namespace-level plugins (i.e. chrome)
+        # If an exact namespace match is found, the pattern is changed to the tree to that namespace.
+        # Examples:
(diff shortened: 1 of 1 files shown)


dissect/target/plugin.py
```diff
@@ -1128,29 +1128,45 @@ def find_plugin_functions(
...
-            # Examples:
-            #     -f browsers -> browsers* (the whole package)
-            #     -f apps.webservers.iis -> apps.webservers.iis* (logs etc)
-            # We do not include  a dot because that does not work if the full path is given:
-            #     -f apps.webservers.iis.logs != apps.webservers.iis.logs.* (does not work)
+        # Change the treematch pattern into an fnmatch-able pattern to give back all functions from the sub-tree
+        # (if there is a subtree).
+        #
+        # Examples:
+        #     -f browser -> apps.browser.browsers* (the whole package, due to a namespace match)
```
cecinestpasunepipe, line 1161: -f browser doesn't work?

target-query: error: argument -f/--function contains invalid plugin(s): browser

cecinestpasunepipe (changes requested):

dissect/target/plugin.py
pyrco: Works here?

cecinestpasunepipe (approved):

PR #490, opened 2023-12-27
Fix Velociraptor Loader VSS recognized as root filesystem

Zawadidone:
Mount volumes that represent drives before the volume shadow copies, otherwise a volume shadow copy is used as root filesystem. 
https://github.com/fox-it/dissect.target/blob/8bb86ca55a7194dd79a938c42e70570740510f9a/dissect/target/plugins/os/windows/_os.py#L24-L26

### DIFF
--- a/dissect/target/loaders/velociraptor.py
+++ b/dissect/target/loaders/velociraptor.py
@@ -36,14 +36,22 @@ def find_fs_directories(path: Path) -> tuple[Optional[OperatingSystem], Optional
 
     # Windows
     volumes = set()
+    vss_volumes = set()
     for accessor in WINDOWS_ACCESSORS:
         accessor_root = fs_root.joinpath(accessor)
         if accessor_root.exists():
             # If the accessor directory exists, assume all the subdirectories are volumes
-            volumes.update(accessor_root.iterdir())
+            for volume in accessor_root.iterdir():
+                # https://github.com/Velocidex/velociraptor/blob/87368e7cc678144592a1614bb3bbd0a0f900ded9/accessors/ntfs/vss.go#L82
+                if "HarddiskVolumeShadowCopy" in volume.name:
+                    vss_volumes.add(volume)
+                else:
+                    volumes.add(volume)
 
     if volumes:
-        return OperatingSystem.WINDOWS, list(volumes)
+        # The volumes that represent drives (C, D) are mounted first,
+        # otherwise one of the volume shadow copies could be detected as the root filesystem which results in errors.
+        return OperatingSystem.WINDOWS, list(volumes) + list(vss_volumes)
 
     return None, None
 
--- a/tests/loaders/test_velociraptor.py
+++ b/tests/loaders/test_velociraptor.py
@@ -9,13 +9,34 @@ from dissect.target.loaders.velociraptor import VelociraptorLoader
 from tests._utils import absolute_path, mkdirs
 
 
-def create_paths(sub_dir: str) -> list[str]:
-    return [
+def create_root(sub_dir: str, tmp_path: Path) -> Path:
+    paths = [
         f"uploads/{sub_dir}/%5C%5C.%5CC%3A/",
         f"uploads/{sub_dir}/%5C%5C.%5CC%3A/$Extend",
         f"uploads/{sub_dir}/%5C%5C.%5CC%3A/windows/system32",
-        f"uploads/{sub_dir}/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1",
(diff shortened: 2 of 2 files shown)


tests/loaders/test_velociraptor.py
```diff
@@ -9,28 +9,23 @@
...
-def test_velociraptor_loader_windows_ntfs(sub_dir: str, target_bare: Target, tmp_path: Path) -> None:
-    paths = create_paths(sub_dir)
     root = tmp_path
     mkdirs(root, paths)
 
     (root / "uploads.json").write_bytes(b"{}")
 
-    with open(absolute_path("_data/plugins/filesystem/ntfs/mft/mft.raw"), "rb") as fh:
-        root.joinpath(paths[0]).joinpath("$MFT").write_bytes(fh.read(10 * 1025))
+    mft = open(absolute_path("_data/plugins/filesystem/ntfs/mft/mft.raw"), "rb").read(10 * 1025)
```
Schamper, line 26: Can you keep the context manager here for reading the MFT data?

dissect/target/loaders/velociraptor.py
```diff
@@ -36,14 +36,22 @@ def find_fs_directories(path: Path) -> tuple[Optional[OperatingSystem], Optional
...
+                if "HarddiskVolumeShadowCopy" in volume.name:
+                    vss_volumes.add(volume)
+                else:
+                    volumes.add(volume)
 
     if volumes:
-        return OperatingSystem.WINDOWS, list(volumes)
+        # The volumes that represent drives (C, D) are mounted first,
+        # otherwise one of the volume shadow copies could be detected as the root filesystem which results in errors.
+        return OperatingSystem.WINDOWS, list(volumes) + list(vss_volumes)
```
Schamper, line 54:
```suggestion
        return OperatingSystem.WINDOWS, list(volumes | vss_volumes)
```

tests/loaders/test_velociraptor.py
Schamper, line 1: Maybe I'm reading over it but I don't see an explicit test for this fix? I.e. asserting the correct root filesystem was chosen. Might be nice to add that.

Schamper (changes requested):

dissect/target/loaders/velociraptor.py
Zawadidone: The `volumes` should be the first items in the list using the join operator `|` cause the test to fail, because the order of the list is random.

```python

>>> first = {"C", "D"}
>>> second = {"VSS1", "VSS2"}

>>> list(first | second)
# Incorrect because the items of `second` should be added at the end of the list
['D', 'VSS1', 'C', 'VSS2'] 

>>> list(first) + list(second)
# Correct 
['D', 'C', 'VSS2', 'VSS1'] 
```

Schamper (approved):

tests/loaders/test_velociraptor.py
```diff
@@ -85,13 +78,15 @@ def test_velociraptor_loader_windows_ntfs_zip(sub_dir: str, target_bare: Target,
...
 
     usnjrnl_records = 0
     for fs in target_bare.filesystems:
         if isinstance(fs, NtfsFilesystem):
             usnjrnl_records += len(list(fs.ntfs.usnjrnl.records()))
-    assert usnjrnl_records == 1
-    assert len(target_bare.filesystems) == 3
+    assert usnjrnl_records == 2
+    assert len(target_bare.filesystems) == 4
+    assert target_bare.fs.path("sysvol/C-DRIVE.txt").exists() is True
```
Schamper, line 89:
```suggestion
    assert target_bare.fs.path("sysvol/C-DRIVE.txt").exists()
```

tests/loaders/test_velociraptor.py
```diff
@@ -39,44 +37,39 @@ def test_velociraptor_loader_windows_ntfs(sub_dir: str, target_bare: Target, tmp
...
     for fs in target_bare.filesystems:
         if isinstance(fs, NtfsFilesystem):
             usnjrnl_records += len(list(fs.ntfs.usnjrnl.records()))
-    assert usnjrnl_records == 1
-
-    # The 2 found directories + the fake NTFS filesystem
-    assert len(target_bare.filesystems) == 3
+    assert usnjrnl_records == 2
+    assert len(target_bare.filesystems) == 4
+    assert target_bare.fs.path("sysvol/C-DRIVE.txt").exists() is True
```
Schamper, line 64:
```suggestion
    assert target_bare.fs.path("sysvol/C-DRIVE.txt").exists()
```

Schamper (changes requested):

PR #492, opened 2024-01-02
Fix yum year rollover unit test

Schamper:
It previously relied on the modified timestamp of the test file, but fresh checkouts will now have a modification time in 2024. This fix mocks the modified timestamp to be in 2023.


JSCU-CNI:
Thanks for the quick fix. That was definitely an oversight from our side. We've built `map_dir_from_tar` and `map_file_from_tar` functions to help in keeping stuff like `mtime` preserved when testing (using `TarFilesystem`). Would you be interested in a PR for these functions?

Schamper:
How would those work, `map_dir_from_tar("/path/to/file.tar", "/path/to/dir")`? 

If it's not too much effort, it couldn't hurt to submit a PR for it.

PR #494, opened 2024-01-03
Refactor TargetPath and add support for Python 3.12

Schamper:
This refactors TargetPath into a structure that's hopefully a little easier to maintain for future Python versions. TargetPath implementations are now split into different files, one for each Python version. Each file has specific changes needed to make it work. Each new release of Python still requires a new patch though.

I've also fixed some small filesystem and other test related bugs I ran into to get this all working.

As I'm writing this, I'm also starting to wonder at what point it might be a better idea to just... subclass `pathlib.Path` but literally override everything. I'm sure that has it's own set of challenges though.

Depends on https://github.com/fox-it/flow.record/pull/91

### DIFF
--- a/dissect/target/helpers/compat/path_39.py
+++ b/dissect/target/helpers/compat/path_39.py
@@ -1,0 +1,545 @@
+"""A pathlib.Path compatible implementation for dissect.target.
+
+This allows for the majority of the pathlib.Path API to "just work" on dissect.target filesystems.
+
+Most of this consists of subclassed internal classes with dissect.target specific patches,
+but sometimes the change to a function is small, so the entire internal function is copied
+and only a small part changed. To ease updating this code, the order of functions, comments
+and code style is kept exactly the same as the original pathlib.py.
+
+Yes, we know, this is playing with fire and it can break on new CPython releases.
+
+The implementation is split up in multiple files, one for each CPython version.
+You're currently looking at the CPython 3.9 implementation.
+
+Commit hash we're in sync with: debb751
+"""
+
+from __future__ import annotations
+
+import fnmatch
+import re
+from pathlib import Path, PurePath, _Accessor, _PosixFlavour
+from stat import S_ISBLK, S_ISCHR, S_ISFIFO, S_ISSOCK
+from typing import IO, TYPE_CHECKING, Any, BinaryIO, Callable, Iterator, Optional
+
+from dissect.target import filesystem
+from dissect.target.exceptions import (
+    FilesystemError,
+    NotASymlinkError,
+    SymlinkRecursionError,
+)
+from dissect.target.helpers.compat.path_common import (
+    _DissectPathParents,
+    io_open,
+    isjunction,
+    scandir,
+)
+from dissect.target.helpers.polypath import normalize, normpath
+
+if TYPE_CHECKING:
+    from dissect.target.filesystem import Filesystem, FilesystemEntry
+    from dissect.target.helpers.compat.path_common import _DissectScandirIterator
+    from dissect.target.helpers.fsutil import stat_result
+
+
+class _DissectFlavour(_PosixFlavour):
+    is_supported = True
+
+    __variant_instances = {}
+
+    def __new__(cls, case_sensitive: bool = False, alt_separator: str = ""):
+        idx = (case_sensitive, alt_separator)
+        instance = cls.__variant_instances.get(idx, None)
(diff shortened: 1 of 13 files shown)


dissect/target/helpers/compat/path_39.py
```diff
@@ -0,0 +1,545 @@
...
+
+_dissect_accessor = _DissectAccessor()
+
+
+class PureDissectPath(PurePath):
+    _fs: Filesystem
+    _flavour = _DissectFlavour(case_sensitive=False)
+
+    def __reduce__(self) -> tuple:
+        raise TypeError("pickling is currently not supported")
```
pyrco, line 218:
```suggestion
        raise TypeError("TargetPath pickling is currently not supported")
```

pyrco (changes requested):
I noticed the 3.12 implementation still uses a "flavour" class insteasd of a module for `TargetPath._flavour`. I guess that is the easiest solution for now given dynamic separators?

I guess we should also tag&release flow.record and bump the dependency here to have the tests work.

Schamper:
> I noticed the 3.12 implementation still uses a "flavour" class insteasd of a module for TargetPath._flavour. I guess that is the easiest solution for now given dynamic separators?

Yeah, unfortunately because we use dynamic separators we need some degree of a "flavour" class. I was thinking about a different approach with static pre-created flavours/PurePath classes for each combination (similar to how there's a `PosixPurePath` and `WindowsPurePath`. That would also allow us to use `__slots__` again. I'll give that a try later to see how well that works.

> I guess we should also tag&release flow.record and bump the dependency here to have the tests work.

Yes, I'll bump the version in this PR already while we do that.

tests/helpers/test_fsutil.py
```diff
@@ -210,6 +221,369 @@ def test_stat_result():
...
+
+
+def test_target_path_is_relative_to(path_fs: VirtualFilesystem) -> None:
+    assert path_fs.path("/some/dir/file.txt").is_relative_to("/some/dir")
+    assert not path_fs.path("/some/dir/file.txt").is_relative_to("/some/other")
+
+
+def test_target_path_is_reserved(path_fs: VirtualFilesystem) -> None:
+    # TODO
+    assert not path_fs.path("CON").is_reserved()
```
pyrco, line 304: Maybe make the test as it should be, but add a skip decorator instead?

tests/tools/test_utils.py
```diff
@@ -28,7 +28,7 @@ def test_persist_execution_report():
             assert full_path.suffix == ".json"
             assert "2000-01-01-000000" in full_path.name
 
-            mocked_json_dumps.called_once_with(report_data)
+            mocked_json_dumps.assert_called_once_with(report_data, sort_keys=True, indent=4)
```
pyrco, line 31: This seems to be part of a different set of changes?

pyrco (changes requested):

tests/tools/test_utils.py
Schamper: Technically correct - it fixes this unit test. Python 3.12's `unittest.mock` errors here as it's more strict. I can move it out to a different PR if you'd like.

tests/helpers/test_fsutil.py
Schamper: Forgot about this TODO, added a comment instead. We do not have any reserved names atm, so this should just always return `False`.

tests/tools/test_utils.py
pyrco: Or make it a separate commit and don't squash it, also fine. I think it is good to have the `TargetPath` change set be as clean as possible.

tests/test_filesystem.py
```diff
@@ -659,12 +659,12 @@ def test_virtual_filesystem_link(vfs_path, link_path):
         ),
     ],
 )
-def test_virtual_filesystem_symlink(vfs_path, link_path):
+def test_virtual_filesystem_symlink(vfs_path: str, link_path: str) -> None:
```
pyrco, line 662: Aren't these changes technically part of the `VirtualFileesystem` symlink fix commit?

pyrco (changes requested):

tests/test_filesystem.py
Schamper: You're right, moved it over.

pyrco (approved):

PR #497, opened 2024-01-08
Support Windows installations on alternative drive letters

Schamper:
Some changes had to be made to the tar loader, since that's a place where we unfortunately hardcoded a drive letter. Depends on https://github.com/fox-it/acquire/pull/119 to make acquire also support non-default drive letters.


Schamper:
Also ping @JSCU-CNI since you guys also made some changes to the tar loader in the past. Unsure if this breaks anything for you.

JSCU-CNI:
Thanks for the heads up @Schamper. I took a quick look and it seems that this change does not break anything for us.

pyrco (approved):

PR #498, opened 2024-01-08
Improve driveletter selection for windows disks

pyrco:
As of c97c344 filesystems are always mounted. In the case of "fake" NTFS filesystems, as used by e.g. the dir and tar loaders, this means such a filesystem can be present under up to three different mount points, e.g.: sysvol, C: and $fs$\fs0

When a plugin does not use the root filesystem but the bare ntfs filesystem to parse ntfs specific properties, e.g. the mft plugin, it extends paths with a "driveletter" (read: mountpoint) to construct a path that is reachable from the root filesystem.

The old setup always returned the last mount point a drive was mounted under. The new setup changes this to return the first mount point present in the following order:

1. the first mount point with a letter and colon, e.g. C:,
2. the first mount point not starting with $fs$, e.g. sysvol,
3. the first mount point starting with $fs$, e.g. $fs$\fs0.

### DIFF
--- a/dissect/target/plugins/filesystem/ntfs/mft.py
+++ b/dissect/target/plugins/filesystem/ntfs/mft.py
@@ -123,6 +123,12 @@ class MftPlugin(Plugin):
 
         The Master File Table (MFT) contains primarily metadata about every file and folder on a NFTS filesystem.
 
+        If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the MFT properties
+        added to it through a "fake" ``NtfsFilesystem``), the paths returned in the MFT records are based on the
+        mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when
+        available.
+        When no drive letter can be determined, the path will show as e.g. ``\\$fs$\\fs0``.
+
         References:
             - https://docs.microsoft.com/en-us/windows/win32/fileio/master-file-table
         """
@@ -136,6 +142,10 @@ class MftPlugin(Plugin):
             if fs.__type__ != "ntfs":
                 continue
 
+            # If this filesystem is a "fake" NTFS filesystem, used to enhance a
+            # VirtualFilesystem, The driveletter (more accurate mount point)
+            # returned will be that of the VirtualFilesystem. This makes sure
+            # the paths returned in the records are actually reachable.
             drive_letter = get_drive_letter(self.target, fs)
             volume_uuid = get_volume_identifier(fs)
 
--- a/dissect/target/plugins/filesystem/ntfs/mft_timeline.py
+++ b/dissect/target/plugins/filesystem/ntfs/mft_timeline.py
@@ -105,6 +105,12 @@ class MftTimelinePlugin(Plugin):
 
         The Master File Table (MFT) contains metadata about every file and folder on a NFTS filesystem.
 
+        If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the MFT properties
+        added to it through a "fake" ``NtfsFilesystem``), the paths returned in the MFT records are based on the
+        mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when
+        available.
+        When no drive letter can be determined, the path will show as e.g. ``\\$fs$\\fs0``.
+
         References:
             - https://docs.microsoft.com/en-us/windows/win32/fileio/master-file-table
         """
@@ -112,6 +118,10 @@ class MftTimelinePlugin(Plugin):
             if fs.__type__ != "ntfs":
                 continue
 
+            # If this filesystem is a "fake" NTFS filesystem, used to enhance a
+            # VirtualFilesystem, The driveletter (more accurate mount point)
+            # returned will be that of the VirtualFilesystem. This makes sure
+            # the paths returned in the records are actually reachable.
             drive_letter = get_drive_letter(self.target, fs)
             extras = Extras(
                 serial=fs.ntfs.serial,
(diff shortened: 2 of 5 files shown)


dissect/target/plugins/filesystem/ntfs/utils.py
```diff
@@ -20,13 +23,33 @@ def get_drive_letter(target: Target, filesystem: NtfsFilesystem):
...
     """
+    # A filesystem can be known under multiple drives (mount points). If it is
+    # a windows system volume, there are the default sysvol and c: drives.
+    # If the target has a virtual ntfs filesystem, e.g. as constructed by the
+    # tar and dir loaders, there is also the /$fs$/fs<n> drive, under which the
+    # "fake" ntfs filesystem is mounted.
+    # The precedence for drives is first the drive letter drives (e.g. c:),
+    # second the "normally" named drives (e.g. sysvol) and finally the anonymous
+    # drives (e.g. /$fs/fs0).
     mount_items = (item for item in target.fs.mounts.items() if hasattr(item[1], "ntfs"))
```
Schamper, line 34:
```suggestion
    mount_items = (item for item in target.fs.mounts.items() if item[1].__type__ ==  "ntfs")
```

Schamper (changes requested):
Unrelated to the changes, but a small improvement on the type check for NTFS filesystems.

dissect/target/plugins/filesystem/ntfs/utils.py
pyrco: This will break dir and tar filesystems with a fake ntfs part though.

If the mft plugin walks the actual NtfsFilesystem (which checks on `__type__ == "ntfs"`). The paths generated from this will get an anonymous fs mountpoint (e.g. `$fs$\fs0`), but these paths are not reachable through things like target-shell, as the directory structures etc are not available, just the mft.

Schamper: Pff right, forgot about that and thought we'd already fixed that.

Schamper (approved):

Schamper:
Can you maybe add a note about `$fs$/fs0` etc to the MFT and other plugins that this touches?

Schamper (changes requested):
I meant moreso in the docstring of those plugins. This comment is also nice, but it'd also nice to see an explanation of where those `$fs$` paths are coming from if you're looking up the documentation for those plugins.

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -123,6 +123,11 @@ def mft(self, compact: bool = False):
 
         The Master File Table (MFT) contains primarily metadata about every file and folder on a NFTS filesystem.
 
+        If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the MFT properties
+        added to it through a "fake" ``NtfsFilesystem``), the paths returned in the MFT records are based on the
+        mount point of the ``VirtualFilesystem``. So although the primary location for the MFT may be under
+        ``\\fs\\$fs0``, any paths will start with e.g. ``C:\\``, which would have been the original location.
```
Schamper, line 129:
```suggestion
        mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when available.
        When no drive letter can be recovered, the path will show as e.g. ``\\$fs$\\fs0``.
```

dissect/target/plugins/filesystem/ntfs/mft_timeline.py
```diff
@@ -105,13 +105,22 @@ def mft_timeline(self, ignore_dos: bool = False):
 
         The Master File Table (MFT) contains metadata about every file and folder on a NFTS filesystem.
 
+        If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the MFT properties
+        added to it through a "fake" ``NtfsFilesystem``), the paths returned in the MFT records are based on the
+        mount point of the ``VirtualFilesystem``. So although the primary location for the MFT may be under
+        ``\\fs\\$fs0``, any paths will start with e.g. ``C:\\``, which would have been the original location.
```
Schamper, line 111:
```suggestion
        mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when available.
        When no drive letter can be recovered, the path will show as e.g. ``\\$fs$\\fs0``.
```

dissect/target/plugins/filesystem/ntfs/usnjrnl.py
```diff
@@ -34,6 +34,12 @@ def usnjrnl(self) -> Iterator[UsnjrnlRecord]:
         The Update Sequence Number Journal (UsnJrnl) is a feature of an NTFS file system and contains information about
         filesystem activities. Each volume has its own UsnJrnl.
 
+        If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the UsnJrnl
+        properties added to it through a "fake" ``NtfsFilesystem``), the paths returned in the UsnJrnl records
+        are based on the mount point of the ``VirtualFilesystem``. So although the primary location for the
+        UsnJrnl may be under ``\\fs\\$fs0``, any paths will start with e.g. ``C:\\``, which would have been the
+        original location.
```
Schamper, line 41:
```suggestion
        are based on the mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when available.
        When no drive letter can be recovered, the path will show as e.g. ``\\$fs$\\fs0``.
```

Schamper (changes requested):

Schamper (approved):

PR #501, opened 2024-01-11
Reduce log verbosity of wildcard keychain values

Schamper:
Also strip quotes from recovery key and file values, in case those were added with a quote for e.g. spaces in them.

### DIFF
--- a/dissect/target/helpers/keychain.py
+++ b/dissect/target/helpers/keychain.py
@@ -19,12 +19,15 @@ class Key(NamedTuple):
     value: Union[str, bytes]
     provider: str = None
     identifier: str = None
+    is_wildcard: bool = False
 
 
 KEYCHAIN: list[Key] = []
 
 
-def register_key(key_type: KeyType, value: str, identifier: str = None, provider: str = None) -> None:
+def register_key(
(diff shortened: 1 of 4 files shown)


dissect/target/helpers/keychain.py
```diff
@@ -58,7 +64,7 @@ def parse_key_type(key_type_name: str) -> KeyType:
 
 def register_wildcard_value(value: str) -> None:
     for key_type in KeyType:
-        register_key(key_type, value)
+        register_key(key_type, value, is_wildcard=True)
 
 
 def register_keychain_file(keychain_path: Path) -> None:
```
Horofic, line 70: Not sure. But dont you want to register wildcard values through keychain files as well? I think this can't be done like this currently. Correct me when im wrong though.

Horofic (changes requested):

dissect/target/helpers/keychain.py
Horofic: Seems like this currently not possible. Should be done at a later time, through a different PR.

Schamper: Wildcard values from the keychain file are currently not allowed, but this is something we could add in a future PR.

PR #503, opened 2024-01-12
Initial FortiOS support

Schamper:
This PR adds initial support for FortiOS images (currently tested with the VMs of FortiGate and FortiAnalyzer. The goal is to have some initial support available at first, and expand later on as more time for research becomes available.

For example, we also have some parsers for various log files, but they require some cleaning up as well as needing to figure out how feasible it is to parse those straight from a FortiOS image.


dissect/target/plugins/os/unix/linux/fortios/_os.py
```diff
@@ -0,0 +1,279 @@
...
+    def os(self) -> str:
+        return OperatingSystem.FORTIGATE.value
+
+    @export(record=UnixUserRecord)
+    def users(self) -> Iterator[UnixUserRecord]:
+        # TODO: Add FortiOS specific users
+        return super().users()
+
+
+def repair_lzma_stream(fh: BinaryIO) -> BinaryIO:
```
sud0woodo, line 113: <3

Schamper:
Now depends on https://github.com/fox-it/dissect.util/pull/39

dissect/target/plugin.py
```diff
@@ -683,10 +683,14 @@ def _walk(
...
-        _get_plugins(),
-        special_keys=special_keys,
-        only_special_keys=only_special_keys,
+    yield from sorted(
+        _walk(
+            _get_plugins(),
+            special_keys=special_keys,
+            only_special_keys=only_special_keys,
+        ),
+        key=lambda el: len(el["module"]),
```
pyrco, line 692:
```suggestion
        key=lambda plugin: len(plugin["module"]),
```
or `plugin_desc` or something

I noticed I didn't add any return type to `_walk()` which doesn't help either.

pyrco (changes requested):

dissect/target/plugins/os/unix/linux/fortios/_os.py
```diff
@@ -0,0 +1,382 @@
...
+
+        return result
+
+    @export(property=True)
+    def dns(self) -> list[str]:
+        """Return configured WAN DNS servers."""
+        entries = []
+        for _, entry in self._config["global-config"]["system"]["dns"].items():
+            entries.append(entry[0])
+        return entries
```
pyrco, line 163: This is not an OSPlugin method and should be put somewhere else.

Schamper: This currently overrides from `LinuxPlugin`, so I propose to leave it be for now, until we do a big OSPlugin cleanup.

pyrco: I see, that is nasty.

pyrco (approved):

pyrco (changes requested):
There is one testcase that fails

Schamper:
> There is one testcase that fails

Likely because this PR now depends on https://github.com/fox-it/dissect.util/pull/39. I moved the XZ repair code into dissect.util.

pyrco (approved):

PR #507, opened 2024-01-17
Add docker logs plugin

JSCU-CNI:
As discussed in #441.


dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
+
+                        pos += entry.header + 8
+
+                        yield DockerLogRecord(
+                            ts=ts.from_unix_us(entry.ts // 1000),
+                            container=container.name,
+                            stream=entry.source,
+                            message=entry.message if raw_messages else strip_log(entry.message, remove_backspaces),
+                            _target=self.target,
+                        )
```
Horofic, line 274:
```suggestion
        for data_root in self.installs:
            containers_path = data_root.joinpath("containers")
            # json-file log driver
            for log_file in containers_path.glob(("**/*.log")):
                container = log_file.parent

                if log_file.name.endswith("-json.log"):
                    for log_entry in self._parse_json_log(log_file):
                        yield DockerLogRecord(
                            ts=log_entry.get("time"),
                            container=container.name,  # container hash
                            stream=log_entry.get("stream"),
                            message=log_entry.get("log")
                            if raw_messages
                            else strip_log(log_entry.get("log"), remove_backspaces),
                            _target=self.target,
                        )
                else:
                    for log_entry in self._parse_local_log(log_file):
                        yield DockerLogRecord(
                            ts=ts.from_unix_us(log_entry.ts // 1000),
                            container=container.parent.name,  # container hash
                            stream=log_entry.source,
                            message=log_entry.message
                            if raw_messages
                            else strip_log(log_entry.message, remove_backspaces),
                            _target=self.target,
                        )

    def _parse_local_log(self, fh):
        fh = open_decompress(fh, "rb")

        if not hasattr(fh, "size"):  # for pytest
            fh.size = fh.seek(0, SEEK_END)
            fh.seek(0)

        while fh.tell() < fh.size:
            entry = c_local.entry(fh)

            if entry.header != entry.footer:
                self.target.log.warning(
                    "Could not reliably parse log entry at offset %i."
                    "Entry could be parsed incorrectly. Please report this "
                    "issue as Docker's protobuf could have changed.",
                    fh.tell(),
                )

            yield entry

    def _parse_json_log(self, fh):
        for line in open_decompress(fh, "rt"):
            try:
                entry = json.loads(line)
            except json.JSONDecodeError as e:
                self.target.log.warning("Could not decode JSON line in file %s", fh)
                self.target.log.debug("", exc_info=e)
                continue

            yield entry

```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
+
+                # local log driver
+                for log_file in container.glob("local-logs/container.log*"):
+                    fh = open_decompress(log_file, "rb")
+                    pos = 0
+
+                    if not hasattr(fh, "size"):  # for pytest
+                        fh.size = len(fh.read())
+                        fh.seek(0)
```
Horofic, line 253: If this is only for pytest. Can you fix it in pytest itself?

Horofic: You will notice I left this in the suggested code as well (albeit some changes). It should still be fixed in pytest itself there.

dissect/target/plugins/apps/container/docker.py
Horofic: Adding typehints is left as an exercise for the reader ;)

tests/plugins/apps/container/test_docker.py
```diff
@@ -58,14 +69,52 @@ def test_docker_plugin_containers(target_unix_users, fs_unix):
...
+def test_docker_plugin_logs_raw(target_linux_docker_logs: Target) -> None:
+    target_linux_docker_logs.add_plugin(DockerPlugin)
+    results = list(target_linux_docker_logs.docker.logs(raw_messages=True))
+
+    assert len(results) == 288
+
+    assert (
+        results[40].message
+        == '~ # \x1b[6necho \'\x08\x1b[J"ths \x08\x1b[J\x08\x1b[Js \x08\x1b[J\x08\x1b[Jis is a secret!" > secret.txt\r\n'
+    )  # noqa
```
Horofic, line 100: Can you make this noqa more specific

tests/conftest.py
```diff
@@ -406,6 +407,32 @@ def target_win_tzinfo_legacy(hive_hklm: VirtualHive, target_win: Target) -> Iter
...
+def fs_docker() -> TarFilesystem:
+    docker_tar = pathlib.Path(absolute_path("_data/plugins/apps/container/docker/docker.tgz"))
+    fh = docker_tar.open("rb")
+    docker_fs = TarFilesystem(fh)
+    yield docker_fs
+
+
+@pytest.fixture
+def target_linux_docker(fs_docker: TarFilesystem) -> Target:
+    mock_target = next(make_mock_target())
```
Horofic, line 427: I believe this is missing a tmp_path argument.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -35,11 +49,62 @@
...
+    [
+        ("datetime", "ts"),
+        ("string", "container"),
+        ("string", "stream"),
+        ("string", "message"),
+    ],
+)
+
+
+def find_installs(target: Target) -> Iterator[Path]:
```
Horofic, line 64: I wanted to suggest some changes to the `find_installs()` function, but couldn't due to some deleted lines. Please see the suggestions in a code-block below:
```python
def get_data_path(path: Path) -> Optional[str]:
    """Returns the configured Docker daemon data-root path."""
    try:
        config = json.loads(path.open("rt").read())
    except json.JSONDecodeError as e:
        log.warning("Could not read JSON file '%s'", path)
        log.debug(exc_info=e)

    return config.get("data-root")


def find_installs(target: Target) -> Iterator[Path]:
    """Attempt to find additional configured and existing Docker daemon data-root folders.

    References:
        - https://docs.docker.com/config/daemon/
    """

    default_config_paths = [
        # Linux
        "/etc/docker/daemon.json",
        "/var/snap/docker/current/config/daemon.json",
        # Windows
        "sysvol/ProgramData/docker/config/daemon.json",
    ]

    user_config_paths = [
        # Docker Desktop (macOS/Windows/Linux)
        ".docker/daemon.json",
    ]

    if (default_root := target.fs.path("/var/lib/docker")).exists():
        yield default_root

    for path in default_config_paths:
        if (config_file := target.fs.path(path)).exists():
            if (data_root_path := target.fs.path(get_data_path(config_file))).exists():
                yield data_root_path

    for path in user_config_paths:
        for user_details in target.user_details.all_with_home():
            if (config_file := user_details.home_path.joinpath(path)).exists():
                if (data_root_path := target.fs.path(get_data_path(config_file))).exists():
                    yield data_root_path
```

Horofic (changes requested):

dissect/target/helpers/docker.py
```diff
@@ -0,0 +1,136 @@
...
+    varint   m_len;         // message length
+    char     message[m_len];
+
+    // partial_log_metadata not implemented
+
+    uint32 footer;
+};
+"""
+
+c_local = cstruct(endian=">")
```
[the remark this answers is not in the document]
Schamper: I don't agree, they're already in a file called `docker.py`

dissect/target/helpers/docker.py
```diff
@@ -0,0 +1,136 @@
...
+c_local.addtype("varint", ProtobufVarint(cstruct=c_local, name="varint", size=1, signed=False, alignment=1))
+c_local.load(local_def, compiled=False)
+
+DOCKER_NS_REGEX = re.compile(r"\.(?P<nanoseconds>\d{7,})(?P<postfix>Z|\+\d{2}:\d{2})")
+
+
+def convert_timestamp(timestamp: str) -> str:
+    """
+    Docker sometimes uses (unpadded) 9 digit nanosecond precision
```
Schamper, line 45:
```suggestion
    """Docker sometimes uses (unpadded) 9 digit nanosecond precision
```
We also have a 120 character or so line limit, don't feel constrained to use it ;) also applies to the other docstrings.

dissect/target/helpers/docker.py
```diff
@@ -0,0 +1,136 @@
...
+    if not match:
+        return timestamp
+
+    # Take the first six digits and reconstruct the timestamp.
+    match = match.groupdict()
+    microseconds = match["nanoseconds"][:6]
+    return f"{timestamp[:19]}.{microseconds}{match['postfix']}"
+
+
+def convert_ports(ports: dict) -> dict:
```
Schamper, line 66:
```suggestion
def convert_ports(ports: dict[str, list | dict) -> dict[str, str]:
```
I think? May need to import `from __future__ import annotations` to make the `|` work.

dissect/target/helpers/docker.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def hash_to_image_id(hash: str) -> str:
+    """Convert the hash to an abbrevated docker image id."""
+    return hash.split(":")[-1][:12]
+
+
+RE_ANSI_ESCAPE = re.compile(r"\x1b(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
+
+ASCII_MAP = {
```
Schamper, line 98: Could you move these up with the other global constants?

dissect/target/helpers/docker.py
```diff
@@ -0,0 +1,136 @@
...
+    "\x08": "[BS]",
+    "\x09": "[TAB]",
+    "\x0A": "",  # \n
+    "\x0D": "",  # \r
+}
+
+
+def strip_log(input: Union[str, bytes], exc_backspace: bool = False) -> str:
```
Schamper, line 106:
```suggestion
def strip_log(input: str | bytes, exc_backspace: bool = False) -> str:
```
I suggested `|` earlier so may as well start being consistent.

dissect/target/helpers/protobuf.py
```diff
@@ -0,0 +1,76 @@
...
+        - https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py
+    """
+
+    def __init__(self, cstruct: cstruct, name: str, size: int, signed: bool, alignment: int = None):
+        self.signed = signed
+        super().__init__(cstruct, name, size, alignment)
+
+    def _read(self, stream: BinaryIO, context: dict[str, Any] = None) -> int:
+        self.size, data = read_varint(stream)
+        return decode_varint(data)
```
Schamper, line 28:
```suggestion
        return read_varint(stream)
```
We don't want to change `self.size` since that's a class variable for the type. So I'd suggest merging `read_varint` and `decode_varint` into one function since we don't really end up using the size.

dissect/target/helpers/protobuf.py
```diff
@@ -0,0 +1,76 @@
...
+    data = b""
+
+    while True:
+        size += 1
+        byte = stream.read(1)
+        data += byte
+        if int.from_bytes(byte, byteorder="big") & 0x80 == 0:
+            break
+
+    return size, data
```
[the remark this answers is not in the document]
Schamper: You could discard this with my previous comment.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -35,11 +49,62 @@
...
+
+        if config_path.startswith("$HOME"):
+            for user_details in target.user_details.all_with_home():
+                user_config = user_details.home_path.joinpath(config_path.replace("$HOME/", ""))
+                if user_config.exists():
+                    default_config_paths.append(str(user_config))
+
+        elif config_file.exists():
+            try:
+                config = json.loads(config_file.open("rt").read())
```
Schamper, line 95:
```suggestion
                config = json.loads(config_file.read_text())
```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -35,11 +49,62 @@
...
+            except json.JSONDecodeError as e:
+                log.warning(f"Could not read JSON file '{config_file}'")
```
Schamper, line 97:
```suggestion
                target.log.warning("Could not read JSON file %s", config_file)
```
Logging from the target log object ties it to the target context.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -35,11 +49,62 @@
...
+                log.debug(exc_info=e)
```
Schamper, line 98:
```suggestion
                target.log.debug("", exc_info=e)
```
I believe this errors otherwise.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
 
     __namespace__ = "docker"
 
-    DOCKER_PATH = "/var/lib/docker"
+    def __init__(self, target) -> None:
```
Schamper, line 116:
```suggestion
    def __init__(self, target: Target):
```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
+        super().__init__(target)
+        self.installs = set(find_installs(target))
 
     def check_compatible(self) -> None:
-        if not self.target.fs.path(self.DOCKER_PATH).exists():
-            raise UnsupportedPluginError("No Docker path found")
+        if not any(self.installs):
+            raise UnsupportedPluginError("No Docker install(s) found.")
```
Schamper, line 122:
```suggestion
            raise UnsupportedPluginError("No Docker install(s) found")
```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
-                    image_id=_hash_to_image_id(hash),
-                    created=created,
-                    hash=hash,
-                    _target=self.target,
-                )
+        for data_root in self.installs:
+            images_path = f"{data_root}/image/overlay2/repositories.json"
+
+            if (fp := self.target.fs.path(images_path)).exists():
+                repositories = json.loads(fp.read_text()).get("Repositories")
```
Schamper, line 132:
```suggestion
            if images_path.exists():
                repositories = json.loads(images_path.read_text()).get("Repositories")
```
Combined with the suggestion from @Horofic above.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
+                self.target.log.debug(f"No docker images found, file {images_path} does not exist.")
+                return
+
+            for name, tags in repositories.items():
+                for tag, hash in tags.items():
+                    image_metadata_path = f"{data_root}/image/overlay2/imagedb/content/sha256/{hash.split(':')[-1]}"
+                    created = None
+
+                    if (fp := self.target.fs.path(image_metadata_path)).exists():
+                        image_metadata = json.loads(fp.read_text())
```
Schamper, line 143:
```suggestion
                    if image_metadata_path.exists():
                        image_metadata = json.loads(image_metadata_path.read_text())
```
Same here.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
-
-
-def _hash_to_image_id(hash: str) -> str:
-    return hash.split(":")[-1][:12]
+        for data_root in self.installs:
+            containers_path = f"{data_root}/containers"
+
+            for container in self.target.fs.path(containers_path).iterdir():
+                if (fp := self.target.fs.path(f"{container}/config.v2.json")).exists():
+                    config = json.loads(fp.read_text())
```
Schamper, line 164:
```suggestion
                if (config_path := containers_path.joinpath("config.v2.json")).exists():
                    config = json.loads(config_path.read_text())
```

tests/_data/plugins/apps/container/docker/docker.tgz
Schamper, line 1: Can you add this to LFS?

dissect/target/helpers/protobuf.py
JSCU-CNI: I think we do need to store the size somewhere, as it seems to be required for `_read_array` to work. When I set `ProtobufVarint(size=0)` in the docker helper, we get a `MemoryError` as `_read_array` ends up in an infinite loop.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +113,162 @@ class DockerPlugin(Plugin):
...
+
+                    if config.get("State").get("Running"):
+                        ports = config.get("NetworkSettings").get("Ports", {})
+                        pid = config.get("Pid")
+                    else:
+                        ports = config.get("Config").get("ExposedPorts", {})
+                        pid = False
```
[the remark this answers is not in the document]
JSCU-CNI: I believe `flow.record` will cast `False` to a `0` if I'm not mistaken

dissect/target/helpers/protobuf.py
Schamper: I expect the `_read_array` to be buggy now then. The `.size` attribute is usually a static size for a single element of this type. You can see this will result in weird behaviour, since `BytesInteger` will try to do an optimized array read: https://github.com/fox-it/dissect.cstruct/blob/main/dissect/cstruct/types/bytesinteger.py#L83. I'm not sure if the current unit test catches this edge case (different sized variable integers?).

The right approach would likely be to either implement a custom `_read_array`, or to fall back on the `_read_array` of the base class: https://github.com/fox-it/dissect.cstruct/blob/main/dissect/cstruct/types/base.py#L95

Note that this will also need to be changed once https://github.com/fox-it/dissect.cstruct/pull/35 is merged, but let's cross that bridge once we get there. At least there are some unit tests :smile:

dissect/target/plugins/apps/container/docker.py
JSCU-CNI: I attempted to mock this in pytest, which is a headache. What do you suggest we do here? I think it would be benificial to have the `size` attribute available just as you would outside of pytest. 

Do you suggest adding `size` to `MappedFile` or `TargetPath`? I am too unfamiliar with these classes to decide which is a good solution.

dissect/target/helpers/docker.py
JSCU-CNI: iirc you wanted us to change `|`'s to `Union`'s in a previous PR. Is this a definitive style change?

JSCU-CNI:
Thanks for the thorough review @Schamper and @Horofic. I've implemented all feedback where applicable and replied to all comments where things are still unclear.

dissect/target/plugins/apps/container/docker.py
Schamper: I don't think `.size` is available either in the case of a compressed file, so this would break then as well.

I think the best approach to do here is just run the file out until it's EOF. That way you also don't unnecessarily decompress until the end of the file in case of a compressed file. This is a common pattern we use with cstruct.

```python
while True:
    try:
        entry = c_local.entry(fh)
        ...
    except EOFError:
        break
```

dissect/target/helpers/docker.py
Schamper: Sorry 😅 Ideally a linter applies these kinds of things for us in the future so that it's easier to be consistent. I'm personally leaning towards using this new syntax for new code. I'll try to be consistent myself 😉

dissect/target/helpers/protobuf.py
```diff
@@ -33,7 +33,7 @@ def decode_varint(stream: BinaryIO) -> int:
     i = 0
     while True:
         byte = stream.read(1)
-        result |= (int.from_bytes(byte) & 0x7F) << (i * 7)
+        result |= (int.from_bytes(byte, byteorder="big") & 0x7F) << (i * 7)
```
Schamper, line 36:
```suggestion
        result |= (byte[0] & 0x7F) << (i * 7)
```
You're only reading a single byte so no need for an expensive `from_bytes` call.

Schamper (changes requested):

dissect/target/helpers/protobuf.py
JSCU-CNI: When setting `ProtobufVarint(size=1)` in `test_protobuf.py` (just like the docker plugin does), it seems that the position of the stream is incorrectly calculated. 

```
        aaa = b"a" * 123456
        foo = cs.foo(BytesIO(b"\x00\x00\x00\x01\xc0\xc4\x07" + aaa + b"\x01\x02\x03"))
        assert foo.foo == 1
        assert foo.size == 123456
>       assert foo.bar == aaa
E       AssertionError: assert b'\xc4\x07aaa...aaaaaaaaaaaaa' == b'aaaaaaaaaaa...aaaaaaaaaaaaa'
```

As you can see the last two bytes of the `varint` are prepended to the value of `char bar[size]`. I think this is because the stream is not read 3 bytes further before parsing the next cstruct type, but just one byte. How can we tell the parent cstruct parser to seek two bytes further? 

Setting `ProtobufVarint(size=0)` in the `docker.py` helper also seems to break, but for a different reason, as the `s_len` or `m_len` is incorrectly calculated as a huge number.

dissect/target/helpers/protobuf.py
```diff
@@ -0,0 +1,55 @@
...
+    If we have not reached the end of a varint, the msb will be 1.
+    We read every byte from our current position until the msb is 0.
+    """
+    result = 0
+    i = 0
+    while True:
+        byte = stream.read(1)
+        result |= (byte[0] & 0x7F) << (i * 7)
+        i += 1
+        if int.from_bytes(byte, byteorder="big") & 0x80 == 0:
```
Schamper, line 38:
```suggestion
        if byte[0] & 0x80 == 0:
```

Schamper (changes requested):

dissect/target/helpers/protobuf.py
Schamper: I've pushed a fix and support for reading/writing arrays for varint, including some tests.

JSCU-CNI: Thanks for the commit, this seems to fix both problems.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +119,165 @@ class DockerPlugin(Plugin):
...
+                        "Entry could be parsed incorrectly. Please report this "
+                        "issue as Docker's protobuf could have changed.",
+                        fh.tell(),
+                        path,
+                    )
+                yield entry
+            except EOFError:
+                break
+
+    def _parse_json_log(self, path: Path) -> Iterator[io.BytesIO]:
```
Horofic, line 275:
```suggestion
    def _parse_json_log(self, path: Path) -> Iterator[dict]:
```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,128 +119,165 @@ class DockerPlugin(Plugin):
...
+                            ts=ts.from_unix_us(log_entry.ts // 1000),
+                            container=container.parent.name,  # container hash
+                            stream=log_entry.source,
+                            message=log_entry.message
+                            if raw_messages
+                            else strip_log(log_entry.message, remove_backspaces),
+                            _target=self.target,
+                        )
+
+    def _parse_local_log(self, path: Path) -> Iterator[io.BytesIO]:
```
Horofic, line 257:
```suggestion
    def _parse_local_log(self, path: Path) -> Iterator[c_local.entry]:
```

Horofic (changes requested):

dissect/target/helpers/docker.py
Schamper, line 1: Could you merge the contents of this file with `plugins/apps/container/docker.py`? 

The reasoning behind this is that `dissect/target/helpers` is mostly for "core" helpers. Ideally we keep plugin related stuff in the plugins tree. Sometimes it's messy to keep it in the file though, and this is a general problem we have in more places. I don't have a great solution for that yet. 

However, I think in this specific case, it would be fine to put this in the `docker.py` of the plugin itself. The protobuf stuff is "generic" enough that it's fine to keep it in there (at some point it may even grow into a separate package with other serialization/protocol parsers).

Schamper (changes requested):

dissect/target/helpers/docker.py
JSCU-CNI: Done. See [6ff435b](https://github.com/fox-it/dissect.target/pull/507/commits/6ff435bcb12da1a3ee3c6b50918358b6f9a23750)

dissect/target/plugins/apps/container/docker.py
```diff
@@ -48,84 +111,218 @@ class DockerPlugin(Plugin):
...
-                    _target=self.target,
-                )
+        for data_root in self.installs:
+            images_path = data_root.joinpath("image/overlay2/repositories.json")
+
+            if images_path.exists():
+                repositories = json.loads(images_path.read_text()).get("Repositories")
+            else:
+                self.target.log.debug("No docker images found, file %s does not exist.", images_path)
+                return
```
Schamper, line 133:
```suggestion
                continue
```
Otherwise we don't process any other installs.

Schamper (changes requested):

Schamper (dismissed):

Horofic (changes requested):
Seems like some of the tests are failing now. Could you fix them? Other than that LGTM!

JSCU-CNI:
Thanks @Horofic. There seems to be a race condition in the order in which records are returned. I am unable to reproduce the failing tests locally. Unfortunately sorting on records is not possible, what do you recommend as a fix?

tests/conftest.py
```diff
@@ -407,6 +408,32 @@ def target_win_tzinfo_legacy(hive_hklm: VirtualHive, target_win: Target) -> Iter
     yield target_win
 
 
+@pytest.fixture
+def target_linux_docker_logs(target_linux: Target, fs_linux: VirtualFilesystem) -> Target:
```
Schamper, line 412: Can you move these fixtures to the `test_docker.py` file? I don't think they are currently used in other files.

Schamper (changes requested):

Schamper:
> Thanks @Horofic. There seems to be a race condition in the order in which records are returned. I am unable to reproduce the failing tests locally. Unfortunately sorting on records is not possible, what do you recommend as a fix?

Maybe the log files are iterated/loaded in a different order (based on some filesystem/Python listdir quirk).

Horofic:
> > Thanks @Horofic. There seems to be a race condition in the order in which records are returned. I am unable to reproduce the failing tests locally. Unfortunately sorting on records is not possible, what do you recommend as a fix?
> 
> Maybe the test log files are iterated/loaded in a different order (based on some filesystem/Python listdir quirk).

Check. In that case, incorporating the last remarks from @Schamper should do it.

Horofic (dismissed):

Schamper:
> > > Thanks @Horofic. There seems to be a race condition in the order in which records are returned. I am unable to reproduce the failing tests locally. Unfortunately sorting on records is not possible, what do you recommend as a fix?
> > 
> > 
> > Maybe the test log files are iterated/loaded in a different order (based on some filesystem/Python listdir quirk).
> 
> Check. In that case, incorporating the last remarks from @Schamper should do it.

It was a remark at a possible cause, the suggestion wasn't to fix that :smile:

JSCU-CNI:
Perhaps https://github.com/fox-it/flow.record/pull/108/files fixes this when adding `results.sort()` in the tests.

Schamper:
> Perhaps [fox-it/flow.record#108](https://github.com/fox-it/flow.record/pull/108) fixes this when adding `results.sort()` in the tests or the following for the meantime:
> 
> ```python
> import operator
> results.sort(key=operator.attrgetter('_generated'))
> ```
> 
> Interesting to see the 3.12 CI environment does not seem to be impacted by this.

I don't think this will fix the problem, as I think the problem lies in the order at which the log files are mapped into the virtual filesystem. At least on my local system, it gets mapped correctly, but for some reason in the CI it gets mapped the other way around (probably something to do with timestamps).

Sorting on the record content (e.g. ts, message) should result in a consistent test output.

Schamper (approved):

JSCU-CNI:
Seems like the `pypy39` pipeline is having a rough Friday :( 

```
Fatal Python error: Segmentation fault

Schamper:
> Seems like the `pypy39` pipeline is having a rough Friday :(
> 
> ```
> Fatal Python error: Segmentation fault
> ```

Yeah unfortunately the PyPy 3.9 CI does that. We have to keep re-running it until it doesn't segfault :smile:

Horofic (approved):

PR #508, opened 2024-01-17
Add `map_dir_from_tar` and `map_file_from_tar`

JSCU-CNI:
As discussed in https://github.com/fox-it/dissect.target/pull/492#issuecomment-1874128205.


Schamper:
Is there a specific reason why a simple `mount` wouldn't suffice? Seems a bit wasteful to walk over the entire tar and map it into the VirtualFilesystem.

JSCU-CNI:
> Is there a specific reason why a simple mount wouldn't suffice? Seems a bit wasteful to walk over the entire tar and map it into the VirtualFilesystem.

I've simplified the helpers in [64a46b2](https://github.com/fox-it/dissect.target/pull/508/commits/64a46b2c853f51fd5d49d8edd67a41ea4abc3c9d). It is indeed not necessary to mount every file individually when using `map_dir_from_tar`. It now uses `self.mount`. However for `map_file_from_tar` we need to set the final destination `vfspath` manually, which can differ from the actual path inside the tar. Do you have a simpler solution?

Schamper:
> I've simplified the helpers in [64a46b2](https://github.com/fox-it/dissect.target/pull/508/commits/64a46b2c853f51fd5d49d8edd67a41ea4abc3c9d). It is indeed not necessary to mount every file individually when using `map_dir_from_tar`. It now uses `self.mount`. However for `map_file_from_tar` we need to set the final destination `vfspath` manually, which can differ from the actual path inside the tar. Do you have a simpler solution?

This is to make some things work correctly with the `.path`/`.name` attribute of the newly mapped file entry, right? I think you're running into a similar problem I was having recently. It stems from the fact that we let filesystem entries have their own path/filename, instead of that being dictated by the directory listing they're in (as is done in every Unix filesystem. Turns out they knew what they were doing). 

We plan to make a change there to introduce "DirEntry" objects sometime in the future, but as this is a pretty big change it won't come any time soon.

I went and found where I was having that problem. I opted to solve it by changing the `.name` attribute of the filesystem entry object. I think you could apply the same logic here, since you're not going to use it anymore. That would also make the `stat` change in `VirtualFile` redundant I believe.

JSCU-CNI:
Thanks for the pointer @Schamper. Implemented the simplified mapping in [50ce246](https://github.com/fox-it/dissect.target/pull/508/commits/50ce2462842703293aa3cc811a669a6b53066411).

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
...
+                file.name = posixpath.basename(vfspath)
+                self.map_file_entry(vfspath, file)
+                break
+
+        self.mount(vfspath, tfs)
+
+    def map_file_from_tar(self, vfspath: str, tar_file: Union[str, pathlib.Path]) -> None:
+        """Map a single file in a tar archive to the given path in the VFS.
+
+        The provided tar archive should contain *one* file."""
```
Schamper, line 1268:
```suggestion
        The provided tar archive should contain *one* file.
        """
```

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
             entry_name = fsutil.basename(vfspath, alt_separator=self.alt_separator)
             directory.add(entry_name, entry)
 
+    def map_dir_from_tar(self, vfspath: str, tar_file: Union[str, pathlib.Path], custom_path: bool = False) -> None:
+        """Map files in a tar onto the VFS."""
```
Schamper, line 1246: Can you add some docstrings for the arguments? I'm not sure what `custom_path` means right now, other than that it triggers the "single file" code path.

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
...
+                tar_file = pathlib.Path(tar_file)
+            except TypeError:
+                raise ValueError("tar_file should be a string or Path instance")
+
+        vfspath = fsutil.normalize(vfspath, alt_separator=self.alt_separator)
+        tfs = TarFilesystem(tar_file.open("rb"))
+
+        if custom_path:
+            for file in [f[0] for _, _, f in tfs.walk_ext("/") if f]:
+                file.name = posixpath.basename(vfspath)
```
Schamper, line 1259:
```suggestion
                file.name = fsutil.basename(vfspath)
```

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
...
+                self.map_file_entry(vfspath, file)
+                break
```
Schamper, line 1261: Should this not be a return? Now the `mount` below still triggers.

dissect/target/filesystem.py
```diff
@@ -4,6 +4,8 @@
 import io
 import logging
 import os
+import pathlib
+import posixpath
```
Schamper, line 8:
```suggestion
```
No longer necessary.

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
             entry_name = fsutil.basename(vfspath, alt_separator=self.alt_separator)
             directory.add(entry_name, entry)
 
+    def map_dir_from_tar(self, vfspath: str, tar_file: Union[str, pathlib.Path], custom_path: bool = False) -> None:
```
Schamper, line 1245:
```suggestion
    def map_dir_from_tar(self, vfspath: str, tar_file: str | pathlib.Path, custom_path: bool = False) -> None:
```
In spirit of my comment from the Docker PR.

dissect/target/filesystem.py
```diff
@@ -1238,6 +1242,32 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
...
+
+        if custom_path:
+            for file in [f[0] for _, _, f in tfs.walk_ext("/") if f]:
+                file.name = posixpath.basename(vfspath)
+                self.map_file_entry(vfspath, file)
+                break
+
+        self.mount(vfspath, tfs)
+
+    def map_file_from_tar(self, vfspath: str, tar_file: Union[str, pathlib.Path]) -> None:
```
Schamper, line 1265:
```suggestion
    def map_file_from_tar(self, vfspath: str, tar_file: str | pathlib.Path) -> None:
```

Schamper (changes requested):

dissect/target/filesystem.py
```diff
@@ -1238,6 +1232,44 @@ def map_file_entry(self, vfspath: str, entry: FilesystemEntry) -> None:
...
+        tfs = TarFilesystem(tar_file.open("rb"))
+
+        if map_single_file:
+            # We map the first file we find in the tar to the provided vfspath.
+            for file in [f[0] for _, _, f in tfs.walk_ext("/") if f]:
+                file.name = fsutil.basename(vfspath)
+                self.map_file_entry(vfspath, file)
+                return
+
+        self.mount(vfspath, tfs)
```
Schamper, line 1260:
```suggestion
        if map_single_file:
            # We map the first file we find in the tar to the provided vfspath.
            for file in [f[0] for _, _, f in tfs.walk_ext("/") if f]:
                file.name = fsutil.basename(vfspath)
                self.map_file_entry(vfspath, file)
                return
        else:
            self.mount(vfspath, tfs)
```
Otherwise we may get an erroneous mount if you try to `map_single_file` with an empty tar. Harmless, but we can avoid it.

PR #510, opened 2024-01-22
Add FortiOS rootfs.gz decrypt functionality

JSCU-CNI:
This PR adds rudimentary support for decrypting `rootfs.gz` files on FortiOS 7.4.1 and up.

### DIFF
--- a/dissect/target/plugins/os/unix/linux/fortios/_os.py
+++ b/dissect/target/plugins/os/unix/linux/fortios/_os.py
@@ -39,7 +39,7 @@ class FortiOSPlugin(LinuxPlugin):
 
     def _load_config(self) -> dict:
         CONFIG_FILES = {
-            "/data/system.conf": None,
+            "/data/system.conf": "global-config",  # FortiManager
             "/data/config/daemon.conf.gz": "daemon",  # FortiOS 4.x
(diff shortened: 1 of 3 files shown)


Schamper:
I've fixed the decryption to work with the existing pycryptodome dependency. It seems to all work alright on my end, but if you could give it a quick test-spin, this should be good to go!

Would be cool if we could dynamically extract the encryption key. I was working on some Ivanti appliances with the loop-AES encryption, which also have hardcoded keys in the kernel. I haven't looked at the FortiOS kernel yet, but those have a symbol pointing to the key, making recovery of it fairly trivial. Perhaps it's similar for FortiOS, or a pattern match could suffice.

Way too much work for relatively little gain, though. But perhaps fun to explore some day :smile:

pyrco:
Would it be an idea to split this PR in 3 not-to-be-squashed commits, 1 for the robustness improvements in `locale.py`, 1 for the additional file locations in `generic.py` and 1 with the actual decrypt functionality in `_os.py`?

Schamper:
> Would it be an idea to split this PR in 3 not-to-be-squashed commits, 1 for the robustness improvements in `locale.py`, 1 for the additional file locations in `generic.py` and 1 with the actual decrypt functionality in `_os.py`?

Split it over 4 to also separate the FortiManager out.

pyrco (approved):

PR #511, opened 2024-01-23
Add schedlgu.txt file parser

Horofic:
(DIS-2136)


dissect/target/plugins/os/windows/tasks.py
```diff
@@ -149,3 +224,57 @@ def tasks(self) -> Iterator[Union[TaskRecord, GroupedRecord]]:
...
+        self.target = target
+        self.paths = [self.target.fs.path(path) for path in self.PATHS if self.target.fs.path(path).exists()]
+
+    def check_compatible(self) -> None:
+        if len(self.paths) == 0:
+            raise UnsupportedPluginError("No SchedLgU.txt file found.")
+
+    @export(record=SchedLgURecord)
+    def schedlgu(self) -> Iterator[SchedLgURecord]:
+        """Return all evnets in the Task Scheduler Service transaction log file (SchedLgU.txt).
```
Poeloe, line 249:
```suggestion
        """Return all events in the Task Scheduler Service transaction log file (SchedLgU.txt).
```

dissect/target/plugins/os/windows/tasks.py
```diff
@@ -149,3 +224,57 @@ def tasks(self) -> Iterator[Union[TaskRecord, GroupedRecord]]:
...
+
+        Older Windows systems may log ``.job`` tasks that get started remotely in the SchedLgU.txt file.
+        In addition this log file records when the Task Scheduler service starts and stops.
```
Poeloe, line 252:
```suggestion
        In addition, this log file records when the Task Scheduler service starts and stops.
```

dissect/target/plugins/os/windows/tasks.py
```diff
@@ -149,3 +224,57 @@ def tasks(self) -> Iterator[Union[TaskRecord, GroupedRecord]]:
...
+
+        Adversaries may use malious ``.job`` files to gain persistence on a system.
+
+        Yield:
+            ts (datetime): The timestamp of the event.
+            job (str): The name of the ``.job`` file.
+            command (str): The command executed.
+            status (str): The status of the event (Finished, completed, exited, stopped).
```
Poeloe, line 260:
```suggestion
            status (str): The status of the event (finished, completed, exited, stopped).
```

dissect/target/plugins/os/windows/tasks.py
```diff
@@ -149,3 +224,57 @@ def tasks(self) -> Iterator[Union[TaskRecord, GroupedRecord]]:
...
+            exit_code (int): The exit code of the event.
+            version (str): The version of the Task Scheduler service.
+        """
+
+        for path in self.paths:
+            content = path.read_text(encoding="UTF-16", errors="surrogateescape")
+            pattern = re.compile(r"\".+\n.+\n\s{4}.+\n|\".+\n.+", re.MULTILINE)
```
Poeloe, line 267: I would suggest to make this an attribute of the plugin, such that it only compiles once and not every iteration ;)

dissect/target/plugins/os/windows/tasks.py
```diff
@@ -71,6 +76,76 @@
...
+
+        return job, command
+
+    @classmethod
+    def from_line(cls, line: str) -> SchedLgU:
+        """Parse a group of SchedLgU.txt lines."""
+        event = cls()
+        lines = line.splitlines()
+
+        if len(lines) == 3:
```
Poeloe, line 127: Since it is not trivial why the number of lines etc. is a reason to treat the whole line differently, it would be nice to supply a small comment. Maybe show an example line, to make it extra clear :)

Poeloe (changes requested):

dissect/target/plugins/os/windows/tasks.py
```diff
@@ -71,6 +76,89 @@
...
+
+        return ts
+
+    @staticmethod
+    def _parse_job(line: str) -> tuple[str, str]:
+        job, command = line.split("(", maxsplit=1)
+        command = command.rstrip(")")
+        job = job.strip('"').rstrip('" ')
+
+        return job, command
```
Poeloe, line 119: Maybe do something like this? (but make the pattern again a global attribute of the class and maybe do some validation for the match ;)
```suggestion
        pattern = re.compile(r"\"([\w\s.]*)\" \(([\w\s.]*)\)")
        match = re.match(pattern, line)
        
        return match.group(1), match.group(2)

        return job, command
```

PR #522, opened 2024-02-01
Add AttributeError to except ImportError statements

Miauwkeru:
In some cases windows throws an AttributeError instead of an ImportError during an import


Schamper:
What are these "some cases"? Is a blanket find & replace of every single `ImportError` really the correct approach, or could a more specific fix be made?

pyrco (approved):

Schamper (changes requested):
The "sometimes" is actually a bug in our internal Python distribution, so I don't think any of these PRs should be merged as they are cluttering and maybe even misleading.

PR #524, opened 2024-02-02
Replace mimikatz binary in quarantine data with DUMMY_PAYLOAD

Miauwkeru:
So windows defender doesn't remove it

(DIS-2787)


Poeloe (approved):
LGTM

PR #526, opened 2024-02-06
Fix some issues with tests not being independent from other tests

pyrco:
These are due to the tests (inadvertently) changing global state.

(DIS-2971)

### DIFF
--- a/tests/loaders/test_cb.py
+++ b/tests/loaders/test_cb.py
@@ -9,12 +9,15 @@ from dissect.target.target import Target
 
 @pytest.fixture
 def mock_cbc_sdk(monkeypatch: pytest.MonkeyPatch) -> Iterator[MagicMock]:
-    try:
-        del sys.modules["dissect.target.loader"]
-    except KeyError:
-        pass
-
     with monkeypatch.context() as m:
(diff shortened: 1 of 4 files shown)


tests/loaders/test_cb.py
```diff
@@ -9,10 +9,13 @@
...
-        del sys.modules["dissect.target.loader"]
-    except KeyError:
-        pass
+    # The references to cbc_sdk properties in the cb loader will point to the
+    # MagicMock created for the first test function that runs.
+    # Thus we need to delete the cb module to force it to reload when importing
+    # from it so the MagickMock used in the cb module will be the one created
+    # for the test function that is running.
+    if "dissect.target.loaders.cb" in sys.modules:
+        del sys.modules["dissect.target.loaders.cb"]
```
Schamper, line 18: Would this work/be better? https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.delitem

Schamper (changes requested):

tests/loaders/test_cb.py
pyrco: Yes, that's even nicer

PR #528, opened 2024-02-06
Add JSON and YAML support to configutil

Miauwkeru:
(DIS-2160)


dissect/target/helpers/configutil.py
```diff
@@ -20,6 +21,7 @@
     Union,
 )
 
+import yaml
```
Schamper, line 24: What library is this from?

Poeloe (changes requested):

Poeloe (approved):
LGTM!

PR #531, opened 2024-02-08
Add basic cpio filesystem

JSCU-CNI:
This PR adds a basic cpio archive loader which should enable you to browse `initrd.img` files found on most Linux distributions. Currently supports gzip, bz2 and zstd compression (using `open_decompress`).

```bash
$ target-shell /tmp/debian-12.4.0/*.vmx
debian /> enter /boot/initrd.img-6.1.0-15-amd64
initrd.img-6.1.0-15-amd64 /> ls
bin
conf
etc
init
lib
lib64
run
sbin
scripts
usr
initrd.img-6.1.0-15-amd64 /> 
```

This PR does not include support for initramfs/initrd archives embedded inside MBR files / `vmlinuz` kernels.


dissect/target/loaders/cpio.py
```diff
@@ -0,0 +1,32 @@
...
+
+        # gzip, bz2 or zstd
+        if magic[:2] == b"\x1f\x8b" or magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
+            magic = open_decompress(path, "rb").read(6)
+
+        return magic in [b"070701", b"070707", b"070702"] or magic[:2] in [b"\x71\xc7", b"\xc7\x71"]
+
+    def map(self, target: Target) -> None:
+        target.filesystems.add(self.cpio)
+        target.fs.mount("/", self.cpio)
```
Schamper, line 32:
```suggestion
        tfs = TarFilesystem(open_decompress(self.path, "rb"), tarinfo=cpio.CpioInfo)
        target.filesystems.add(tfs)
```
We don't necessarily have to open it in the `__init__`. We should also leave the mounting to the OSPlugin's.

dissect/target/loaders/cpio.py
```diff
@@ -0,0 +1,32 @@
...
+
+class CpioLoader(Loader):
+    """Load (compressed) CPIO files."""
+
+    def __init__(self, path: Path, **kwargs):
+        super().__init__(path)
+
+        fh = open_decompress(path, "rb")
+        self.cpio = TarFilesystem(fh, tarinfo=cpio.CpioInfo)
+
```
Schamper, line 19:
```suggestion
```

dissect/target/loaders/cpio.py
```diff
@@ -0,0 +1,32 @@
...
+    @staticmethod
+    def detect(path: Path) -> bool:
+        magic = path.open("rb").read(6)
```
Schamper, line 22: We might be dealing with real files here, so best to use some context managers for all the opening you're doing in `detect()`.

dissect/target/loaders/cpio.py
```diff
@@ -0,0 +1,32 @@
...
+
+        # gzip, bz2 or zstd
+        if magic[:2] == b"\x1f\x8b" or magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
+            magic = open_decompress(path, "rb").read(6)
+
+        return magic in [b"070701", b"070707", b"070702"] or magic[:2] in [b"\x71\xc7", b"\xc7\x71"]
```
Schamper, line 28: You could chose to re-use `dissect.util.cpio.detect_header` here: https://github.com/fox-it/dissect.util/blob/main/dissect/util/cpio.py#L190

tests/loaders/test_cpio.py
```diff
@@ -0,0 +1,36 @@
...
+
+from dissect.target import Target
+from dissect.target.loaders.cpio import CpioLoader
+from tests._utils import absolute_path
+
+
+def test_cpio_uncompressed(target_default: Target) -> None:
+    cpio_path = Path(absolute_path("_data/loaders/cpio/initrd.img-6.1.0-17-amd64"))
+
+    loader = CpioLoader(cpio_path)
```
Schamper, line 11: Can you also test the `detect()`?

dissect/target/helpers/fsutil.py
```diff
@@ -16,9 +16,16 @@
 try:
     import bz2
 
-    HAVE_BZ2 = True
+    HAS_BZ2 = True
```
Schamper, line 19: As much as I like the more correct grammar, we currently use `HAVE_` in a ton of places in Dissect, so it's best to stick with it for now. We could change it at some point, but I'd prefer to change it all at once. And then this PR is not the place for that.

Schamper (changes requested):

dissect/target/loaders/cpio.py
```diff
@@ -0,0 +1,28 @@
...
+
+from dissect.util import cpio
+
+from dissect.target import Target
+from dissect.target.filesystems.tar import TarFilesystem
+from dissect.target.helpers.fsutil import open_decompress
+from dissect.target.loader import Loader
+
+
+class CpioLoader(Loader):
```
Schamper, line 11: I was thinking about this a little bit more and it might be better to implement this as a `CpioFilesystem` which would be a subclass of `TarFilesystem`. That way it would get picked up automatically by the `RawLoader`, and it's a little more flexible in other use-cases.

Schamper (changes requested):

dissect/target/loaders/cpio.py
JSCU-CNI: Sounds like a good improvement for the next PR ;)

dissect/target/helpers/fsutil.py
```diff
@@ -480,13 +502,23 @@ def open_decompress(
 
     if magic[:2] == b"\x1f\x8b":
         return gzip.open(file, mode, encoding=encoding, errors=errors, newline=newline)
-    # In a valid bz2 header the 4th byte is in the range b'1' ... b'9'.
+
     elif HAVE_BZ2 and magic[:3] == b"BZh" and 0x31 <= magic[3] <= 0x39:
```
Horofic, line 506:
```suggestion
    if HAVE_BZ2 and magic[:3] == b"BZh" and 0x31 <= magic[3] <= 0x39:
```

dissect/target/helpers/fsutil.py
```diff
@@ -480,13 +502,23 @@ def open_decompress(
...
+        # In a valid bz2 header the 4th byte is in the range b'1' ... b'9'.
         return bz2.open(file, mode, encoding=encoding, errors=errors, newline=newline)
-    else:
+
+    elif HAVE_ZSTD and magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
+        # stream_reader is not seekable, so we have to resort to the less
+        # efficient decompressor which returns bytes.
+        return io.BytesIO(zstandard.decompress(file.read()))
+
+    elif path:
```
Horofic, line 515:
```suggestion
    if HAVE_ZSTD and magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
        # stream_reader is not seekable, so we have to resort to the less
        # efficient decompressor which returns bytes.
        return io.BytesIO(zstandard.decompress(file.read()))

    if path:
```

dissect/target/helpers/fsutil.py
```diff
@@ -480,13 +502,23 @@ def open_decompress(
...
         file.close()
         return path.open(mode, encoding=encoding, errors=errors, newline=newline)
 
+    else:
+        return file
+
```
Horofic, line 521:
```suggestion
    return file

```

Horofic (changes requested):

Horofic (dismissed):

Horofic (approved):

Schamper (approved):

PR #532, opened 2024-02-08
Add mount by LABEL= for ext filesystems

Miauwkeru:
(DIS-2894)

### DIFF
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -203,32 +203,29 @@ class UnixPlugin(OSPlugin):
                 fs_subvol = None
                 fs_subvolid = None
                 fs_volume_name = fs.volume.name if fs.volume and not isinstance(fs.volume, list) else None
-                last_mount = None
+                fs_last_mount = None
 
-                if dev_id:
-                    if fs.__type__ == "xfs":
-                        fs_id = fs.xfs.uuid
-                    elif fs.__type__ == "ext":
-                        fs_id = fs.extfs.uuid
-                        last_mount = fs.extfs.last_mount
-                    elif fs.__type__ == "btrfs":
-                        fs_id = fs.btrfs.uuid
-                        fs_subvol = fs.subvolume.path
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/unix/_os.py
```diff
@@ -222,6 +222,9 @@ def _add_mounts(self) -> None:
                         # but instead a volume_id which is not case-sensitive
                         fs_id = fs_id[:4].upper() + "-" + fs_id[4:].upper()
 
+                if fs.__type__ == "ext":
```
Schamper, line 225: Can't comment that far up, but might be nice to refactor the code above it a bit too look like this:

```python
                if fs.__type__ == "xfs":
                    fs_id = fs.xfs.uuid
                elif fs.__type__ == "ext":
                    fs_id = fs.extfs.uuid
                    fs_volume_name = fs.extfs.volume_name
                    last_mount = fs.extfs.last_mount
                elif fs.__type__ == "btrfs":
                    fs_id = fs.btrfs.uuid
                    fs_subvol = fs.subvolume.path
                    fs_subvolid = fs.subvolume.objectid
                elif fs.__type__ == "fat":
                    fs_id = fs.fatfs.volume_id
                    # This normalizes fs_id to comply with libblkid generated UUIDs
                    # This is needed because FAT filesystems don't have a real UUID,
                    # but instead a volume_id which is not case-sensitive
                    fs_id = fs_id[:4].upper() + "-" + fs_id[4:].upper()
```
And maybe also rename `last_mount` to `fs_last_mount`.

Schamper (changes requested):

dissect/target/plugins/os/unix/_os.py
Schamper: I'd keep it simpler and just add different fields/properties to the base `Filesystem` class. Similarly to how volume  systems have a serial. But that's indeed not something for this PR.

Schamper (approved):

cecinestpasunepipe (approved):

PR #538, opened 2024-02-14
Add Brave browser plugin

JSCU-CNI:
This PR adds a `BrowserPlugin` for Brave. 

While working on this PR we found that cookies are stored in the `Network` subfolder in newer versions of Chromium-based browsers.  We therefore made a small addition to `_iter_db` to support checking additional subfolders, which can then be used in the `cookies` method.

### DIFF
--- a/dissect/target/plugins/apps/browser/brave.py
+++ b/dissect/target/plugins/apps/browser/brave.py
@@ -1,0 +1,69 @@
+from typing import Iterator
+
+from dissect.target.helpers.descriptor_extensions import UserRecordDescriptorExtension
+from dissect.target.helpers.record import create_extended_descriptor
+from dissect.target.plugin import export
+from dissect.target.plugins.apps.browser.browser import (
+    GENERIC_COOKIE_FIELDS,
+    GENERIC_DOWNLOAD_RECORD_FIELDS,
+    GENERIC_EXTENSION_RECORD_FIELDS,
+    GENERIC_HISTORY_RECORD_FIELDS,
+    BrowserPlugin,
+)
+from dissect.target.plugins.apps.browser.chromium import (
+    CHROMIUM_DOWNLOAD_RECORD_FIELDS,
+    ChromiumMixin,
+)
+
+
+class BravePlugin(ChromiumMixin, BrowserPlugin):
+    """Brave browser plugin."""
+
+    __namespace__ = "brave"
+
+    DIRS = [
+        # Windows
+        "AppData/Local/BraveSoftware/Brave-Browser/User Data/Default",
+        "AppData/Roaming/BraveSoftware/Brave-Browser/User Data/Default",
+        # Linux
+        ".config/BraveSoftware/Default",
+        # Macos
+        "Library/Application Support/BraveSoftware/Default",
+    ]
+
+    BrowserHistoryRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/brave/history", GENERIC_HISTORY_RECORD_FIELDS
+    )
+
+    BrowserCookieRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/brave/cookie",
+        GENERIC_COOKIE_FIELDS,
+    )
+
+    BrowserDownloadRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/brave/download", GENERIC_DOWNLOAD_RECORD_FIELDS + CHROMIUM_DOWNLOAD_RECORD_FIELDS
+    )
+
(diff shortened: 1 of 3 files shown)


dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -69,11 +70,12 @@ def _build_userdirs(self, hist_paths: list[str]) -> list[tuple[UserDetails, Targ
                 users_dirs.append((user_details.user, cur_dir))
         return users_dirs
 
-    def _iter_db(self, filename: str) -> Iterator[SQLite3]:
+    def _iter_db(self, filename: str, *subfolders: Optional[list[str]]) -> Iterator[SQLite3]:
```
Schamper, line 73:
```suggestion
    def _iter_db(self, filename: str, subdirs: Optional[list[str]] = None) -> Iterator[SQLite3]:
```
I don't think using `*` is the nicest approach here. I feel like an explicit list of sub directories is better suited.

I've also renamed it to use the `dir` name, mostly for consistency with the other methods in this file.

dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -69,11 +70,12 @@ def _build_userdirs(self, hist_paths: list[str]) -> list[tuple[UserDetails, Targ
...
         """Generate a connection to a sqlite database file.
 
         Args:
             filename: The filename as string of the database where the data is stored.
+            subfolders (optional): Subfolder(s) to also try for every configured directory.
```
Schamper, line 78:
```suggestion
            subdirs: Subdirectories to also try for every configured directory.
```

dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -83,7 +85,11 @@ def _iter_db(self, filename: str) -> Iterator[SQLite3]:
             SQLError: If the history file could not be opened.
         """
 
-        for user, cur_dir in self._build_userdirs(self.DIRS):
+        dirs = self.DIRS
+        if subfolders:
+            dirs = dirs + [join(dir, subdir) for dir, subdir in itertools.product(self.DIRS, subfolders)]
```
Schamper, line 90:
```suggestion
        if subdirs:
            dirs.extend([join(dir, subdir) for dir, subdir in itertools.product(self.DIRS, subdirs)])
```

dissect/target/plugins/apps/browser/chromium.py
```diff
@@ -198,7 +204,7 @@ def cookies(self, browser_name: Optional[str] = None) -> Iterator[BrowserCookieR
                 is_http_only (bool): Cookie http only flag.
                 same_site (bool): Cookie same site flag.
         """
-        for user, db_file, db in self._iter_db("Cookies"):
+        for user, db_file, db in self._iter_db("Cookies", "Network"):
```
Schamper, line 207:
```suggestion
        for user, db_file, db in self._iter_db("Cookies", subdirs=["Network"]):
```

dissect/target/plugins/apps/browser/brave.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class BravePlugin(ChromiumMixin, BrowserPlugin):
+    """Brave browser plugin."""
+
+    __namespace__ = "brave"
+
+    DIRS = [
+        # Windows
+        "AppData/Local/BraveSoftware/Brave-Browser/User Data/Default",
```
Schamper, line 26: Could you please also add these paths to acquire? :smile:

Schamper (changes requested):

dissect/target/plugins/apps/browser/brave.py
JSCU-CNI: Added in https://github.com/fox-it/acquire/pull/129.

dissect/target/plugins/apps/browser/brave.py
```diff
@@ -0,0 +1,69 @@
...
+        "Library/Application Support/BraveSoftware/Default",
+    ]
+
+    BrowserHistoryRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/brave/history", GENERIC_HISTORY_RECORD_FIELDS
+    )
+
+    BrowserCookieRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+        "browser/brave/cookie",
+        GENERIC_COOKIE_FIELDS,
```
Schamper, line 40:
```suggestion
        "browser/brave/cookie", GENERIC_COOKIE_FIELDS
```

Schamper (changes requested):

Schamper (approved):

PR #543, opened 2024-02-19
Fix bug in defender evtx that resulted in ts field value being None

Poeloe:
(DIS-2561)


Schamper (changes requested):
Can you add a regression test?

Schamper (approved):

PR #544, opened 2024-02-19
Fix bug in WER plugin caused by special characters in field name

Poeloe:
(DIS-2507)


tests/_data/plugins/os/windows/wer/wer_test.wer
Schamper, line 1: Test file added but no additional unit test? :smile:

Poeloe: No, true! I added special character fields to the WER file and the test checks them all. So it's all covered by adding lines there ;)

Schamper: Ah it's just stuff added to an existing test file? GitHub interface doesn't show that nicely unfortunately.

Poeloe: Yeah, indeed. Added two additional values for the WER file. The following two:
```
ui[1]=test_ui1
SpécialCharactér=test_spcial_charactr
```
So the left side is the field as it is in the WER file. The right side is how it should be parsed by the plugin (minus `test_`)

dissect/target/plugins/os/windows/wer.py
```diff
@@ -87,9 +87,14 @@ def _create_record_descriptor(record_name: str, record_fields: list[tuple[str, s
...
 
 
-def _key_to_snake_case(key: str) -> str:
+def _sanitize_key(key: str) -> str:
+    # Convert camel case to snake case
     for pattern in camel_case_patterns:
         key = pattern.sub(r"\1_\2", key)
+
+    # Keep only basic characters in key
+    key = re.sub(r"[^a-zA-Z0-9_]", "", key)
```
cecinestpasunepipe, line 96: I am not sure how important this is but I foresee two possibly unwanted results:

1. The key becomes empty ("невидимый")
2. Or you might end up with duplicate keys (áErreur, éErreur)*
* 'lowercasing' might also cause this

If that is no problem, then it does not matter of course.

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/windows/wer.py
Poeloe: True! Changed the code a bit around to be able to warn the user when these edge cases happen. I chose for a warning, because I don't think it's worth it to fully handle these edge cases. But please let me know if you think differently about this.

cecinestpasunepipe (approved):

PR #549, opened 2024-02-20
IIS plugin does not process logs in default dir without ApplicationHo…

cecinestpasunepipe:
…st config file (DIS-2795)

### DIFF
--- a/dissect/target/plugins/apps/webserver/iis.py
+++ b/dissect/target/plugins/apps/webserver/iis.py
@@ -96,6 +96,7 @@ class IISLogsPlugin(WebserverPlugin):
 
         for log_path in self.DEFAULT_LOG_PATHS:
             try:
+                # later on we use */*.log to collect the files, so we need to move up 2 levels
                 log_dir = self.target.fs.path(log_path).parents[1]
             except IndexError:
                 continue
--- a/dissect/target/plugins/apps/webserver/iis.py
+++ b/dissect/target/plugins/apps/webserver/iis.py
@@ -106,8 +106,8 @@ class IISLogsPlugin(WebserverPlugin):
                 log_paths.add(("auto", log_dir))
                 continue
 
-            for _log_dir in self.target.fs.glob(str(log_dir)):
-                if not _log_dir.is_dir():
+            for _log_dir_str in self.target.fs.glob(str(log_dir)):
+                if not (_log_dir := self.target.fs.path(_log_dir_str)).is_dir():
                     continue
                 log_paths.add(("auto", _log_dir))
 
--- a/dissect/target/plugins/apps/webserver/iis.py
+++ b/dissect/target/plugins/apps/webserver/iis.py
@@ -107,7 +107,7 @@ class IISLogsPlugin(WebserverPlugin):
                 continue
 
             for _log_dir in self.target.fs.glob(str(log_dir)):
-                if not (_log_dir := self.target.fs.path(_log_dir)).is_dir():
+                if not _log_dir.is_dir():
                     continue
                 log_paths.add(("auto", _log_dir))
 
--- a/dissect/target/plugins/apps/webserver/iis.py
+++ b/dissect/target/plugins/apps/webserver/iis.py
@@ -54,17 +54,26 @@ class IISLogsPlugin(WebserverPlugin):
 
     APPLICATION_HOST_CONFIG = "sysvol/windows/system32/inetsrv/config/applicationHost.config"
 
+    DEFAULT_LOG_PATHS = [
+        "sysvol\\Windows\\System32\\LogFiles\\W3SVC*\\*.log",
+        "sysvol\\Windows.old\\Windows\\System32\\LogFiles\\W3SVC*\\*.log",
+        "sysvol\\inetpub\\logs\\LogFiles\\*.log",
+        "sysvol\\inetpub\\logs\\LogFiles\\W3SVC*\\*.log",
+        "sysvol\\Resources\\Directory\\*\\LogFiles\\Web\\W3SVC*\\*.log",
+    ]
+
     __namespace__ = "iis"
 
     def __init__(self, target):
         super().__init__(target)
         self.config = self.target.fs.path(self.APPLICATION_HOST_CONFIG)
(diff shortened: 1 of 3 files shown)


dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -84,11 +93,23 @@ def get_log_dirs(self) -> list[tuple[str, Path]]:
         except (ElementTree.ParseError, DissectFileNotFoundError) as e:
             self.target.log.warning(f"Error while parsing {self.config}: {e}")
 
+        for log_path in self.DEFAULT_LOG_PATHS:
+            for log_file in self.target.fs.glob(log_path):
+                try:
+                    log_dir = self.target.fs.path(log_file).parents[1]
```
Poeloe, line 99: Since this is the only place you use `DEFAULT_LOG_PATHS`, I don't really get why you need to get `parents[1]`? Why not just specify two dirs "higher" for each path? 

Or do you use it to check if there are any files in that directory at all? Maybe then it would be nice to add a small comment saying so.

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -84,11 +93,23 @@ def get_log_dirs(self) -> list[tuple[str, Path]]:
...
+                    if (
+                        "auto",
+                        log_dir,
+                    ) not in log_paths:
+                        log_paths.append(("auto", log_dir))
```
Poeloe, line 104: Maybe let `log_paths` be a set instead of a list, such that you don't need this if-statement

Poeloe (changes requested):

dissect/target/plugins/apps/webserver/iis.py
cecinestpasunepipe: This is because we use the acquire glob patterns, which may contain wildcards at arbitrary path segments "a/*/b" "a/b/c/*/*/"  etc. So, to be able to reuse the, we basically glob the files and then inquire about the parent folder. Otherwise you would have to create several recursive loops, or something like that.

dissect/target/plugins/apps/webserver/iis.py
cecinestpasunepipe: ok, good one

dissect/target/plugins/apps/webserver/iis.py
Poeloe: But that's only for the last item in the list, the others are all paths without a glob (if you see it from parents[1], which is the `LogFiles` directory). In my opinion, it now iterates those paths unnecessarily. 

Maybe check for `*` in the path entry and then do the logic below. Else just check if the path exists and add it?

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -54,41 +55,67 @@ class IISLogsPlugin(WebserverPlugin):
...
 
         except (ElementTree.ParseError, DissectFileNotFoundError) as e:
             self.target.log.warning(f"Error while parsing {self.config}: {e}")
 
-        return log_paths
+        for log_path in self.DEFAULT_LOG_PATHS:
+            try:
+                log_dir = self.target.fs.path(log_path).parents[1]
+            except IndexError:
+                continue
```
Poeloe, line 101: Since it is highly unlikely that this happens (we define `DEFAULT_LOG_PATHS` ourselves), I would suggest to log an error here such that we're early notified about a possible bug in the code.

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -54,41 +55,67 @@ class IISLogsPlugin(WebserverPlugin):
...
+            if not has_glob_magic(str(log_dir)) and log_dir.exists():
+                log_paths.add(("auto", log_dir))
+                continue
+
+            for log_file in self.target.fs.glob(log_path):
+                try:
+                    log_dir = self.target.fs.path(log_file).parents[1]
+                    log_paths.add(("auto", log_dir))
+                except IndexError:
+                    pass
```
Poeloe, line 112: You already have the parent dir that you want in `log_dir`, so you only need to glob this pattern and add all the entries (because they are all valid if they exist). the `iter_log_format_path_pairs` method handles the retrieval of the proper `.log` afterwards.
 
```suggestion
            for log_file in self.target.fs.glob(log_dir):
                log_paths.add(("auto", log_dir))
```

Poeloe (changes requested):

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -54,41 +55,67 @@ class IISLogsPlugin(WebserverPlugin):
...
-                    log_paths.append((log_format, log_dir))
+                    log_paths.add((log_format, log_dir))
 
         except (ElementTree.ParseError, DissectFileNotFoundError) as e:
             self.target.log.warning(f"Error while parsing {self.config}: {e}")
 
-        return log_paths
+        for log_path in self.DEFAULT_LOG_PATHS:
+            try:
+                log_dir = self.target.fs.path(log_path).parents[1]
```
Poeloe, line 99: Maybe also provide a comment why you need `parents[1]`, since it's not trivial

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -54,41 +55,67 @@ class IISLogsPlugin(WebserverPlugin):
...
+            except IndexError:
+                self.target.log.error("Incompatible path found: %s", log_path)
+                continue
+
+            if not has_glob_magic(str(log_dir)) and log_dir.exists():
+                log_paths.add(("auto", log_dir))
+                continue
+
+            for _log_dir in self.target.fs.glob(str(log_dir)):
+                if not (_log_dir := self.target.fs.path(_log_dir)).is_dir():
```
Poeloe, line 110: _log_dir is already a TargetPath right? Then this should just work:

```suggestion
                if not _log_dir.is_dir():
```

Poeloe: Nope, sorry, my bad. I was thinking of a glob on a path. Glob on fs indeed returns strings!

Poeloe (approved):
LGTM!

PR #554, opened 2024-02-23
add tests for symlink logic MappedFile

cecinestpasunepipe:
(DIS-1796)


cecinestpasunepipe:
Current behaviour is ok. Attempt to change it breaks. Added unittests to confirm behaviour and protect against regressions.

Poeloe (approved):
LGTM! Only commit message needs to be changed ;)

Issue #557, opened 2024-02-27
Drive letter is not rego


Zawadidone:
During an investigation I observed that some data types were not shown in a timeline produced by Dissect, see the browser plugin as an example of this:

```
# Dissect Target version 3.15
target-query -t <TARGET> -f browser
[...]
# The plugin works and shows result

# Dissect Target version 3.16.dev4 (https://github.com/fox-it/dissect.target/commit/4b0c7ccb78b8ec47e80ca376cfe5ff543ef90984)
target-query -t <TARGET> -f browser
# The plugin does not work
```

Using `target-shell` the output shows that when using version `3.16.dev4`, that the drive letter `C:` is not recognized. Because of that every plugins that uses the function `target.user_details.all_with_home()` will not work depending on which loader is used.

```bash
# Dissect Target version 3.15
In [1]: print(t.fs.path("sysvol/Users/<username>").exists())
True

In [2]: print(t.fs.path("C:/Users/<username>").exists())
True

# Dissect Target version 3.16.dev4 (https://github.com/fox-it/dissect.target/commit/4b0c7ccb78b8ec47e80ca376cfe5ff543ef90984)

In [1]: print(t.fs.path("sysvol/Users/<username>").exists())
True

In [2]: print(t.fs.path("C:/Users/<username>").exists())
False
```

Schamper:
What kind of target are you using?

I think I know what's breaking, and have an idea for a fix. Will try it out tomorrow.

Zawadidone:
Velociraptor Windows collection created with the following configuration https://github.com/fox-it/dissect.target/pull/558/files#diff-d21a3f7b5a8a12a1531cd0193f1d5d28eed550ad78782baaf78ce79f25b66631R63-R67.

PR #560, opened 2024-03-01
Mount volumes to drive letters in Velociraptor loader

Schamper:
Fixes https://github.com/fox-it/dissect.target/issues/557

@Zawadidone can you confirm whether my assumption with regards to the drive letter directory name is correct?


Zawadidone:
@Schamper yes it works!

pyrco (approved):

PR #562, opened 2024-03-05
Add all test files to git LFS

Schamper:
The decision on when to include a file in LFS was getting a bit annoying, and sometimes we forget. Just put it all in LFS.

We should update either the README or the docs at some point to include a small explanation on LFS too.


Schamper:
Hm this does remove the ability, to preview a (text based) test file in a PR, which is also very annoying.

pyrco:
> Hm this does remove the ability to preview a (text based) test file in a PR, which is also very annoying.

That's an unfortunate side effect. Putting `test/*` in LFS is right thing to do I think. This is a shortcoming of GitHub and I'm leaning to not letting that stand in the way of the this.

Schamper:
> > Hm this does remove the ability to preview a (text based) test file in a PR, which is also very annoying.
> 
> That's an unfortunate side effect. Putting `tests/_data/*` in LFS is right thing to do I think. This is a shortcoming of GitHub and I'm leaning to not letting that stand in the way of the this.

Yes, agreed.

pyrco (approved):

pyrco (approved):

PR #541, opened 2024-02-16
Add password extraction to browser plugins

JSCU-CNI:
This PR adds support for extracting saved passwords and decrypting cookie values from Chromium-based browsers and Firefox on both Windows and Linux targets.

This PR also adds support for decrypting DPAPI user secrets as the current DPAPI implementation only accounts for system secrets.

PR #566, opened 2024-03-05
Add windows credhist plugin

JSCU-CNI:
This PR adds support for decrypting CREDHIST entries on Windows targets. 

Currently supports decrypting full and partial CREDHIST chains and can tell you if the decryption worked or not. Does not (yet) provide decrypted SHA1 hashes to the DPAPI plugin. Perhaps @cobyge can tell us where it's best to implement this.


dissect/target/plugins/os/windows/dpapi/credhist.py
```diff
@@ -0,0 +1,134 @@
...
+        for user in self.target.user_details.all_with_home():
+            for path in ["AppData/Roaming/Microsoft/Protect/CREDHIST", "Application Data/Microsoft/Protect/CREDHIST"]:
+                if (credhist_path := user.home_path.joinpath(path)).exists():
+                    yield credhist_path
+
+    def check_compatible(self) -> None:
+        if not any(self.files):
+            raise UnsupportedPluginError("No CREDHIST files found on target")
+
+    def all(self) -> Iterator[CredHistFile]:
```
Schamper, line 131: Would it be nice to add an export plugin to dump some of the info to a record? Can also use the keychain for that (with the changes from the browser password PR). As well as maybe an additional internal function to read a file from just a single user provided by argument.

dissect/target/plugins/os/windows/dpapi/credhist.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+        sha_size = self.raw.dwShaHashSize
+        nt_size = self.raw.dwNtHashSize
+
+        self.hash_sha = data[:sha_size].hex()
+        self.hash_nt = data[sha_size : sha_size + nt_size].rstrip(b"\x00").hex()
+
+
+class CredHistFile:
```
Schamper, line 68: I'm not feeling the current setup with this class and the dataclass just yet, but I also couldn't think of a better way right now. Just noting my thoughts in case maybe someone else has a better idea.

Schamper (changes requested):

Schamper (changes requested):

Schamper:
@cobyge option 1 is at least preferred for providing the "initial"/current password, as is currently in review in #541. While it would be nice, I don't immediately see a nice way how CREDHIST could provide passwords to the keychain (it would need to be invoked somehow). You could make something similar to how we have "child plugins", i.e. "keychain plugins" and bind a keychain to a target (instead of being global right now). But that feels a bit overkill, at least for the moment where we would only have a single plugin for it.

I'm more than happy to hear some ideas for this, but for complexity sake, option 2 might otherwise be preferred.

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+    decrypted: bool = False
+
+    def decrypt(self, password_hash: bytes) -> None:
+        """Decrypt this CREDHIST entry using the provided password hash. Modifies ``CredHistEntry.sha1``
+        and ``CredHistEntry.nt`` values.
+
+        If the decrypted ``nt`` value is 32 characters we assume the decryption was successful.
+
+        Args:
+            password_hash: bytes of SHA1 password hash digest
```
Schamper, line 79:
```suggestion
            password_hash: Bytes of SHA1 password hash digest.
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+                    _user=user,
+                    _target=self.target,
+                )
+
+
+def keychain_passwords() -> set:
+    passphrases = set()
+    for key in keychain.get_keys_for_provider("user") + keychain.get_keys_without_provider():
+        if key.key_type == keychain.KeyType.PASSPHRASE:
+            passphrases.add(key.value)
```
Schamper, line 210: Just curious, no empty password possibility here?

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+        if isinstance(password_hash, str):
+            password_hash = bytes.fromhex(password_hash)
+
+        for entry in reversed(self.entries):
+            try:
+                entry.decrypt(password_hash)
+            except ValueError as e:
+                log.warning("Could not decrypt entry %s with password %s", entry.guid, password_hash.hex())
+                log.debug("", exc_info=e)
+                continue
```
Schamper, line 150: Took me a bit to really dig into what's happening here. You call `CredHistFile.decrypt` multiple times with different passwords. For some of those, encryption will work, for some it will fail. I assume you keep on trucking on an error in the case that you may have a password in your set that will successfully decrypted a _later_ password? (Does that even make sense? Presumably once you successfully decrypt the first in the chain, the rest should decrypt just fine?)

Wouldn't a flow like this make it a bit more clear what's happening (plus a comment about this behaviour in the docstring):
```
for entry in reversed(self.entries):
    if not entry.decrypted:
        entry.decrypt(password_hash)  # Let it raise the error - decryption of the file failed
    password_hash = entry.sha1
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+c_credhist.load(credhist_def)
+
+
+@dataclass
+class CredHistEntry:
+    version: int
+    guid: str
+    user_sid: str
+    sha1: Optional[str]
+    nt: Optional[str]
```
Schamper, line 66:
```suggestion
    sha1: Optional[bytes]
    nt: Optional[bytes]
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+    hash_alg: HashAlgorithm = field(repr=False)
+    cipher_alg: CipherAlgorithm = field(repr=False)
+    raw: c_credhist.entry = field(repr=False)
+    decrypted: bool = False
+
+    def decrypt(self, password_hash: bytes) -> None:
+        """Decrypt this CREDHIST entry using the provided password hash. Modifies ``CredHistEntry.sha1``
+        and ``CredHistEntry.nt`` values.
+
+        If the decrypted ``nt`` value is 32 characters we assume the decryption was successful.
```
Schamper, line 76:
```suggestion
        If the decrypted ``nt`` value is 16 bytes we assume the decryption was successful.
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+            iv=self.raw.pSalt,
+            hash_algorithm=self.hash_alg,
+            rounds=self.raw.dwPbkdf2IterationCount,
+        )
+
+        sha_size = self.raw.dwShaHashSize
+        nt_size = self.raw.dwNtHashSize
+
+        sha1 = data[:sha_size].hex()
+        nt = data[sha_size : sha_size + nt_size].rstrip(b"\x00").hex()
```
Schamper, line 96:
```suggestion
        sha1 = data[:sha_size]
        nt = data[sha_size : sha_size + nt_size].rstrip(b"\x00")
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if len(nt) != 32:
```
Schamper, line 98:
```suggestion
        if len(nt) != 16:
```

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+                    nt=None,
+                    raw=entry,
+                )
+        except EOFError:
+            pass
+
+    def decrypt(self, password_hash: bytes | str) -> None:
+        """Decrypt a CREDHIST chain using the provided password SHA1 hash."""
+        if isinstance(password_hash, str):
+            password_hash = bytes.fromhex(password_hash)
```
Schamper, line 142:
```suggestion
    def decrypt(self, password_hash: bytes) -> None:
        """Decrypt a CREDHIST chain using the provided password SHA1 hash."""
```

Schamper: Can you also add an appropriate docstring for the `password_hash` argument.

dissect/target/plugins/os/windows/credhist.py
```diff
@@ -0,0 +1,211 @@
...
+
+            for password in passwords:
+                credhist.decrypt(hashlib.sha1(password.encode("utf-16-le")).digest())
+
+            for entry in credhist.entries:
+                yield CredHistRecord(
+                    guid=entry.guid,
+                    decrypted=entry.decrypted,
+                    sha1=entry.sha1,
+                    nt=entry.nt,
```
Schamper, line 200:
```suggestion
                    sha1=entry.sha1.hex(),
                    nt=entry.nt.hex(),
```

Schamper (changes requested):
Can you also rebase onto main.

dissect/target/plugins/os/windows/credhist.py
Schamper: That's a very good point, looks like I wasn't paying enough attention after all. I was very confused on the `password_hash = entry.sha1`  line below it, but for some reason I was completely blind to the `continue` in the `except` branch.

Apologies :wink:

Schamper (approved):

PR #567, opened 2024-03-06
Add lru_cache for UsersPlugin's find method

Poeloe:


dissect/target/plugins/general/users.py
```diff
@@ -1,10 +1,13 @@
+from functools import lru_cache
 from typing import Generator, NamedTuple, Optional, Union
 
 from dissect.target.exceptions import UnsupportedPluginError
 from dissect.target.helpers.fsutil import TargetPath
 from dissect.target.helpers.record import UnixUserRecord, WindowsUserRecord
 from dissect.target.plugin import InternalPlugin
 
+from ... import Target
```
Schamper, line 9: Can you make this an absolute import?

Schamper (changes requested):

dissect/target/plugins/general/users.py
Poeloe: Oh, I missed that one. Unwanted side effects from letting your IDE create the init ;)

dissect/target/plugins/general/users.py
```diff
@@ -16,6 +18,10 @@ class UsersPlugin(InternalPlugin):
 
     __namespace__ = "user_details"
 
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self.find = lru_cache(4096)(self.find)
```
Schamper, line 23:
```suggestion
        self.find = lru_cache(64)(self.find)
```
Shouldn't matter too much, but maybe a more sensible amount? Or maybe even 32?

Schamper (changes requested):

Schamper (approved):

PR #570, opened 2024-03-07
Amend documentation for `none` plugin export type

Poeloe:
(DIS-2088)

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -82,7 +82,7 @@ def export(*args, **kwargs) -> Callable:
         - default: Single return value
         - record: Yields records. Implicit when record argument is given.
         - yield: Yields printable values.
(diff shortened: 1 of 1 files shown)


dissect/target/plugin.py
```diff
@@ -82,7 +82,7 @@ def export(*args, **kwargs) -> Callable:
         - default: Single return value
         - record: Yields records. Implicit when record argument is given.
         - yield: Yields printable values.
-        - none: No return value.
+        - none: No return value. Plugin is responsible for output formatting and should return None.
```
Schamper, line 85:
```suggestion
        - none: No return value. Plugin is responsible for output formatting and should return ``None``.
```

Schamper (changes requested):

Schamper (approved):

PR #571, opened 2024-03-09
Add disc filesystem

MaxGroot:
Depends on https://github.com/fox-it/dissect.disc/pull/1.


MaxGroot:
While testing against some ISOs, I ran into something interesting with this one: https://cdimage.debian.org/debian-cd/current/amd64/iso-dvd/debian-12.6.0-amd64-DVD-1.iso

This disc is both a FAT and an ISO9660 filesystem, and uses the system use area of the disc for volume information. Dissect recognizes this MBR and opens the file-handle as a volume. However, the file-handle is _also_ an ISO9660 disc, the information of which is lost. I'm not too familiar with the volume system of dissect, but playing around I came across this part of the volume system:

```python3
# Some LVM configurations (i.e. RAID with the metadata at the end of the disk)
            # may be misidentified as having a valid MBR/GPT on some of the disks
            # To counter this, first check if the disk is part of any LVM configurations that we support
            if not volume.is_lvm_volume(disk):
```

(https://github.com/fox-it/dissect.target/blob/main/dissect/target/target.py#L717)

If i add an `or true` at the end of this line, both the FAT and ISO9660 filesystems are added to the target. It looks to me as though we would need to edit the volume opening logic to 're-submit' a file handle after having opened it, and checking if other filesystems other than the one we already have, detect it. Is this correct? I don't mind fidgetting in the volume opening logic a bit, but would like to hear what you think before making too drastic changes just to support this edge case.

PR #575, opened 2024-03-14
Create separate LayerFilesystem from RootFilesystem

Schamper:


tests/test_filesystem.py
```diff
@@ -1164,3 +1165,37 @@ def test_virtual_filesystem_map_file_from_tar() -> None:
...
+    lfs.add_fs_layer(vfs3)
+    assert lfs.path("file1").read_text() == "value1"
+    assert lfs.path("file2").read_text() == "value2"
+    assert lfs.path("file3").read_text() == "value3"
+
+    lfs.add_fs_layer(vfs4)
+    assert lfs.path("file1").read_text() == "value1"
+    lfs.remove_fs_layer(vfs4)
+    lfs.prepend_fs_layer(vfs4)
+    assert lfs.path("file1").read_text() == "value4"
```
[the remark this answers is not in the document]
Schamper: Maybe. I've currently kept it this way because that's how it behaved before. I'll throw it up for opinion within the team.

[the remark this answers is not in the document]
Schamper: After a short but powerful vote we've indeed decided that it's best to reverse this now. We'll need to perform some testing to make sure everything works as expected for ESXi and FortiOS (two heavy users of the layers). Do you still have some FortiOS test images that you could perhaps test these changes against?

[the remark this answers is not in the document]
Schamper: Yeah, syntax is the same, just the resolution order is changed. I didn't expect anything to break, but always good to verify!

dissect/target/filesystem.py
```diff
@@ -1326,45 +1346,92 @@ def mount(self, path: str, fs: Filesystem) -> None:
...
+        self.add_fs_layer(layer)
         return layer
 
+    def prepend_layer(self, **kwargs) -> VirtualFilesystem:
+        """Prepend a new layer."""
+        layer = VirtualFilesystem(case_sensitive=self.case_sensitive, alt_separator=self.alt_separator, **kwargs)
+        self.prepend_fs_layer(layer)
+        return layer
+
+    def add_fs_layer(self, fs: Filesystem) -> None:
```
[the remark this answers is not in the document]
Horofic: This kind of reflects the intent better, but is no longer consistent with `add_layer()` naming.

dissect/target/filesystem.py
```diff
@@ -1326,45 +1346,92 @@ def mount(self, path: str, fs: Filesystem) -> None:
...
+        del self._root_entry.entries[idx]
+
     @property
     def case_sensitive(self) -> bool:
+        """Whether the filesystem is case sensitive."""
         return self._case_sensitive
 
     @property
     def alt_separator(self) -> str:
+        """The alternative separator for the filesystem."""
```
Horofic, line 1414:
```suggestion
        """The alternative separator of the filesystem."""
```

dissect/target/filesystem.py
```diff
@@ -1326,45 +1346,92 @@ def mount(self, path: str, fs: Filesystem) -> None:
...
     def case_sensitive(self, value: bool) -> None:
+        """Set the case sensitivity of the filesystem (and all layers)."""
         self._case_sensitive = value
         self.root.case_sensitive = value
         for layer in self.layers:
             layer.case_sensitive = value
 
     @alt_separator.setter
     def alt_separator(self, value: str) -> None:
+        """Set the alternative separator for the filesystem (and all layers)."""
```
Horofic, line 1427:
```suggestion
        """Set the alternative separator of the filesystem (and all layers)."""
```

dissect/target/filesystem.py
```diff
@@ -1405,11 +1473,11 @@ def _get_from_entry(self, path: str, entry: FilesystemEntry) -> FilesystemEntry:
 class EntryList(list):
     """Wrapper list for filesystem entries.
 
-    Expose a getattr on a list of items. Useful in cases where
-    there's a virtual filesystem entry as well as a real one.
+    Expose a getattr on a list of items. Useful to access internal objects from specific filesystem implementations.
```
Horofic, line 1476:
```suggestion
    Expose a :func:`getattr` on a list of items. Useful to access internal objects from specific filesystem implementations.
```

dissect/target/filesystem.py
```diff
@@ -1475,7 +1533,6 @@ def iterdir(self) -> Iterator[str]:
                 yielded.add(name)
 
     def scandir(self) -> Iterator[FilesystemEntry]:
```
Horofic, line 1535:
```suggestion
    def scandir(self) -> Iterator[LayerFilesystemEntry]:
```

Horofic (changes requested):

dissect/target/filesystem.py
```diff
@@ -1388,9 +1455,10 @@ def get(self, path: str, relentry: FilesystemEntry = None) -> FilesystemEntry:
                 raise NotASymlinkError(full_path)
             raise FileNotFoundError(full_path)
 
-        return RootFilesystemEntry(self, full_path, entries)
+        return LayerFilesystemEntry(self, full_path, entries)
 
-    def _get_from_entry(self, path: str, entry: FilesystemEntry) -> FilesystemEntry:
+    def _get_from_entry(self, path: str, entry: FilesystemEntry) -> LayerFilesystemEntry:
```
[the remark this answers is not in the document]
Schamper: This was intentional to enhance type-hinting where this is used.

dissect/target/filesystem.py
```diff
@@ -614,7 +614,7 @@ def walk_ext(
         onerror: Optional[Callable] = None,
         followlinks: bool = False,
```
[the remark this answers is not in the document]
Schamper: This function is modelled against `os.walk` (https://docs.python.org/3/library/os.html#os.walk). The base `Filesystem` stuff is all purposefully designed to closely resemble the functions in `os`. `TargetPath` is designed to closely resemble `pathlib.Path`

dissect/target/filesystem.py
Schamper: I'm not sure it works like that. `:func:` and friends are to deeplink to functions and classes within your code base. This is a Python native function. This would require `sphinx.ext.intersphinx` (https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) which we don't have configured for our docs.

Horofic: You are right indeed. In that case \`\`getattr\`\` should do it instead. While playing around a little I noticed that :func:\`getattr\` renders to `getattr()`, with added parenthesises, but obviously no deeplink. It might be an idea to stick with the :func: notation, if we want to ever add the Intersphinx extention in the future.

Schamper: Went in a different direction.

Horofic (approved):

PR #651, opened 2024-03-25
Change socket tests to use NetSocket plugin

Horofic:


tests/plugins/os/unix/linux/test_sockets.py
```diff
@@ -11,182 +11,126 @@ def test_sockets_plugin(target_linux_users: Target, fs_linux_proc_sockets: Virtu
 
 def test_tcp(target_linux_users: Target, fs_linux_proc_sockets: VirtualFilesystem):
     target_linux_users.add_plugin(ProcPlugin)
```
Schamper, line 13: Is adding the `ProcPlugin` here strictly necessary? Or otherwise maybe add both `ProcPlugin` and `SocketsPlugin` explicitly.

Schamper (changes requested):

Schamper (approved):

PR #654, opened 2024-03-27
Add libvirt loader and QEMU child plugin

Miauwkeru:


dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,29 @@
...
+        path = path.resolve()
+
+        super().__init__(path)
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        if path.suffix.lower() != ".xml":
+            return False
+
+        return "<domain>" in path.read_text().lower()
```
Schamper, line 23: Is this the first element? Maybe better to limit how many bytes you read to a sensible limit. You may accidentally read a .xml of several GB large.

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def map(self, target: Target) -> None:
+        xml_data = ElementTree.XML(self.path.read_text())
+        for disk in xml_data.findall("devices/disk/source"):
+            if file := disk.get("file"):
+                target.disk.add(container.open(file))
```
Schamper, line 29: Are these absolute paths? You should join on `self.path` regardless to ensure that you're reading the file from the same target (local `Path` or a bound `TargetPath`).

Schamper (changes requested):

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,31 @@
...
+
+        super().__init__(path)
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        if path.suffix.lower() != ".xml":
+            return False
+
+        with path.open("rb") as fh:
+            return b"<domain>" in fh.read(512)
```
Schamper, line 24: This doesn't seem to match the `qemu.xml` that you included in the test file. Also `<domain>` seems a little too generic, can you make this check more strict?

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def map(self, target: Target) -> None:
+        xml_data = ElementTree.XML(self.path.read_text())
+        for disk in xml_data.findall("devices/disk/source"):
+            if file := disk.get("file"):
+                target_file = self.path.joinpath(file)
```
Schamper, line 30: Another question - in a common libvirt setup, is there a relation between where the `.xml` file is and where the virtual disk files are? E.g. are they in the same directory?

I'm asking because sometimes we get a bunch of files in a single directory as "evidence". In that case, we may want to check if a file with the disk name exists in the same directory as the .xml file. The original absolute path is usually only checked as a last resort. This is also done in all similar loaders (`.vmx` etc)

tests/loaders/test_libvirt.py
```diff
@@ -0,0 +1,34 @@
...
+    assert LibvirtLoader.detect(xml_file)
+
+    xml_file = tmp_path.joinpath("base.xml")
+    xml_file.write_text("not a libvirt file")
+
+    assert LibvirtLoader.detect(xml_file) is False
+
+
+@patch(f"{LibvirtLoader.__module__}.container.open")
+def test_libvirt_map(mocked_container_open: Mock, target_bare: Target, fs_linux: VirtualFilesystem):
```
Schamper, line 27:
```suggestion
def test_libvirt_map(mocked_container_open: Mock, target_bare: Target, fs_linux: VirtualFilesystem) -> None:
```

tests/loaders/test_libvirt.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+from unittest.mock import Mock, patch
+
+from dissect.target import Target
+from dissect.target.filesystem import VirtualFilesystem
+from dissect.target.loaders.libvirt import LibvirtLoader
+from tests.conftest import absolute_path
+
+
+def test_libvirt_detection(target_bare: Target, tmp_path: Path) -> None:
```
Schamper, line 10: Why not also test against the actual `qemu.xml` you included as test file?

tests/plugins/child/test_qemu.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_compatible(target_linux: Target, fs_linux: VirtualFilesystem) -> None:
+    with pytest.raises(UnsupportedPluginError):
+        QemuChildTargetPlugin(target_linux).check_compatible()
+
+    fs_linux.map_file_fh("/etc/libvirt/qemu", None)
+    QemuChildTargetPlugin(target_linux).check_compatible()
+
+
+def test_list_children(target_linux: Target, fs_linux: VirtualFilesystem):
```
Schamper, line 18:
```suggestion
def test_list_children(target_linux: Target, fs_linux: VirtualFilesystem) -> None:
```

Schamper (changes requested):

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,46 @@
...
+            part_xml_data = fh.read(512).split(b"\n")
+            # From what I've seen, these are are always at the start of the file
+            # If its generated using virt-install
+            needles = [b"<domain", b"<name>", b"<uuid>"]
+
+            output = []
+            for needle in needles:
+                output.append(any(needle in line for line in part_xml_data))
+
+            return all(output)
```
Schamper, line 33:
```suggestion
            lines = fh.read(512).split(b"\n")
            # From what I've seen, these are are always at the start of the file
            # If its generated using virt-install
            needles = [b"<domain", b"<name>", b"<uuid>"]
            return all(any(needle in line for line in lines) for needle in needles)
```

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def map(self, target: Target) -> None:
+        xml_data = ElementTree.XML(self.path.read_text())
```
Schamper, line 36:
```suggestion
        xml_data = ElementTree.fromstring(self.path.read_text())
```
I think this is the more common way to use `ElementTree`?

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,46 @@
...
+        for disk in xml_data.findall("devices/disk/source"):
+            if not (file := disk.get("file")):
+                continue
+
+            path = Path(file)
+
+            for part in [path.name, file]:
+                if (target_file := self.base_dir.joinpath(part)).exists():
+                    target.disks.add(container.open(target_file))
```
Schamper, line 45:
```suggestion
                if (path := self.base_dir.joinpath(part)).exists():
                    target.disks.add(container.open(path))
```

dissect/target/loaders/libvirt.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def map(self, target: Target) -> None:
+        xml_data = ElementTree.XML(self.path.read_text())
+        for disk in xml_data.findall("devices/disk/source"):
+            if not (file := disk.get("file")):
+                continue
+
+            path = Path(file)
+
+            for part in [path.name, file]:
```
Schamper, line 43:
```suggestion
            for part in [fsutil.basename(path), file]:
```
I think it's nicer to use this here instead of a runtime OS biased `Path`.

dissect/target/plugins/child/qemu.py
```diff
@@ -0,0 +1,21 @@
...
+
+from typing import Iterator
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import ChildTargetRecord
+from dissect.target.plugin import ChildTargetPlugin
+
+
+class QemuChildTargetPlugin(ChildTargetPlugin):
+    """Child target that yields all domains from a kvm libvirt deamon"""
```
Schamper, line 11:
```suggestion
    """Child target plugin that yields all domains from a kvm libvirt deamon."""
```

dissect/target/plugins/child/qemu.py
```diff
@@ -0,0 +1,21 @@
...
+
+    __type__ = "qemu"
+
+    def check_compatible(self) -> None:
+        if not self.target.fs.path("/etc/libvirt/qemu").exists():
+            raise UnsupportedPluginError("Libvirt was not installed on this system.")
```
Schamper, line 17:
```suggestion
            raise UnsupportedPluginError("No libvirt installation found")
```

Schamper (approved):

PR #657, opened 2024-03-28
Fix gzip trailer_offset calculation in FortiFirmwareFile

yunzheng:
Also added a regression test

### DIFF
--- a/tests/containers/test_fortifw.py
+++ b/tests/containers/test_fortifw.py
@@ -57,6 +57,19 @@ def test_deobfuscate_firmware_file(header, is_gzipped):
     assert ff.size == len(FIRMWARE_HEADER)
 
 
+def test_gzip_trailer():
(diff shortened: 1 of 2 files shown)


tests/containers/test_fortifw.py
```diff
@@ -57,6 +57,19 @@ def test_deobfuscate_firmware_file(header, is_gzipped):
     assert ff.size == len(FIRMWARE_HEADER)
 
 
+def test_gzip_trailer():
```
Schamper, line 60:
```suggestion
def test_gzip_trailer() -> None:
```
Can you also add these to the other tests?

Schamper (changes requested):

tests/containers/test_fortifw.py
yunzheng: done

Schamper (approved):

PR #662, opened 2024-04-04
Fix log message calls in network manager

### LINKED ISSUE #661: Incorrect call to `self.target.log` in network manager `Template` class
Some log messages in the `Template` class incorrectly call `self.target.log` in a context where there is no `self.target`.

https://github.com/fox-it/dissect.target/blob/main/dissect/target/helpers/network_managers.py#L88

This breaks dissect when conditions of calling those log messages are met.
```
  File "/tmp/dissect.target/dissect/target/helpers/network_managers.py", line 88, in _parse_netplan_config
    self.target.log.error("Failed to parse %s. Cannot import ruamel.yaml", self.name)
    ^^^^^^^^^^^
AttributeError: 'Template' object has no attribute 'target'
```

JSCU-CNI:
This PR fixes #661.

### DIFF
--- a/dissect/target/helpers/network_managers.py
+++ b/dissect/target/helpers/network_managers.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import re
 from collections import defaultdict
--- a/dissect/target/helpers/network_managers.py
+++ b/dissect/target/helpers/network_managers.py
@@ -1,4 +1,5 @@
 import re
+import logging
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


Horofic (approved):

Horofic (dismissed):

dissect/target/helpers/network_managers.py
```diff
@@ -1,16 +1,19 @@
+import logging
```
Horofic, line 1:
```suggestion
from __future__ import annotations

import logging
```

Horofic (approved):

PR #668, opened 2024-04-05
Add JumpList plugin

Zawadidone:


Zawadidone:
I had some issue with Git, hopefully you didn't receive any Github notification because of that.

PR #672, opened 2024-04-09
Yield MFT segments in specified range

### LINKED ISSUE #665: Parse MFT records based on segment numbers
I would like to select which MFT records the MFT plugin outputs based on segment numbers. For example, `target-query TARGET -f mft --fs 1 --segments 0-9` executes the MFT plugin to output the first ten MFT records of the first NTFS filesystem of a target.

We use Dissect Target as part of an automated processing pipeline that has 15 CPU's available to execute plugins on a target using `xargs`. Of all the plugins the MFT plugin takes the longest of all plugins, this varies from a few minutes to a few hours. Because of that we would like to speed up the processing of MFT records.

I don't know if this should be part of Dissect Target or that we will just use a custom MFT plugin with some BASH commands.

Zawadidone:
Closes #665


dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -138,53 +145,62 @@ def mft(self, compact: bool = False):
...
-                continue
-
-            # If this filesystem is a "fake" NTFS filesystem, used to enhance a
-            # VirtualFilesystem, The driveletter (more accurate mount point)
-            # returned will be that of the VirtualFilesystem. This makes sure
-            # the paths returned in the records are actually reachable.
-            drive_letter = get_drive_letter(self.target, fs)
-            volume_uuid = get_volume_identifier(fs)
-
+        if fs:
```
Poeloe, line 148: Because the default is `-1`, if will by default evaluate to `True`. When the user passes `0`, it will evaluated to `False`.

I would also argue that the user should be able to use negative numbers, for example, when the users wants the last NTFS filesystem.

So long story short, I would suggest to set the default to None and check for `if fs is None` instead. Please then also change the type hint to `Optional[int]` for `fs`

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -138,53 +145,62 @@ def mft(self, compact: bool = False):
...
-                            )
-                    except Exception as e:
-                        self.target.log.warning("An error occured parsing MFT segment %d: %s", record.segment, str(e))
-                        self.target.log.debug("", exc_info=e)
-
-            except Exception:
-                log.exception("An error occured constructing FilesystemRecords")
+                filesystem = self.ntfs_filesystems[fs]
+            except IndexError:
+                self.target.log.exception("%s does not have a filesystem with index number: %s", self.target, fs)
```
Poeloe, line 152: This error message might be confusing for the user, because `filesystems` is a subset of the available filesystems. Therefore, the index numbering of `filesystems` may differ from the actual filesystems. Maybe a better alternative is to make `filesystems` a `dict` where the indices are the keys and filesystems objects are the values?

Poeloe (changes requested):
Hi @Zawadidone, thanks for your contribution! 😄 

I have some small change requests, please see below.

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Fixed

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Fixed

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -167,55 +174,64 @@ def noaggr(records: list[Record]) -> Iterator[Record]:
...
-                                    owner=owner,
-                                    size=size,
-                                    resident=resident,
-                                    inuse=inuse,
-                                    volume_uuid=volume_uuid,
-                                    record_formatter=record_formatter,
-                                )
+                filesystem = self.ntfs_filesystems[fs]
+            except KeyError:
+                self.target.log.error("NTFS filesystem with the index number (%s) does not exists", fs)
```
Poeloe, line 181:
```suggestion
                self.target.log.error("NTFS filesystem with index number %s does not exist", fs)
```

Poeloe (changes requested):

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Fixed

Poeloe (approved):
LGTM!

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -133,12 +137,15 @@ def check_compatible(self) -> None:
...
+    @arg("--fs", type=int, default=None, help="optional filesystem index, zero indexed")
+    @arg("--start", type=int, default=0, help="the first MFT segment number")
+    @arg("--end", type=int, default=-1, help="the last MFT segment number")
     @arg(
         "--macb",
         action="store_true",
         help="compacts the MFT entry timestamps into aggregated records with MACB bitfield",
     )
-    def mft(self, compact: bool = False, macb: bool = False):
+    def mft(self, compact: bool = False, fs: Optional[int] = None, start: int = 0, end: int = -1, macb: bool = False):
```
Schamper, line 148:
```suggestion
    def mft(self, compact: bool = False, fs: int | None = None, start: int = 0, end: int = -1, macb: bool = False):
```

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from typing import Callable, Iterator
+from typing import Callable, Iterator, Optional
```
Schamper, line 2:
```suggestion
from typing import Callable, Iterator
```

Schamper (changes requested):

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Fixed

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Fixed

dissect/target/plugins/filesystem/ntfs/mft.py
Schamper: Can you also add a `from __future__ import annotations` at the top?

Zawadidone: The CI fails due to this change https://github.com/fox-it/dissect.target/actions/runs/10266162635/job/28404734201?pr=672#step:10:41

Zawadidone: The previous change passes the tests.

Schamper: I know, if you add `from __future__ import annotations` at the top of the file it will fix it 😄

Zawadidone: Nice, I did not know that Python has this feature.

Schamper (approved):

Poeloe (dismissed):

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -167,55 +176,64 @@ def noaggr(records: list[Record]) -> Iterator[Record]:
...
+            except KeyError:
+                self.target.log.error("NTFS filesystem with index number %s does not exist", fs)
+                return
+
+            yield from self.segments(filesystem, record_formatter, aggr, start, end)
+        else:
+            for filesystem in self.ntfs_filesystems.values():
+                yield from self.segments(filesystem, record_formatter, aggr, start, end)
+
+    def segments(self, fs: NtfsFilesystem, record_formatter: Callable, aggr: Callable, start: int, end: int):
```
Schamper, line 191: Missing return type hint

Schamper (changes requested):

dissect/target/plugins/filesystem/ntfs/mft.py
Zawadidone: Should that also be added to the other functions that miss return type hints `-> Iterator[Record]:`?

* https://github.com/Zawadidone/dissect.target/blob/feature/mft_segments/dissect/target/plugins/filesystem/ntfs/mft.py#L318
* https://github.com/Zawadidone/dissect.target/blob/feature/mft_segments/dissect/target/plugins/filesystem/ntfs/mft.py#L307
* https://github.com/Zawadidone/dissect.target/blob/feature/mft_segments/dissect/target/plugins/filesystem/ntfs/mft.py#L238

Schamper: Didn't know there were so many 😄 yes please.

Schamper (approved):

Poeloe (approved):

PR #677, opened 2024-04-12
Fix for using the correct volume name to mount filesystems in Unix

Schamper:


Poeloe (changes requested):
Could you maybe add a small regression test?

Poeloe (approved):
LGTM!

PR #681, opened 2024-04-15
Rename field digest to digests

Zawadidone:
It helps when all fields that use the type `digest` are named `digests`, where possible. Depending on which search platform is used, e.g. Elasticsearch https://github.com/huntandhackett/ir-automation/blob/a005f5d2ea58a8beabe7c6ae0e25e0c0ae16ce85/logstash-dissect.conf#L54-L62, this field type must be stored as an object compared to the other field types that only contain one value (integer, string, etc.).


Zawadidone:
Yes indeed

PR #694, opened 2024-04-29
Added the 'sync' command which attempts to reload the current target

Paradoxis:
Added a `sync` command to `target-shell` which attempts to invalidate the current target by recursively looping over all `dissect.*` attributes and looks for `@lru_cache` entries to invalidate with a max-depth of 8. Should help in situations where the a live target (i.e.: a vm) has synced, but the `target-shell` hasn't, leading to an inconsistent state.

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -26,6 +26,7 @@ from flow.record import RecordOutput
 from dissect.target.exceptions import (
     FileNotFoundError,
     PluginError,
+    PluginNotFoundError,
     RegistryError,
     RegistryKeyNotFoundError,
     RegistryValueNotFoundError,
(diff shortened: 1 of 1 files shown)


Schamper:
Hi @Paradoxis,

I was so in awe by your TED talk that it took me 5 months to recover and be able to reply to your PR!

While a nice contribution, the proposed method is not something we'd like to explore further. Instead, I think it'd be better to implement a `reload` command in `target-shell`, that simply reloads the `Target` object with the same path. Note that this will only work `Target` objects that have been instantiated from a path, and not custom created ones. But we could guard against that by checking if `.path` is `None`.

Reloading the entire target object is more heavy, but it should ensure any and all caches and intermediate states are properly cleared.

PR #697, opened 2024-05-01
Make ESXi Plugin work without crypto and fix vm_inventory

Matthijsy:
Currently the ESXi module works partly without having PyCryptodome installed, however this package is not 100% needed. Furthermore the vm_inventory method did always return the vm list. This PR does 3 things:

- Check if PyCryptodome or Pystandalone is availble, if not do not use the envelope
- Don't crash if there is no ESXi config, this makes the vm_inventory method available without crypto
- Creating the local layer after the mount modules

Currently the local layer is created before the mount modules. Some of these modules (in my case vim.v00) can contain empty `vmInventory.xml` files, which results in the `vm_inventory` method returning an empty list. By creating this local layer later the version in the local layer is used.


Matthijsy:
@DissectBot agree company="Northwave"

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -37,6 +38,7 @@
     ],
 )
 
+log = logging.getLogger(__name__)
```
Schamper, line 41:
```suggestion
```

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -2,6 +2,7 @@
 
 import gzip
 import json
+import logging
```
Schamper, line 5:
```suggestion
```

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -70,7 +72,8 @@ def __init__(self, target: Target):
 
     def _cfg(self, path: str) -> Optional[str]:
         if not self._config:
-            raise ValueError("No ESXi config!")
+            log.warning("No ESXi config!")
```
Schamper, line 75:
```suggestion
            self.target.log.warning("No ESXi config!")
```
We can use the target-bound logging to get additional context on the target where the warning occurred.

Schamper (changes requested):

Schamper (approved):

Schamper (approved):

Issue #699, opened 2024-05-03
ZIP loader triggers an exception when loading Velociraptor Linux collections (Ubuntu)


Zawadidone:
Note this issue only occurs with a specific Linux collection that I made when developing https://github.com/fox-it/dissect.target/pull/698.

Executing `target-query` on a Velociraptor collection of a Linux system triggers the following exception:

```bash
target-query -t Collection-Linux.zip -f hostname
Traceback (most recent call last):
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/target.py", line 414, in _load
    ldr.map(target)
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/loaders/velociraptor.py", line 130, in map
    map_dirs(target, dirs, os_type)
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/loaders/dir.py", line 62, in map_dirs
    dfs = ZipFilesystem(path.root.fp, path.at, alt_separator=alt_separator, case_sensitive=case_sensitive)
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/filesystems/zip.py", line 60, in __init__
    self._fs.map_file_entry(rel_name, file_entry)
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/filesystem.py", line 1231, in map_file_entry
    directory = self.makedirs(sub_dirs)
  File "/dissect.target/.tox/py3/lib/python3.10/site-packages/dissect/target/filesystem.py", line 1158, in makedirs
    if part not in directory:
TypeError: argument of type 'ZipFilesystemEntry' is not iterable

The above exception was the direct cause of the following exception:
```

What stands out is that the variable `directory` in case of th exception has the type `ZipFilesystemEntry`, which is weird because in all other cases it has the type `dissect.target.filesystem.VirtualDirectory`.

Zawadidone:
Adding the following `try/except` to https://github.com/fox-it/dissect.target/blob/a6796c446c3b5f8b33ce28ea1be7814b65aecabc/dissect/target/filesystems/zip.py#L60, 'fixes' the issue.

```bash
try:
    self._fs.map_file_entry(rel_name, file_entry)
except Exception:
    continue
```

Zawadidone:
The Velociraptor collector was made with the following settings:

```bash
./velociraptor --config server.config.yaml artifacts collect Server.Utils.CreateCollector \
[...]

PR #700, opened 2024-05-05
Fix URL-encoded filesystem entries in the Velociraptor loader

### LINKED ISSUE #707: Separate mapping of filesystem entries in ZipFilesystem into another method
Currently it's done from within `__init__`, but there are some cases where you may want to subtly tweak how filesystem entries are mapped. The logic should be separated out into another method(s) so that subclasses may customize the behaviour.

Zawadidone:
Closes #707


Schamper:
Wouldn't this also be an issue on non-zip (directory) Velociraptor targets?

Zawadidone:
@Schamper yes it will, I would like to enforce the usage of a ZIP file as target, because extracting a ZIP file causes all kinds of errors. For example, the only option to extract a ZIP file in an automated manner is with `unzip -o [...]`, which enables overwriting existing files without prompting, which possible removes valuable traces.

Should this explicitly be documented in the comments of the loader or should the loader also support a non-zip directory Velociraptor target?

Schamper:
We don't use Velociraptor so I can't really speak for it, but I do think I've heard people say they sometimes work with unzipped collections. If there are no other reasons to drop support for it, I'd suggest to let it in.

I'm not a huge fan of the explicit URL decoding under a generic `decode_name` flag, but I don't immediately have a better suggestion. I was thinking maybe to pass a callback function, but I'm not sure I like that either :smile:. I'll also throw it up for discussion in the team to see if anyone has a nice idea.

Schamper:
After a short discussion we decided it may be best to tweak the `ZipFilesystem` and `DirectoryFilesystem` implementations slightly to facilitate customization by subclasses a little better. You can then subclass these filesystems within your loader and provide the customized behaviour. This subclass can live within the same file as the loader as far as I'm concerned (we also do this in other cases where we e.g. subclass the registry plugin to provide a custom implementation).

I've created #707 and #708 to implement this in the respective filesystems. Would you be up for making these changes?

Zawadidone:
Yes but I don't understand how to fix the issues. I will add comments to the issues.

OlafHaalstra:
Would be nice to keep the support for a non-zipped folder since we noticed a huge hit on performance when working with zip archives rather than unzipped directories. 

@Zawadidone wondering if you do not have the same problem with zipped vs unzipped

OlafHaalstra:
To add to the need for this.

Example, the following path is not picked up by `target-*`:

```
uploads/auto/C%3A/Users/Test/%2Essh/known_hosts 
```

This totally makes sense since the `openssh` [module](https://github.com/fox-it/dissect.target/blob/620ee4378b8bbed7c4d3676943d20d68072ed98c/dissect/target/plugins/apps/ssh/openssh.py#L52) is looking for the `.ssh` folder: 

Can we include the fix not only for the zip loader, but also for the folder loader?

OlafHaalstra:
> @Schamper yes it will, I would like to enforce the usage of a ZIP file as target, because extracting a ZIP file causes all kinds of errors. For example, the only option to extract a ZIP file in an automated manner is with `unzip -o [...]`, which enables overwriting existing files without prompting, which possible removes valuable traces.
> 
> Should this explicitly be documented in the comments of the loader or should the loader also support a non-zip directory Velociraptor target?

@Zawadidone Under which circumstances do you run into files that are overwritten?

Zawadidone:
> Would be nice to keep the support for a non-zipped folder since we noticed a huge hit on performance when working with zip archives rather than unzipped directories.
> 
> @Zawadidone wondering if you do not have the same problem with zipped vs unzipped

Yes we have the same performance hit when working with a collection stored in a ZIP file, but only when executing the MFT plugin. The other plugins are 'fast' but that's only because they return less records compared to the MFT plugin.

I haven't had the time for it, but I want to look into solving the following issues:
* Improving the speed of the MFT plugin by reading the MFT records using the segment size (https://github.com/fox-it/dissect.target/issues/665)
* Debugging performance issues with the MFT plugin when using a collection stored in a ZIP file

Zawadidone:
> To add to the need for this.
> 
> Example, the following path is not picked up by `target-*`:
> 
> ```
> uploads/auto/C%3A/Users/Test/%2Essh/known_hosts 
> ```
> 
> This totally makes sense since the `openssh` [module](https://github.com/fox-it/dissect.target/blob/620ee4378b8bbed7c4d3676943d20d68072ed98c/dissect/target/plugins/apps/ssh/openssh.py#L52) is looking for the `.ssh` folder:
> 
> Can we include the fix not only for the zip loader, but also for the folder loader?

Yes this PR will include a fix for both the ZIP as well as the DIR loader.

Zawadidone:
> > @Schamper yes it will, I would like to enforce the usage of a ZIP file as target, because extracting a ZIP file causes all kinds of errors. For example, the only option to extract a ZIP file in an automated manner is with `unzip -o [...]`, which enables overwriting existing files without prompting, which possible removes valuable traces.
> > Should this explicitly be documented in the comments of the loader or should the loader also support a non-zip directory Velociraptor target?
> 
> @Zawadidone Under which circumstances do you run into files that are overwritten?

On Linux systems that use symlinks see https://github.com/fox-it/dissect.target/issues/699, but this could occur on all kind of systems that use symlinks.

Zawadidone:
@OlafHaalstra the ZIP implementation is fixed, I am working on a fix for https://github.com/fox-it/dissect.target/issues/708.

Zawadidone:
@Schamper, I have implemented my suggestion (https://github.com/fox-it/dissect.target/issues/708) which fixes the issue, but the tests are failing because it assumes that `DirectoryFilesystem` should also work when `path` is a `str`.

dissect/target/filesystems/dir.py
```diff
@@ -1,63 +1,72 @@
...
 
 class DirectoryFilesystem(Filesystem):
     __type__ = "dir"
 
-    def __init__(self, path: Path, *args, **kwargs):
+    def __init__(self, path: Path, unquote_path: bool = False, *args, **kwargs):
         super().__init__(None, *args, **kwargs)
         self.base_path = path
 
+        self._fs = VirtualFilesystem(alt_separator=self.alt_separator, case_sensitive=self.case_sensitive)
```
Schamper, line 28: I don't think we want to do this in `DirectoryFilesystem`. It's a big resource and performance hit. The nice thing about the current implementation was that all the entries were dynamically resolved, instead of ahead of time. See also the additional comment I made in #708.

Schamper (changes requested):

dissect/target/loaders/velociraptor.py
```diff
@@ -87,11 +87,13 @@ def __init__(self, path: Path, **kwargs):
         super().__init__(path)
 
         if path.suffix == ".zip":
-            log.warning(
-                f"Velociraptor target {path!r} is compressed, which will slightly affect performance. "
-                "Consider uncompressing the archive and passing the uncompressed folder to Dissect."
-            )
             self.root = zipfile.Path(path.open("rb"))
+            compression_type = zipfile.ZipFile(str(path)).getinfo("uploads.json").compress_type
```
Schamper, line 91:
```suggestion
            compression_type = self.root.root.getinfo("uploads.json").compress_type
```
I didn't test this but I think this achieves the same without reopening the archive twice.

Schamper (changes requested):

dissect/target/loaders/velociraptor.py
Zawadidone: Fixed

Zawadidone:
The tests fail, I will look into this in a month or less.

dissect/target/filesystems/dir.py
Zawadidone: Fixed [`84a9cc0` (#700)](https://github.com/fox-it/dissect.target/pull/700/commits/84a9cc0b6c2c179174cd4aedaf063cf13a362d5e)

Schamper:
I've made some suggestions for how I envisioned the filesystem thing to work. I'll leave fixing the unit tests up to you :wink:.

tests/loaders/test_velociraptor.py
```diff
@@ -17,12 +17,14 @@ def create_root(sub_dir: str, tmp_path: Path) -> Path:
         f"uploads/{sub_dir}/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1/",
         f"uploads/{sub_dir}/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1/$Extend",
         f"uploads/{sub_dir}/%5C%5C%3F%5CGLOBALROOT%5CDevice%5CHarddiskVolumeShadowCopy1/windows/system32",
+        f"uploads/{sub_dir}/%5C%5C.%5CC%3A/%2ETEST",
```
Schamper, line 20: Are you sure this is correct? Why would they arbitrarily encode some periods as `%2E` but not others? E.g. file extensions are left alone, as is the drive path. Who thought that was a good idea :smile:.

Zawadidone:
@Schamper the implementation of the loader works, but a test keeps failing due to an unknown bug (https://github.com/fox-it/dissect.target/pull/700/commits/bc076dabf7bfd7cd3027352a749dc5f23b8ad2bd). Do you understand why this doesn't work?

Schamper:
I didn't verify (can try later if needed) but my guess is it's because of the blanket `.replace(".", "%2E")`, which will also replace periods which aren't encoded as `%2E` in the zip file (such as in file extensions).

Zawadidone:
Oh yeah indeed, I will fix this.

Zawadidone:
@Schamper the issue is fixed.

dissect/target/loaders/velociraptor.py
```diff
@@ -136,9 +136,14 @@ def map(self, target: Target) -> None:
 
 class VelociraptorDirectoryFilesystem(DirectoryFilesystem):
     def _resolve_path(self, path: str) -> Path:
-        return super()._resolve_path(quote(path, "$/").replace(".", "%2E"))
+        path = quote(path, "$/")
```
Schamper, line 139:
```suggestion
        path = quote(path, safe="$/")
```
Had to lookup what the second argument was, don't think it's commonly known.

dissect/target/loaders/velociraptor.py
```diff
@@ -136,9 +136,14 @@ def map(self, target: Target) -> None:
...
+        path = path.replace(".", "%2E", 1) if path.startswith(".") else path
```
Schamper, line 140: Does this work correctly in all cases? I'm not sure if `_resolve_path` gets called with an absolute path or relative path components. Does it still work as expected when there's a file or directory several levels deep in the filesystem tree?

dissect/target/loaders/velociraptor.py
```diff
@@ -136,9 +136,14 @@ def map(self, target: Target) -> None:
...
+        path = super()._resolve_path(path)
+        return path
```
Schamper, line 142:
```suggestion
        return super()._resolve_path(path)
```

dissect/target/loaders/velociraptor.py
```diff
@@ -136,9 +136,14 @@ def map(self, target: Target) -> None:
...
 
 
 class VelociraptorZipFilesystem(ZipFilesystem):
     def _resolve_path(self, path: str) -> str:
-        return unquote(super()._resolve_path(path)).replace("%2E", ".")
+        path = unquote(super()._resolve_path(path))
+        path = path.replace("%2E", ".", 1) if path.startswith("%2E") else path
+        return path
```
Schamper, line 149:
```suggestion
        return path.replace("%2E", ".", 1) if path.startswith("%2E") else path
```
And same question here.

Schamper (changes requested):

dissect/target/loaders/velociraptor.py
Zawadidone: Fixed

dissect/target/loaders/velociraptor.py
Zawadidone: Fixed

dissect/target/loaders/velociraptor.py
Zawadidone: Fixed

dissect/target/loaders/velociraptor.py
Zawadidone: Hmm it didn't I fixed it by using `pathlib`, it gets called with a relative path.

dissect/target/loaders/velociraptor.py
```diff
@@ -116,14 +121,26 @@ def detect(path: Path) -> bool:
...
+            os_type,
+            dirfs=VelociraptorDirectoryFilesystem,
+            zipfs=VelociraptorZipFilesystem,
+        )
+
+
+class VelociraptorDirectoryFilesystem(DirectoryFilesystem):
+    def _resolve_path(self, path: str) -> Path:
+        path = Path(quote(path, safe="$/"))
+        path = str(path.with_name(path.name.replace(".", "%2E", 1))) if path.name.startswith(".") else str(path)
```
Schamper, line 140: Using Path may not work exactly as intended on different platforms. I think in this case it's safer to use the dirname/basename/join helpers from fsutil.

Schamper (changes requested):

dissect/target/loaders/velociraptor.py
Zawadidone: Fixed.

I improved the function due to files that have encoded values by default. For example, the EVTX file `Microsoft-Windows-Windows Defender%4WHC.evtx` is stored as `Microsoft-Windows-Windows Defender%254WHC.evtx` in a Velociraptor collection, thus the values `%<space>` are not encoded.

dissect/target/loaders/velociraptor.py
```diff
@@ -116,14 +122,28 @@ def detect(path: Path) -> bool:
...
+            zipfs=VelociraptorZipFilesystem,
+        )
+
+
+class VelociraptorDirectoryFilesystem(DirectoryFilesystem):
+    def _resolve_path(self, path: str) -> Path:
+        path = quote(path, safe="$/% ")
+        path = (
+            join(dirname(path), str(basename(path)).replace(".", "%2E", 1)) if basename(path).startswith(".") else path
+        )
```
Schamper, line 143:
```suggestion
        if (fname := basename(path)).startswith("."):
            path = join(dirname(path), fname.replace(".", "%2E", 1))
```
Or something like that.

Is the single replace also correct, what if a file starts with double period?

dissect/target/filesystems/dir.py
```diff
@@ -48,6 +43,16 @@ def get(self, path: str) -> FilesystemEntry:
         else:
             entry = self.base_path.joinpath(path.strip("/"))
 
+        return entry
+
+    def get(self, path: str) -> FilesystemEntry:
+        path = path.strip("/")
+
+        if not path:
```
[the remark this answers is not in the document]
Zawadidone: Fixed

dissect/target/loaders/velociraptor.py
Zawadidone: Only the first period is encoded.

dissect/target/loaders/velociraptor.py
```diff
@@ -116,14 +122,27 @@ def detect(path: Path) -> bool:
...
+            dirfs=VelociraptorDirectoryFilesystem,
+            zipfs=VelociraptorZipFilesystem,
+        )
+
+
+class VelociraptorDirectoryFilesystem(DirectoryFilesystem):
+    def _resolve_path(self, path: str) -> Path:
+        path = quote(path, safe="$/% ")
+        if (fname := basename(path)).startswith("."):
+            path = str(join(dirname(path), fname.replace(".", "%2E", 1)))
```
Schamper, line 142:
```suggestion
            path = join(dirname(path), fname.replace(".", "%2E", 1))
```
This should already be a string I think.

Schamper (changes requested):

Schamper:
The tests are failing unfortunately.

Schamper (changes requested):
Can you check why the unit tests are failing?

Schamper:
@Zawadidone the unit tests are still failing.

Zawadidone:
I can't reproduce that, the unit tests work on my system: `platform darwin -- Python 3.10.14[pypy-7.3.17-final], pytest-8.3.3, pluggy-1.5.0`.

Schamper (approved):

PR #702, opened 2024-05-07
Return descriptive error when target tools point to a non-existing file

Poeloe:
(DIS-2565)

### DIFF
--- a/dissect/target/loaders/raw.py
+++ b/dissect/target/loaders/raw.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 
 from dissect.target import container
-from dissect.target.exceptions import TargetError
+from dissect.target.exceptions import TargetPathNotFoundError
 from dissect.target.loader import Loader
 from dissect.target.target import Target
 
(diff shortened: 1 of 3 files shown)


dissect/target/loaders/raw.py
```diff
@@ -1,13 +1,20 @@
...
 from dissect.target.loader import Loader
 from dissect.target.target import Target
 
 
 class RawLoader(Loader):
     """Load raw container files such as disk images."""
 
+    def __init__(self, path: Path, **kwargs):
+        if not path.exists():
+            raise TargetError(f"Provided path '{path}' does not exist")
```
Schamper, line 14:
```suggestion
            raise TargetError(f"Provided path does not exist: {path}")
```

dissect/target/target.py
```diff
@@ -284,7 +284,11 @@ def _find(find_path: Path, parsed_path: Optional[urllib.parse.ParseResult]):
                     try:
                         ldr = loader_cls(sub_entry, parsed_path=parsed_path)
                     except Exception as e:
-                        getlogger(sub_entry).error("Failed to initiate loader: %s", e)
+                        message = "Failed to initiate loader: %s"
+                        if isinstance(e, TargetError) and "does not exist" in str(e):
```
Schamper, line 288: Could also create a new `TargetPathNotFoundError` or something like that.

Schamper (changes requested):

Schamper (approved):

PR #704, opened 2024-05-13
Parse hostname from /proc/sys/kernel/hostname

CinisSec:
Adds the /proc/sys/kernel/hostname path to the paths mentioned in issue 606


dissect/target/plugins/os/unix/_os.py
```diff
@@ -149,7 +149,7 @@ def _parse_hostname_string(self, paths: Optional[list[str]] = None) -> Optional[
             paths (list): list of paths
         """
         redhat_legacy_path = "/etc/sysconfig/network"
-        paths = paths or ["/etc/hostname", "/etc/HOSTNAME", redhat_legacy_path]
+        paths = paths or ["/etc/hostname", "/etc/HOSTNAME", "/proc/sys/kernel/hostname", redhat_legacy_path]
```
Paradoxis, line 152: Do we always want to add the file, or only when a user doesn't specify their own paths? If we always want to include the file you're better off cloning the paths, and merging these four hostname paths together:

If the list order doesn't matter:

```python
paths = set(paths + ["/etc/hostname", "/etc/HOSTNAME", "/proc/sys/kernel/hostname", redhat_legacy_path])
````

If the list order does matter:

```python
paths = paths.copy()  # ensures that we don't modify the existing list in-place
for path in ["/etc/hostname", "/etc/HOSTNAME", "/proc/sys/kernel/hostname", redhat_legacy_path]:
    if path not in paths:
        paths.append(path)
```

Horofic (approved):

PR #716, opened 2024-05-29
Add ls parameters to target-fs

### LINKED ISSUE #626: Add ls -ah functionality to target-fs
Some functionality could probably be reused from target-shell.

DevJoost:
Add ls -ah functionality to target-fs

Mainly based it on the target-shell implementation. Also added -h (human readable) implementation that is missing in target-shell.

Resolves #626


dissect/target/tools/fs.py
```diff
@@ -23,9 +25,70 @@
 logging.raiseExceptions = False
 
 
+def human_size(bytes, units=["", "K", "M", "G", "T", "P", "E"]):
```
pyrco, line 28: Could you add type hints here also?

dissect/target/tools/fs.py
```diff
@@ -23,9 +25,70 @@
...
+    """Helper function to return the human readable string representation of bytes"""
+    return str(bytes) + units[0] if bytes < 1024 else human_size(bytes >> 10, units[1:])
+
+
 def ls(t, path, args):
```
pyrco, line 33:
```suggestion
def ls(t, path, args):
```

Since you're changing the whole function anyways, could you add type hints here also ;-)

pyrco (changes requested):
Nice addition, 2 very small things, but otherwise lgtm.

Schamper (approved):

pyrco (approved):

PR #718, opened 2024-05-30
Add support for deleted files in deleted RecycleBin directories

### LINKED ISSUE #638: Iterate through all directories inside the recycle.bin folder
Currently, the recycle.bin plugin iterates through all the folders until it finds a `$I` prefix of a directory or file. Indicating some meta data and that the deleted directory is inside the same folder with a `$R` prefix.

However, some files with `$R` prefixes that are directories with their own file inside, which are not returned by the recyclebin plugin. These files should also be returned by the recyclebin plugin.

DevJoost:
Add support for deleted files in deleted RecycleBin directories

Also made some minor changes to variable names to make it more clear what function does what. 

Resolves #638

### DIFF
--- a/dissect/target/plugins/os/windows/recyclebin.py
+++ b/dissect/target/plugins/os/windows/recyclebin.py
@@ -130,13 +130,11 @@ class RecyclebinPlugin(Plugin):
         )
 
     def read_recycle_deleted_folder(self, folder_path: TargetPath) -> RecycleBinRecord:
-        """
-        Generally speaking when deleting a file, the $R* file is the actual renamed deleted file.
-        This is however also the case for deleted folders. When a folder is deleted,
-        it is also renamed and placed here (with original recursive content).
-
-        This function will create RecycleBin records for each file in a deleted folder.
-        """
+        # Generally speaking when deleting a file, the $R* file is the actual renamed deleted file.
+        # This is however also the case for deleted folders. When a folder is deleted,
+        # it is also renamed and placed here (with original recursive content).
+        #
+        # This function will create RecycleBin records for each file in a deleted folder.
 
         meta_file = self.target.fs.path(str(folder_path.parent / folder_path.name.replace("$R", "$I")).lstrip("/"))
         if not meta_file.exists():
--- a/dissect/target/plugins/os/windows/recyclebin.py
+++ b/dissect/target/plugins/os/windows/recyclebin.py
@@ -17,7 +17,7 @@ RecycleBinRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
         ("path", "path"),
         ("filesize", "filesize"),
         ("path", "deleted_path"),
-        ("string", "source"),
+        ("path", "source"),
     ],
 )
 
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/os/windows/recyclebin.py
```diff
@@ -87,30 +87,34 @@ def recyclebin(self) -> Generator[RecycleBinRecord, None, None]:
...
 
-    def _is_recycle_file(self, path: TargetPath) -> bool:
+    def _is_recycle_meta_file(self, path: TargetPath) -> bool:
         """Check wether the path is a recycle bin metadata file"""
         return path.name and path.name.lower().startswith("$i")
 
-    def read_recycle_bin(self, bin_path: TargetPath) -> Generator[RecycleBinRecord, None, None]:
-        if self._is_recycle_file(bin_path):
-            yield self.read_bin_file(bin_path)
+    def read_recycle_file(self, path: TargetPath) -> Generator[RecycleBinRecord, None, None]:
```
Schamper, line 96:
```suggestion
    def read_recycle_file(self, path: TargetPath) -> Iterator[RecycleBinRecord]:
```

dissect/target/plugins/os/windows/recyclebin.py
```diff
@@ -125,6 +129,44 @@ def read_bin_file(self, bin_path: TargetPath) -> RecycleBinRecord:
...
         )
 
+    def read_recycle_deleted_folder(self, folder_path: TargetPath) -> RecycleBinRecord:
+        """
+        Generally speaking when deleting a file, the $R* file is the actual renamed deleted file.
+        This is however also the case for deleted folders. When a folder is deleted,
+        it is also renamed and placed here (with original recursive content).
+
+        This function will create RecycleBin records for each file in a deleted folder.
+        """
```
Schamper, line 139:
```suggestion
        # Generally speaking when deleting a file, the $R* file is the actual renamed deleted file.
        # This is however also the case for deleted folders. When a folder is deleted,
        # it is also renamed and placed here (with original recursive content).
        #
        # This function will create RecycleBin records for each file in a deleted folder.
```
This reads more like a code comment than a docstring.

dissect/target/plugins/os/windows/recyclebin.py
```diff
@@ -125,6 +129,44 @@ def read_bin_file(self, bin_path: TargetPath) -> RecycleBinRecord:
             _user=user,
         )
 
+    def read_recycle_deleted_folder(self, folder_path: TargetPath) -> RecycleBinRecord:
```
Schamper, line 132:
```suggestion
    def read_recycle_deleted_folder(self, path: TargetPath) -> RecycleBinRecord:
```
And rename accordingly.

dissect/target/plugins/os/windows/recyclebin.py
```diff
@@ -125,6 +129,44 @@ def read_bin_file(self, bin_path: TargetPath) -> RecycleBinRecord:
...
+        meta_data = meta_file.read_bytes()
+        header = self.select_header(meta_data)
+        entry = header(meta_data)
+
+        sid = self.find_sid(folder_path)
+        original_folder_path = self.target.fs.path(entry.filename.rstrip("\x00"))
+
+        user_details = self.target.user_details.find(sid=sid)
+        user = user_details.user if user_details else None
+        for parent_path, _, childs in folder_path.walk():
```
Schamper, line 154: Maybe `rglob` works better for you? Then you don't have to do the nasty path construction at line 156 and 157.

Schamper (changes requested):

skepppy:
It appears that this PR has become redundant, as this issue seems to have been fixed in #1377

Schamper:
> It appears that this PR has become redundant, as this issue seems to have been fixed in #1377

Closing.

PR #727, opened 2024-06-18
Fix edge case where unix history path is a directory

JSCU-CNI:
In the edge case where the path of a unix history file is a directory, the unix history plugin fails as it only checks the path exists and, if so, assumes it is a file. This PR fixes this problem by explicitly checking the path is a file.


cecinestpasunepipe:
Thanks, on first sight this looks good. We'll try to process this PR asap.

pyrco (approved):

PR #732, opened 2024-06-25
Add username and password options to MQTT loader

cecinestpasunepipe:
(DIS-3240)

### DIFF
--- a/dissect/target/loaders/mqtt.py
+++ b/dissect/target/loaders/mqtt.py
@@ -10,6 +10,7 @@ import time
 import urllib
 from dataclasses import dataclass
 from functools import lru_cache
+from getpass import getpass
 from pathlib import Path
 from struct import pack, unpack_from
 from threading import Thread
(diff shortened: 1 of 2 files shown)


dissect/target/loaders/mqtt.py
```diff
@@ -411,6 +418,8 @@ def connect(self) -> None:
 @arg("--mqtt-ca", dest="ca", help="certificate authority file")
 @arg("--mqtt-command", dest="command", help="direct command to client(s)")
 @arg("--mqtt-diag", action="store_true", dest="diag", help="show MQTT diagnostic information")
+@arg("--mqtt-username", dest="username", help="Username for connection")
+@arg("--mqtt-password", dest="password", help="Password for connection")
```
pyrco, line 422: Maybe an idea to prompt for the password, so it does not end up on the commandline?

pyrco (changes requested):

pyrco (approved):

PR #757, opened 2024-07-23
Fix directory argument for lnk plugin

Poeloe:


dissect/target/plugins/os/windows/lnk.py
```diff
@@ -168,7 +168,18 @@ def lnk(self, directory: Optional[str] = None) -> Iterator[LnkRecord]:
 
     def lnk_entries(self, path: Optional[str] = None) -> Iterator[TargetPath]:
```
Horofic, line 169: I believe this typehint is not correct anymore. A list of strings can be passed now? The same is true for the exported `lnk()` function.

Horofic: Could you also change the help for the `--directory` argument to reflect these changes?

dissect/target/plugins/os/windows/lnk.py
```diff
@@ -168,7 +168,18 @@ def lnk(self, directory: Optional[str] = None) -> Iterator[LnkRecord]:
...
         if path:
-            yield self.target.fs.path(path)
+            target_path = self.target.fs.path(path)
+            if not target_path.exists():
+                self.target.log.error("Provided path does not exist on target")
```
Horofic, line 173:
```suggestion
                self.target.log.error("Provided path to '.lnk' file does not exist on target: %s", target_path)
```

dissect/target/plugins/os/windows/lnk.py
```diff
@@ -168,7 +168,18 @@ def lnk(self, directory: Optional[str] = None) -> Iterator[LnkRecord]:
...
+                return
+            if target_path.is_file() and target_path.suffix.lower() != ".lnk":
+                self.target.log.error("Provided file should have suffix '.lnk'")
```
Horofic, line 176: The `Lnk` class already handles 'invalid' lnk files by raising an error.  I think its better to just try and parse the file than quitting out on a suffix check, especially when the user explicitly points to a file.

Horofic (changes requested):
Could you add a small test to test this new behavior. Also, the `dissect.shellitems` project has a tool called `parse-lnk` could you check if that tool handles directories correctly?

dissect/target/plugins/os/windows/lnk.py
Poeloe: It is still correct, because the user can still only provide a single path. However, it will do an recursive glob whenever a directory was provided by the user, which results in multiple `TargetPath` objects to be returned.

dissect/target/plugins/os/windows/lnk.py
Poeloe: Changed!

dissect/target/plugins/os/windows/lnk.py
Poeloe: All good, I removed this check

Horofic (approved):

PR #759, opened 2024-07-24
Add output_def to PluginFunction

JSCU-CNI:
This PR adds the `output_def` attribute to the `PluginFunction` dataclass and populates it during runtime in `find_plugin_functions`. The `output_def` attribute should be filled with `RecordDescriptor` instances when the `output_type` is `record`. For further assumptions see `test_find_plugin_function_record_def`.

We use this attribute to discover all possible output field types of `target-query` in our (to be published) dissect-elastic tool. This tool facilitates in generating valid Elasticsearch field mapping objects.

### DIFF
--- a/dissect/target/filesystems/config.py
+++ b/dissect/target/filesystems/config.py
@@ -3,11 +3,11 @@ from __future__ import annotations
 import io
 import textwrap
 from logging import getLogger
-from typing import Any, BinaryIO, Iterator, Optional, Union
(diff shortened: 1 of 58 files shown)


Schamper:
Thanks - this is also already covered by my own refactor of the plugin system internals, so don't be sad if this PR will be rejected 😉. I've been sick for a few days but I'm trying to get that wrapped up as quick as I can.

PR #646, opened 2024-03-19
Improve YARA plugin

JSCU-CNI:
This PR aims to improve the YARA plugin.

- `dissect.target.plugins.filesystem.yara` is now an `InternalPlugin`
- `target-yara` is now a command which calls the `filesystem.yara` plugin
- it is now possible to run compiled YARA rules with the YARA plugin
- provided rules can be compiled together which improves scan speed
- provided rules can be checked on validity before compiling them together
- arguments aim to be fully backwards compatible
- `target-query -f yara` will no longer work as it is replaced by `target-yara`
- added tests for `filesystem.yara` and `target-yara`

PR #780, opened 2024-08-01
Add option to scan children in YARA plugin

JSCU-CNI:
This is a small improvement on #646 that adds `--children` to `target-yara` / `target-query -f yara`.

### DIFF
--- a/dissect/target/plugins/filesystem/yara.py
+++ b/dissect/target/plugins/filesystem/yara.py
@@ -47,7 +47,6 @@ class YaraPlugin(Plugin):
     @arg("-p", "--path", default="/", help="path on target(s) to recursively scan")
     @arg("-m", "--max-size", default=DEFAULT_MAX_SCAN_SIZE, help="maximum file size in bytes to scan")
     @arg("-c", "--check", default=False, action="store_true", help="check if every YARA rule is valid")
-    @arg("--children", action="store_true", help="include children")
     @export(record=YaraMatchRecord)
     def yara(
         self,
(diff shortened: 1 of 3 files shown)


dissect/target/tools/yara.py
```diff
@@ -45,8 +45,8 @@ def main():
         parser.exit(1)
 
     try:
-        for target in Target.open_all(args.targets):
-            target.log.info("Scanning target")
+        for target in Target.open_all(args.targets, args.children):
+            target.log.info("Scanning target %s", target)
```
Schamper, line 49: Because you use `target.log`, it should already include the target repr in the log line, does logging it again really help?

Schamper (changes requested):

dissect/target/plugins/filesystem/yara.py
```diff
@@ -47,6 +47,7 @@ def check_compatible(self) -> None:
     @arg("-p", "--path", default="/", help="path on target(s) to recursively scan")
     @arg("-m", "--max-size", default=DEFAULT_MAX_SCAN_SIZE, help="maximum file size in bytes to scan")
     @arg("-c", "--check", default=False, action="store_true", help="check if every YARA rule is valid")
+    @arg("--children", action="store_true", help="include children")
```
Schamper, line 50: This argument is better added in `dissect/target/tools/yara.py`.

Schamper (changes requested):

dissect/target/tools/yara.py
Schamper: Now you removed the entire log line, was that intended?

Schamper (approved):

PR #781, opened 2024-08-01
Add loader option to target-info and target-yara

JSCU-CNI:
This PR adds the `--loader` / `-L` option to `target-info` and `target-yara`. Also adds a test for `target-info` since that did not exist yet.


Schamper:
Can you fix the merge conflict?

PR #782, opened 2024-08-01
Fix MdVolumeSystem

Miauwkeru:

### DIFF
--- a/dissect/target/volumes/md.py
+++ b/dissect/target/volumes/md.py
@@ -1,8 +1,7 @@
 from collections import defaultdict
 from typing import BinaryIO, Iterator, Union
 
-from dissect.volume.md import MD, MDPhysicalDisk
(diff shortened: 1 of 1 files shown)


dissect/target/volumes/md.py
```diff
@@ -1,7 +1,8 @@
 from collections import defaultdict
 from typing import BinaryIO, Iterator, Union
 
-from dissect.volume.md.md import MD, Device, find_super_block
+from dissect.volume.md import MD, MDPhysicalDisk
+from dissect.volume.md.md import find_super_block
```
Schamper, line 5:
```suggestion
from dissect.volume.md.md import MD, MDPhysicalDisk, find_super_block
```
This would also work right?

Schamper (changes requested):

pyrco (approved):

Schamper (approved):

Issue #650, opened 2024-03-21
Add configuration option on where to store the target-shell history


DissectBot:
For reasons we’d like to configure where to store the history file. By default, it should just be in the current users home directory (or the XDG data directory). But we also want the option to store it somewhere else. Current suggestion:

- In the .targetcfg.py you can have two configurable options
- First option is the base directory for a history file (e.g. /scratch/case/.history)
- Second option is the template filename for the history file (e.g. {uname}-{target_name}-history)
  
  This all should provide the ability to have per-user, per-target and per-case history files.

PR #786, opened 2024-08-05
Add target-shell history

JSCU-CNI:
This PR adds support for target-shell history files to dissect as described in #650.

Small changes were made to the config helper utility to also scan the current working directory for `.targetcfg.py` files. The storage location of the shell history by default is `$HOME/.dissect_history` and can be changed with the `HISTFILE` variable in `.targetcfg.py`.

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -136,12 +136,18 @@ class TargetCmd(cmd.Cmd):
 
     def preloop(self) -> None:
         if readline and self.histfile.exists():
-            readline.read_history_file(self.histfile)
+            try:
+                readline.read_history_file(self.histfile)
+            except Exception as e:
+                log.debug("Error reading history file: %s", e)
 
     def postloop(self) -> None:
         if readline:
             readline.set_history_length(self.histfilesize)
-            readline.write_history_file(self.histfile)
+            try:
+                readline.write_history_file(self.histfile)
+            except Exception as e:
+                log.debug("Error writing history file: %s", e)
 
     def __getattr__(self, attr: str) -> Any:
         if attr.startswith("help_"):
@@ -1271,10 +1277,6 @@ def run_cli(cli: cmd.Cmd) -> None:
         except KeyboardInterrupt:
             # Add a line when pressing ctrl+c, so the next one starts at a new line
             print()
-
-        except PermissionError as e:
-            log.error(e)
-            break
 
         except Exception as e:
             if cli.debug:
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -112,26 +112,37 @@ class TargetCmd(cmd.Cmd):
 
     CMD_PREFIX = "cmd_"
 
+    DEFAULT_HISTFILE = "~/.dissect_history"
+    DEFAULT_HISTFILESIZE = 10_000
+    DEFAULT_HISTDIR = None
+    DEFAULT_HISTDIRFMT = ".dissect_history_{uid}_{target}"
+
     def __init__(self, target: Target):
         cmd.Cmd.__init__(self)
         self.target = target
         self.debug = False
         self.identchars += "."
 
-        self.DEFAULT_HISTFILE = "~/.dissect_history"
-        self.DEFAULT_HISTFILESIZE = 10_000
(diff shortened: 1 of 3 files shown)


Schamper:
Very nice! Do you think you could implement the feature as described in #650 with regards to templating too? I like these additions but it would be awesome if you could also add that part.

For example, to keep with the naming scheme:
- `HISTDIR` - default to `None`, if set, takes precedence over `HISTFILE`
- `HISTDIRFMT` (?) - default to `{uid}-{target}-history` perhaps. Only used if `HISTDIR` is set

Or maybe a better name for `HISTDIRFMT`. It should only allow for a hardcoded set of template variables. I'd say only `uid` and `target` for now (with `{target}` being `t.name`, the slugified target name).

dissect/target/tools/shell.py
```diff
@@ -117,6 +118,21 @@ def __init__(self, target: Target):
         self.debug = False
         self.identchars += "."
 
+        self.DEFAULT_HISTFILE = "~/.dissect_history"
+        self.DEFAULT_HISTFILESIZE = 10_000
```
Schamper, line 122: These might work better as class variables.

dissect/target/tools/shell.py
```diff
@@ -117,6 +118,21 @@ def __init__(self, target: Target):
...
+
+        self.HISTFILE = pathlib.Path(getattr(target._config, "HISTFILE", self.DEFAULT_HISTFILE)).expanduser()
+        self.HISTFILESIZE = getattr(target._config, "HISTFILESIZE", self.DEFAULT_HISTFILESIZE)
```
Schamper, line 125: Probably better to keep the instance variables lowercase.

Schamper (changes requested):

Schamper (changes requested):

Schamper (approved):

Schamper:
> Is that not sufficient? 

It was:
image

Schamper:
The PyPy tests on Linux are struggling real hard, not sure if it's a CI problem or related to this PR.

Schamper:
Should be `tests/tools/test_shell.py::test_target_cli_unicode_argparse`.

Schamper:
> Running `tox -e pypy39` however seems to hang when using `pypy3.9`. Removing `os.getcwd()` from `config.load` does not fix it. Downgrading `tox` or `virtualenv` also does not seem to help.

Unfortunately I can't reproduce this on Apple Silicon.

dissect/target/tools/shell.py
```diff
@@ -111,12 +112,37 @@ class TargetCmd(cmd.Cmd):
...
+            self.histfile = pathlib.Path(getattr(target._config, "HISTFILE", self.DEFAULT_HISTFILE)).expanduser()
+
+    def preloop(self) -> None:
+        if readline and self.histfile.exists():
+            readline.read_history_file(self.histfile)
+
+    def postloop(self) -> None:
+        if readline:
+            readline.set_history_length(self.histfilesize)
+            readline.write_history_file(self.histfile)
```
Schamper, line 144:
```suggestion
            try:
                readline.write_history_file(self.histfile)
            except Exception as e:
                log.debug("Error writing history file: %s", e)
```

dissect/target/tools/shell.py
```diff
@@ -1241,18 +1267,22 @@ def run_cli(cli: cmd.Cmd) -> None:
...
+
         except KeyboardInterrupt:
             # Add a line when pressing ctrl+c, so the next one starts at a new line
             print()
-            pass
+
+        except PermissionError as e:
+            log.error(e)
+            break
+
```
Schamper, line 1278:
```suggestion
```

Schamper (changes requested):
I somehow managed to get a reproducable case on my own system. I think the test got stuck in an infinite loop because of some sort of unhandled error (OSError on my system) when trying to read the history file, which would cause `run_cli` to loop infinitely. The proposed changes moves the error handling directly to the reading and writing of the history files, which should fix the issue.

Schamper:
> Nice find! I wonder why those stacktraces were not showing up during testing. Perhaps because of the new `self.debug` behaviour of `TargetCmd`?

`test_target_cli_unicode_argparse` uses `capsys` so they were captured by pytest 😅.

Schamper (approved):

Issue #411, opened 2023-10-09
make dissect.target handle registry hive files directly


Politie-SOC:
Dissect currently does not know how to parse registry hive files directly. They have to be inside a valid "target" before they can be processed. It would be nice to be able to use target-query and target-reg directly against a registry hive file.

Schamper:
This could perhaps fit nicely with the idea proposed here: https://github.com/fox-it/dissect.target/pull/399#issuecomment-1736346892

Specifically for registry hives we would need to add a little bit of heuristic checks to figure out which hive we're dealing with (user hive, SYSTEM, SOFTWARE, etc).

PR #787, opened 2024-08-05
Add a path remapping option to LogLoader

MaxGroot:
This PR is a band-aid to address the issues discussed in https://github.com/fox-it/dissect.target/pull/399. When I have forensic artifacts without a target (e.g, when doing a ctf, or processing outputs from other tooling) I'd like to be able to invoke a plugin to parse it. While I agree with the discussion in #399 that it would be better to have a uniform API where every plugin specifies what paths they need to work (and then having an option to provide / extend these paths on the fly), it might take a while to implement that.

The change in this PR allows the user to 'remap' a given file / directory to a specific location in the virtual filesystem, so that plugins that do little more than parsing a (set of) file(s) will work. For example, a wtmp file could be parsed like this:

`target-query "log://wtmp?path=/var/log/wtmp" -f wtmp`

Or apache access logs like this:
`target-query "log://apache_dir?path=/var/log/apache" -f apache.access`

Of course, for a lot of plugins, this still won't be enough to make it work. For example, the commandhistory plugin requires the target to have users. Windows plugins doing anything with the registry are also not addressed by this (as also mentioned in https://github.com/fox-it/dissect.target/issues/411). Lastly, it still requires the user to know where a given forensic artifact is expected to be located to map it accordingly.

If you have something better in mind, feel free to just close this PR and implement that instead. Perhaps it's better to just extend the `LOGS_DIRS` dict, but that seemed to me like further hardcoding away the underlying problem.

### DIFF
--- a/dissect/target/loaders/log.py
+++ b/dissect/target/loaders/log.py
@@ -11,13 +11,20 @@ from dissect.target.plugin import arg
 
 
 @arg("--log-hint", dest="hint", help="hint for file type")
+@arg("--path", dest="path", help="Map log file(s) to a specific path in the target filesystem")
 class LogLoader(Loader):
-    """Load separate log files without a target.
+    """Load separate log files without a target. By default attempts to map discovered log files based on their file
+    extension. The loader can also map log files to a specific path in the target filesystem using the ``--path``
+    option.
 
     Usage:
 
     ``target-query /evtx/* -L log -f evtx``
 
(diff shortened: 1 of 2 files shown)


Schamper:
Since the log loader will be deprecated soon and the long term plan is to separate the parser from the plugin (GitHub issue number still pending it seems), I'm closing this. If you need this functionality in the meantime, feel free to make a new PR based around the `DirectLoader` instead!

PR #809, opened 2024-08-07
Improve Linux OS detection

### LINKED ISSUE #807: Windows targets get incorrectly classified as Linux targets
When a `/proc` or `/sys` folder exists in the root of a Windows target the `LinuxPlugin` incorrectly classifies it as a Linux target, even though the `WindowsPlugin` also classifies it as a Windows target. Then due to how OS Plugin precedence is determined the `LinuxPlugin` is chosen over the `WindowsPlugin`

Horofic:
Fixes #807


dissect/target/plugins/os/unix/linux/_os.py
```diff
@@ -26,11 +26,10 @@ def detect(cls, target: Target) -> Optional[Filesystem]:
                 fs.exists("/var")
                 and fs.exists("/etc")
                 and fs.exists("/opt")
-                or (fs.exists("/sys") or fs.exists("/proc"))
-                and not fs.exists("/Library")
+                or (fs.exists("/proc") or fs.exists("/sys"))
```
Schamper, line 29: Why is this necessary in the first place? Are there scenarios where you _only_ have a `/proc` or `/sys`? Why not make this stricter.

Horofic: Acquire allows for them to be separately collected, so technically a acquire collection of just `/proc` or `/sys` could happen.

dissect/target/plugins/os/unix/linux/_os.py
```diff
@@ -26,11 +26,13 @@ def detect(cls, target: Target) -> Optional[Filesystem]:
...
             ):
+                # If the target has registry function, it is most likely not a Linux system
+                if target.has_function("registry") or fs.exists("/Library"):
```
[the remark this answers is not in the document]
Horofic: Currently typing `registry` in `target-shell` does not register the `registry` plugin to the target if the os is `linux`. Was this behaviour changed in #812? If not, this should not be a problem.

Schamper:
What about this? This also _finally_ fixes the problem that if you run the local loader with the incorrect permissions, it will wrongly select FreeBSD as the OS with `/proc` as the root filesystem because `/proc/net` exists.

Schamper (approved):

PR #810, opened 2024-08-07
Create base class for the network plugin

### LINKED ISSUE #774: Create base class for default OS network interface plugin
Task 1/4 for network interface plugin

<u>Motivation</u>:

Currently there's only `-f ips`, but it's very limiting. E.g. if the machine has DHCP, `-f ips` returns nothing. Ideally there'd be something akin to `-f interfaces`, which would list all network interfaces (probably as a record), including the IP configuration (DHCP, netmask, search domain). Specific information could also be filtered/queried using namespace exports such as `-f interfaces.ips`, `-f interfaces.gateways`.

 <u>Technical information</u>:

- Search domain in **unix variants** is mostly not by interface (at least not in a resolv.conf or one of its newer incarnations, systemd may differ).
- Both in **windows** and \***nix** there may be a global configurable list (not interface specific).

<u>**Important notes for implementation**</u>**:** 

- **First** (under this issue) set down a "framework"/**base classes** for the network interface plugin that allows OS-specific implementations and customization. 
- The framework/base class **design** should consider the **complexity of multiple network managers**
- Create OS-specific classes for (#773) Windows, mac-OS (#775) and Linux (#776). Start with Windows OS-specific class since it’s a little easier and less chaotic (only one network manager!)
- Below is the design scheme with notes for the base class:

```
windows/
    _os.py
    network.py

_os.py:
    Still has the ips export for backwards compatibility
    
network.py
    Can take inspiration from datetime namespace plugin.
    
    Define a NetworkPlugin based on a base class with some defined exports.
    It should be defined in plugins/general/network.py, so that the Default OS plugin can also make use of it. 
    The exported functions can just return empty iterators.
    
    Should also have some @internal functions for interface lookup (similar to user_details)
    
    - interfaces
      - InterfaceRecord
        - Base record for common stuff, OS specialized fields where applicable
          - Can take inspiration from UserRecord/Browser*Record
        - Fields (in no particular order):
          - Type (TUN/TAP/Physical)
          - Name
          - Enabled/disabled
          - MAC (optional)
          - IP / subnet
          - VLAN? (maybe macOS specific?/Windows)
          - Gateway
          - DNS server / Search domain (Windows/macOS specific?)
          - Proxy (at least macOS specific)
          - Metric (Windows specific)
          - Interface service order (macOS)
          - Last connected (maybe only Windows?)
          - Config source (similar to how we use the source field in other records)
    - ips -> shorthand for getting all ips
    - gateways/mac/dns/other common fields -> idem as ^
```

Miauwkeru:
closes #774

### DIFF
--- a/dissect/target/loaders/libvirt.py
+++ b/dissect/target/loaders/libvirt.py
@@ -1,0 +1,40 @@
+from pathlib import Path
+
+from defusedxml import ElementTree
+
+from dissect.target import container
+from dissect.target.helpers import fsutil
+from dissect.target.loader import Loader
+from dissect.target.target import Target
+
+
+class LibvirtLoader(Loader):
+    """Load libvirt xml configuration files."""
+
+    def __init__(self, path: Path, **kwargs):
+        path = path.resolve()
+        self.base_dir = path.parent
+        super().__init__(path)
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        if path.suffix.lower() != ".xml":
+            return False
+
+        with path.open("rb") as fh:
+            lines = fh.read(512).split(b"\n")
+            # From what I've seen, these are are always at the start of the file
+            # If its generated using virt-install
+            needles = [b"<domain", b"<name>", b"<uuid>"]
+            return all(any(needle in line for line in lines) for needle in needles)
+
+    def map(self, target: Target) -> None:
+        xml_data = ElementTree.fromstring(self.path.read_text())
+        for disk in xml_data.findall("devices/disk/source"):
+            if not (file := disk.get("file")):
+                continue
+
+            for part in [fsutil.basename(file), file]:
+                if (path := self.base_dir.joinpath(part)).exists():
+                    target.disks.add(container.open(path))
+                    break
--- a/dissect/target/plugins/child/qemu.py
+++ b/dissect/target/plugins/child/qemu.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from typing import Iterator
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import ChildTargetRecord
+from dissect.target.plugin import ChildTargetPlugin
+
+
+class QemuChildTargetPlugin(ChildTargetPlugin):
+    """Child target plugin that yields all QEMU domains from a KVM libvirt deamon."""
+
+    __type__ = "qemu"
+
+    def check_compatible(self) -> None:
+        if not self.target.fs.path("/etc/libvirt/qemu").exists():
+            raise UnsupportedPluginError("No libvirt QEMU installation found")
+
+    def list_children(self) -> Iterator[ChildTargetRecord]:
+        for domain in self.target.fs.path("/etc/libvirt/qemu").glob("*.xml"):
+            yield ChildTargetRecord(type=self.__type__, path=domain)
--- a/tests/_data/loaders/libvirt/qemu.xml
+++ b/tests/_data/loaders/libvirt/qemu.xml
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:b2b0baced7cade7aee07d6fe4c1e504e591e3b3bef70522992ec9d4413f2a528
+size 7690
--- a/tests/loaders/test_libvirt.py
+++ b/tests/loaders/test_libvirt.py
@@ -1,0 +1,40 @@
+from pathlib import Path
+from unittest.mock import Mock, patch
+
+from dissect.target import Target
+from dissect.target.filesystem import VirtualFilesystem
+from dissect.target.loaders.libvirt import LibvirtLoader
+from tests.conftest import absolute_path
+
+
+def test_libvirt_detection(target_bare: Target, tmp_path: Path) -> None:
+    xml_file = tmp_path.joinpath("base")
+    xml_file.write_text("not a libvirt file")
+
+    assert LibvirtLoader.detect(xml_file) is False
+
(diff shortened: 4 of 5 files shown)


dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,77 @@
...
+    def gateways(self) -> Iterator[Record]:
+        yield from self._get_record_type("gateway")
+
+    @export(output="record", export=Record)
+    def macs(self) -> Iterator[Record]:
+        yield from self._get_record_type("mac")
+
+    @export(output="record", export=Record)
+    def dns(self) -> Iterator[Record]:
+        yield from self._get_record_type("dns")
```
Schamper, line 77: What's the value in making these yield records? Why not a list? You already have them in the record from `interfaces`.

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,77 @@
...
+            type_record = create_record(record, field_name)
+            yield type_record(
+                source=record.source,
+                **{
+                    field_name: getattr(record, field_name, None),
+                },
+                _target=self.target,
+            )
+
+    @export(output="record", export=InterfaceRecord)
```
Schamper, line 59:
```suggestion
    @export(record=InterfaceRecord)
```

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return TargetRecordDescriptor(record_descriptor, record_fields)
+
+
+class NetworkPlugin(Plugin):
+    __namespace__ = "network"
+
+    def __init__(self, target: Target):
+        super().__init__(target)
+
```
Schamper, line 40:
```suggestion
```

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,77 @@
...
+from dissect.target.helpers.record import (
+    MacInterfaceRecord,
+    TargetRecordDescriptor,
+    UnixInterfaceRecord,
+    WindowsInterfaceRecord,
+)
+from dissect.target.plugin import Plugin, Record, export, internal
+from dissect.target.target import Target
+
+# Using this notation as it didn't like the pipe symbol
```
Schamper, line 13:
```suggestion
```
Don't think a comment in the code is necessary for this. It's because we're being naughty boys and using variables as type hints.

dissect/target/helpers/record.py
```diff
@@ -142,3 +142,41 @@ def DynamicDescriptor(types):  # noqa
...
+    ("boolean", "enabled"),
+    ("string", "mac"),
+    ("net.ipaddress", "ip"),
+    ("net.ipaddress", "gateway"),
+    ("string", "source"),
+]
+
+
+UnixInterfaceRecord = TargetRecordDescriptor(
+    "unix/interface",
```
Schamper, line 158:
```suggestion
    "unix/network/interface",
```
Maybe? And for the others.

dissect/target/helpers/record.py
```diff
@@ -142,3 +142,41 @@ def DynamicDescriptor(types):  # noqa
...
+    "unix/network/interface",
+    COMMON_INTERFACE_ELEMENTS,
+)
+
+WindowsInterfaceRecord = TargetRecordDescriptor(
+    "windows/network/interface",
+    [
+        *COMMON_INTERFACE_ELEMENTS,
+        ("string", "vlan"),
+        ("string", "dns_server"),
```
cecinestpasunepipe, line 167: should this not be just "dns"?

cecinestpasunepipe (changes requested):

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,54 @@
...
+
+class NetworkPlugin(Plugin):
+    __namespace__ = "network"
+
+    def check_compatible(self) -> None:
+        pass
+
+    @internal
+    def find_interfaces(self) -> list[InterfaceRecord]:
+        yield from ()
```
Schamper, line 23: This shouldn't be internal? And probably named something else, maybe just `_interfaces`, so that subclasses can just implement that and not have to re-export `interfaces`.

I think you got confused with a part in the ticket that references internal methods for looking up/finding interfaces by IP/MAC/CIDR, like we have methods to look up users by [SU]ID.

Schamper (changes requested):

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def _get_record_type(self, field_name: str) -> list[Any]:
+        for record in self.find_interfaces():
+            output = getattr(record, field_name, None)
+            if output is None:
+                continue
```
Schamper, line 29:
```suggestion
            if (output := getattr(record, field_name, None)) is None:
                continue
```
Not sure if this works but I think it does.

Schamper (changes requested):

dissect/target/plugins/general/network.py
```diff
@@ -0,0 +1,76 @@
...
+    WindowsInterfaceRecord,
+)
+from dissect.target.plugin import Plugin, export, internal
+
+InterfaceRecord = Union[UnixInterfaceRecord, WindowsInterfaceRecord, MacInterfaceRecord]
+
+
+class NetworkPlugin(Plugin):
+    __namespace__ = "network"
+    _interface_list: list = None
```
Schamper, line 17:
```suggestion

    def __init__(self, target: Target):
        self._interface_list: list[InterfaceRecord] | None = None
```

Schamper (changes requested):

Schamper (approved):

cecinestpasunepipe (approved):

PR #823, opened 2024-08-15
Added support for MSSQL error log files

fox-evv:
This can show e.g. that xp_cmdshell was enabled


Horofic (changes requested):
Made some suggestions and added some tests with a commit. Please test in your own environment to see if the suggestions hold up.

Horofic (approved):

Horofic (approved):

PR #824, opened 2024-08-15
Add computer SID for Windows systems

fox-evv:
This addition to the generic Windows plugin exports the machine and domain SIDs
of the target. If the system is not joined to a domain, it uses only the/
SAM hive, but if the system is domain joined, it will all so use
the Security Policy.


Horofic:
@fox-evv I've suggested some changes and also added some test with this commit. One thing I could not verify myself is whether there can be multiple Machine / Domain SIDs. The initial implementation kind of seemed to suggest this. Do you encounter this scenario yourself? I left it out of this implementation because it seemed illogical to me.

dissect/target/plugins/os/windows/generic.py
```diff
@@ -571,3 +585,36 @@ def codepage(self) -> Optional[str]:
             return self.target.registry.key(key).value("ACP").value
         except RegistryError:
             pass
+
+    @export(record=ComputerSidRecord)
+    def sid(self) -> Iterator[ComputerSidRecord]:
+        """Return the machine- and optional domain SID of the system."""
+
+        try:
+            key = self.target.registry.key("HKLM\\SAM\\SAM\\Domains\\Account")
```
Horofic, line 594: Can there be multiple Machine- and Domain SIDS associated with one system?

Horofic (changes requested):

Horofic (approved):

tests/plugins/os/windows/test_generic.py
```diff
@@ -2,13 +2,45 @@
...
     currentversion_key_name = "Software\\Microsoft\\Windows NT\\CurrentVersion"
     currentversion_key = VirtualKey(hive_hklm, currentversion_key_name)
     currentversion_key.add_value("InstallDate", VirtualValue(hive_hklm, "InstallDate", 0))
     hive_hklm.map_key(currentversion_key_name, currentversion_key)
     target_win_users.add_plugin(GenericPlugin)
 
     assert target_win_users.install_date == from_unix(0)
+
+
+def test_windows_generic_sid(target_win: Target, hive_hklm: VirtualHive):
```
Schamper, line 19: Missing return type hint

tests/plugins/os/windows/test_generic.py
```diff
@@ -2,13 +2,45 @@
 
 from dissect.target.helpers.regutil import VirtualKey, VirtualValue
 from dissect.target.plugins.os.windows.generic import GenericPlugin
+from dissect.target.plugins.os.windows.registry import VirtualHive
+from dissect.target.target import Target
 
 
-def test_windows_generic_install_date(target_win_users, fs_win, hive_hklm):
+def test_windows_generic_install_date(target_win_users: Target, hive_hklm: VirtualHive):
```
Schamper, line 9: Missing return type hint

dissect/target/plugins/os/windows/generic.py
```diff
@@ -113,6 +118,15 @@
     ],
 )
 
+ComputerSidRecord = TargetRecordDescriptor(
+    "windows/sid/computer",
+    [
+        ("datetime", "ts"),
+        ("string", "sidtype"),
```
Schamper, line 125: Should be `type` and not `sidtype`.

Schamper (commented):
@Horofic can you fix these notes or create a ticket for them?

PR #829, opened 2024-08-20
Add notice for PyPy users 

cecinestpasunepipe:
(DIS-3295)


cecinestpasunepipe:
> We wait until we they do a release? [pypy/pypy#5006](https://github.com/pypy/pypy/issues/5006)

Yes, let's give them some time to respond before we implement the notice.
If the manage to fix the issue we can probably do a version check as well, and urge our users to download the latest pypy.

Issue #773, opened 2024-07-31
Create Windows class for network interface plugin 


DissectBot:
Task 2/4 for network interface plugin, for background look into issue #774

- Create **Windows OS-specific class** since it’s a little easier and less chaotic (only one network manager!)
- Below is the design scheme with notes for windows:

```
windows/
    _os.py
    network.py

_os.py:
    Still has the ips export for backwards compatibility
    
network.py
    Can take inspiration from datetime namespace plugin.
    
    Define a NetworkPlugin based on a base class with some defined exports.
    It should be defined in plugins/general/network.py (plugins/os/default/network.py after https://github.com/fox-it/dissect.target/pull/763), so that the Default OS plugin can also make use of it. 
    The exported functions can just return empty iterators.
    
    Should also have some @internal functions for interface lookup (similar to user_details)
    
    - interfaces
      - InterfaceRecord
        - Base record for common stuff, OS specialized fields where applicable
          - Can take inspiration from UserRecord/Browser*Record
        - Fields (in no particular order):
          - Type (TUN/TAP/Physical)
          - Name
          - Enabled/disabled
          - MAC (optional)
          - IP / subnet
          - VLAN? (maybe macOS specific?)
          - Gateway
          - DNS server / Search domain (Windows/macOS specific?)
          - Proxy (at least macOS specific)
          - Metric (Windows specific)
          - Interface service order (macOS)
          - Last connected (maybe only Windows?)
          - Config source (similar to how we use the source field in other records)
    - ips -> shorthand for getting all ips
    - gateways/mac/dns/other common fields -> idem as ^
    
Windows notes:
  - Most of this information is stored in the registry
  - Maybe not initially, but maybe nice to research if there's dynamic information available in evtx/etl
  
```

Issue #778, opened 2024-07-31
Create pathlib-like layer for registry abstraction layer


Schamper:
Currently the registry abstraction layer only works on "existing" nodes (like our filesystem implementation). A pathlib-like layer on top could work on "hypothetical" paths and could allow for things like `exists()` on keys or values.

Schamper:
I'm playing a little bit with this with https://github.com/fox-it/dissect.target/pull/1390 as basis. Having a new, stable "PurePath" variant to build a `PureRegistryPath` off is quite nice. I think the `RegistryPath` will not subclass `Path` since there's little point to it, but will instead have its own set of similar methods.

If you think about it, the registry is basically a filesystem anyway. So I would propose `RegistryPath` to treat registry paths like such as well. This means that the final component of a path can also be a value. E.g. `HKLM\\path\\to\\value`, `value` would be a value in the `to` subkey.

(I thought about making a target filesystem implementation for the registry, but besides the novelty of being able to `target-shell` into a bare registry file (which I believe would be the wrong way to "solve" accessing bare registry files), I don't think it would solve any of the issues we actually have surrounding making registry access easier. The `TargetPath` on that filesystem would be awkward to interact with as they would have filesystem naming conventions).

Some just from the top of my head:
`PureRegistryPath`:
- `hive -> str` - property, to retrieve the "hive" portion of the registry key (e.g. first part)

`RegistryPath`:
- `exists() -> bool`
- `is_subkey() -> bool`
- `is_value() -> bool`
- `iter() -> Iterator[Self]` - combined iterator of subkeys and values
- `itervalues() -> Iterator[Self]` - iterator of just values
- `iterkeys() -> Iterator[Self]` - iterator of just subkeys
- `glob() -> Iterator[Self]`
- `rglob() -> Iterator[Self]`
- `walk() -> Iterator[Self]`
- `user()` ?
- Something to nicely read values
- Something to nicely get underlying attributes (value type, key class name, key timestamp, etc)

Open to suggestions. I will park this for now and just slowly continue working on the required parts to make this work in the first place.

PR #830, opened 2024-08-21
Add Windows network interface plugin

narimantos:
Closes: #773


dissect/target/helpers/record.py
```diff
@@ -164,8 +164,11 @@ def DynamicDescriptor(types):  # noqa
     "windows/network/interface",
     [
         *COMMON_INTERFACE_ELEMENTS,
-        ("varint", "vlan"),
+        ("string", "vlan"),
+        ("net.IPAddress[]", "subnetmask"),
```
Schamper, line 168: Might be nice to use `net.ipnetwork` here instead, and call it `network`, and move it to the common interface elements.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+            "*PriorityVlanTag",
+            "*IfType",
+        ]
+
+        for r in self.target.registry.keys(key):
+            for s in r.subkeys():
+                device_info = {}
+                config = None
+
+                for field in fields:
```
Schamper, line 55: If you're going to access every field anyway and check for it with an `if`, I'd not use a loop and just hardcode every field access.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+    RegistryKeyNotFoundError,
+    RegistryValueNotFoundError,
+)
+from dissect.target.helpers.record import WindowsInterfaceRecord
+from dissect.target.plugin import internal
+from dissect.target.plugins.general.network import NetworkPlugin
+
+WINDOWS_EPOCH = datetime.datetime(1601, 1, 1)
+UNIX_EPOCH = datetime.datetime(1970, 1, 1)
+EPOCH_DIFF = (UNIX_EPOCH - WINDOWS_EPOCH).total_seconds()  # Difference in seconds
```
Schamper, line 13: Use `dissect.util.ts.wintimestamp` instead.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+class WindowsNetworkPlugin(NetworkPlugin):
+    @internal
+    def _interfaces(self) -> list[WindowsInterfaceRecord]:
+        for device_info in self._extract_network_devices():
```
Schamper, line 19: I'd inline the `self._extract_network_devices` method inside this one. Don't think there's a reason to have it separate.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+            "NameServer",
+            "DhcpSubnetMask",
+            "SubnetMask",
+            "DhcpDomain",
+            "Domain",
+            "InterfaceMetric",
+        ]
+        config = {"domain": [], "gateway": [], "ip": [], "dns": [], "subnetmask": [], "metric": []}
+
+        for field in fields:
```
Schamper, line 117: Same comment here about looping through the fields. I'd hardcode it. If you're doing this for re-use of the `try/except` construction for the registry stuff, throw that in a function. #778 would improve this later down the road.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+            + "\\Connection"
+        )
+        try:
+            name = self.target.registry.key(key).value("Name").value
+            return name
+        except (RegistryValueNotFoundError, RegistryKeyNotFoundError):
+            return "Unknown"
+
+
+IfTypes = {
```
Schamper, line 177: Can you move this to the top of the file? And make it a proper `enum.Enum` maybe.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+    192: "REACH_DSL",
+    193: "FR_DLCI_ENDPT",
+    194: "ATM_VCI_ENDPT",
+    195: "OPTICAL_CHANNEL",
+    196: "OPTICAL_TRANSPORT",
+    243: "WWANPP",
+    244: "WWANPP2",
+}
+
+PriorityVlanTagEnum = {
```
Schamper, line 378: Also move this to the top but probably fine keeping it like a dict because of the string.

dissect/target/helpers/record.py
```diff
@@ -164,8 +164,11 @@ def DynamicDescriptor(types):  # noqa
     "windows/network/interface",
     [
         *COMMON_INTERFACE_ELEMENTS,
-        ("varint", "vlan"),
+        ("string", "vlan"),
```
Schamper, line 167: Is there no VLAN tag number on Windows?

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+
+        if (
+            len(config["gateway"])
+            + len(config["ip"])
+            + len(config["dns"])
+            + len(config["subnetmask"])
+            + len(config["domain"])
+            + len(config["metric"])
+            > 0
+        ):
```
cecinestpasunepipe, line 161:
```suggestion
if (sum(map(len, config.values()))):
```

maybe more elegant?

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+            try:
+                value = self.target.registry.key(key).value(field).value
+            except RegistryValueNotFoundError:
+                continue
+            except RegistryKeyNotFoundError:
+                break
+
+            if isinstance(value, str):
+                value = [value]
+            if isinstance(value, int):
```
cecinestpasunepipe, line 127:
```suggestion
            elif isinstance(value, int):
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,383 @@
...
+                value = [str(value)]
+
+            for i in value:
```
cecinestpasunepipe, line 130: maybe rename i to value and value to values?

cecinestpasunepipe (changes requested):

tests/plugins/os/windows/test_network.py
```diff
@@ -0,0 +1,196 @@
...
+            },
+        ),
+    ],
+)
+def test_windows_network(
+    mock_values, expected_ips, expected_dns, expected_gateways, expected_macs, expected_interface, target_win: Target
+):
+    with patch("dissect.target.plugins.os.windows.generic.GenericPlugin", return_value=""):
+        with patch("dissect.target.plugins.os.windows._os.WindowsPlugin.hostname", return_value="hostname"):
+            with patch.object(target_win, "registry", MockRegistry(mock_values)):
```
cecinestpasunepipe, line 183:
```suggestion
           mock_value_dict = dict(map(lambda el: (el[0], MockRegVal(name=el[0], value=el[1])), mock_values.items()))
           with patch.object(target_win, "registry", MockRegistry(mock_value_dict)):
```

I think you can shorten the mock_values by using a map...

tests/plugins/os/windows/test_network.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+@dataclass
+class MockRegVal:
+    name: str
+    value: str
+
+
+class MockRegistry:
+    def __init__(self, values):
```
cecinestpasunepipe, line 22: I think you forgot some typehints...

cecinestpasunepipe (changes requested):

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+    RegistryKeyNotFoundError,
+    RegistryValueNotFoundError,
+)
+from dissect.target.helpers.record import WindowsInterfaceRecord
+from dissect.target.plugin import internal
+from dissect.target.plugins.general.network import NetworkPlugin
+
+REGISTRY_KEY_INTERFACE = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces\\"
+REGISTRY_KEY_CONTROLSET = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}"
+REGISTRY_KEY_CONNECTION = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Network\\{4D36E972-E325-11CE-BFC1-08002BE10318}\\"
```
Schamper, line 15: I'd just put these in the code.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+    0: "Priority & VLAN Disabled",
+    1: "Priority Enabled",
+    2: "VLAN Enabled",
+    3: "Priority & VLAN Enabled",
+}
+
+
+class WindowsNetworkPlugin(NetworkPlugin):
+    @internal
+    def _interfaces(self) -> list[WindowsInterfaceRecord]:
```
Schamper, line 243: Wrong return type, you yield instead of returning a list.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+        # Get all the network interfaces
+        for r in self.target.registry.keys(REGISTRY_KEY_CONTROLSET):
+            for subkey in r.subkeys():
+                device_info = {}
+                config = None
+                if net_cfg_instance_id := self._try_value(subkey, "NetCfgInstanceId"):
```
Schamper, line 249: Make `_try_value` a function instead of a method on this class.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+                    config = self._extract_network_device_config(net_cfg_instance_id)
+                    if config:
+                        device_info.update(config)
+
+                        # Extract the network device name for given interface id
+                        device_info["Name"] = self._extract_network_device_name(net_cfg_instance_id)
+
+                        device_info["NetworkAddress"] = self._try_value(subkey, "NetworkAddress")
+                        device_info["NetworkInterfaceInstallTimestamp"] = wintimestamp(
+                            self._try_value(subkey, "NetworkInterfaceInstallTimestamp")
```
Schamper, line 260: Won't this error when `_try_value` returns `None`?

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+                        )
+                        device_info["DriverDesc"] = self._try_value(subkey, "DriverDesc")
+                        device_info["Type"] = IfTypes(int(self._try_value(subkey, "*IfType"))).name
+                        device_info["Vlan"] = PriorityVlanTagEnum.get(
+                            int(self._try_value(subkey, "*PriorityVlanTag")), "OTHER"
+                        )
+
+                        if "Type" in device_info:
+                            yield WindowsInterfaceRecord(
+                                _target=self.target,
```
Schamper, line 270: Can you put `_target` at the bottom?

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,327 @@
...
+                        config[MAPDEF[name]] = str(value)
+                    else:
+                        config[MAPDEF[name]].extend([value])
+                elif isinstance(value, list):
+                    config[MAPDEF[name]].extend(value)
+                elif isinstance(value, int):
+                    config[MAPDEF[name]] = value
+
+        if any(config.values()):
+            return config
```
Schamper, line 313: This function is very hard to comprehend and understand what it actually does, while what it's supposed to do is actually very simple. I'd lose the dictionary stuff and looping stuff and just make it a static series of lookups, value extractions and assignments over the individual values.

Schamper (changes requested):

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+    OPTICAL_CHANNEL = 195
+    OPTICAL_TRANSPORT = 196
+    WWANPP = 243
+    WWANPP2 = 244
+
+
+def _try_value(subkey: KeyCollection, value: str) -> str | list | None:
+    try:
+        return subkey.value(value).value
+    except (RegistryValueNotFoundError, RegistryKeyNotFoundError):
```
Schamper, line 222:
```suggestion
    except RegistryValueNotFoundError:
```
I think, because you already pass in a subkey.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+    REACH_DSL = 192
+    FR_DLCI_ENDPT = 193
+    ATM_VCI_ENDPT = 194
+    OPTICAL_CHANNEL = 195
+    OPTICAL_TRANSPORT = 196
+    WWANPP = 243
+    WWANPP2 = 244
+
+
+def _try_value(subkey: KeyCollection, value: str) -> str | list | None:
```
Schamper, line 219:
```suggestion
def _try_value(subkey: RegistryKey, value: str) -> str | list | None:
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+    except (RegistryValueNotFoundError, RegistryKeyNotFoundError):
+        return None
+
+
+class WindowsNetworkPlugin(NetworkPlugin):
+    REGISTRY_KEY_INTERFACE = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces\\"
+    REGISTRY_KEY_CONTROLSET = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}"
+    REGISTRY_KEY_CONNECTION = (
+        "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Network\\{4D36E972-E325-11CE-BFC1-08002BE10318}\\"
+    )
```
Schamper, line 231: I meant to use them directly in the code, no variables at all :smile: You only ever use them once.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+                yield WindowsInterfaceRecord(
+                    **device_info,
+                    source="registry",
+                    last_connected=None,
+                    enabled=None,
+                    _target=self.target,
+                )
+
+    def _extract_network_device_config(self, interface_id: str) -> dict[str, str | list] | None:
+        config = {"search_domain": None, "gateway": [], "ip": [], "dns": [], "network": [], "metric": None}
```
Schamper, line 275:
```suggestion
        config = {}
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for name in ["DhcpDomain", "Domain"]:
+            if (value := _try_value(keys, name)) not in ["", None]:
+                config["search_domain"] = value
+
+        if (value := _try_value(keys, "InterfaceMetric")) not in ["", None]:
+            config["metric"] = value
+
+        if any(config.values()):
+            return config
```
Schamper, line 310:
```suggestion
        return config
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+        for keys in self.target.registry.keys(self.REGISTRY_KEY_CONTROLSET):
+            for subkey in keys.subkeys():
+                device_info = {}
+
+                if (net_cfg_instance_id := _try_value(subkey, "NetCfgInstanceId")) is None:
+                    # if no NetCfgInstanceId is found, skip this network interface
+                    continue
+
+                # Extract the network device configuration for given interface id
+                if (config := self._extract_network_device_config(net_cfg_instance_id)) is None:
```
Schamper, line 245:
```suggestion
                if not (config := self._extract_network_device_config(net_cfg_instance_id)):
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+        except RegistryKeyNotFoundError:
+            return None
+
+        if not len(keys):
+            return None
+
+        # Static lookups and assignments
+        for name in ["DhcpDefaultGateway", "DefaultGateway"]:
+            if (value := _try_value(keys, name)) not in ["", "0.0.0.0", None, []]:
+                config["gateway"].extend(value)
```
Schamper, line 288:
```suggestion
                config.setdefault("gateway", []).extend(value)
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for name in ["DhcpIPAddress", "IPAddress"]:
+            if (value := _try_value(keys, name)) not in ["", "0.0.0.0", None]:
+                config["ip"].append(value)
```
Schamper, line 292:
```suggestion
                config.setdefault("ip", []).append(value)
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for name in ["DhcpNameServer", "NameServer"]:
+            if (value := _try_value(keys, name)) not in ["", "0.0.0.0", None]:
+                config["dns"].append(value)
```
Schamper, line 296:
```suggestion
                config.setdefault"dns", []).append(value)
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for name in ["DhcpSubnetMask", "SubnetMask"]:
+            if (value := _try_value(keys, name)) not in ["", "0.0.0.0", None]:
+                config["network"].append(value)
```
Schamper, line 300:
```suggestion
                config.setdefault("network", []).append(value)
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for name in ["DhcpDomain", "Domain"]:
+            if (value := _try_value(keys, name)) not in ["", None]:
+                config["search_domain"] = value
```
Schamper, line 304: `Domain` can override `DhcpDomain`, is that intentional (in comparison with the other ones)?

Schamper (changes requested):

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,365 @@
...
+    OPTICAL_CHANNEL = 195
+    OPTICAL_TRANSPORT = 196
+    WWANPP = 243
+    WWANPP2 = 244
+
+
+def _try_value(subkey: RegistryKey, value: str) -> str | list | None:
+    try:
+        return subkey.value(value).value
+    except (RegistryValueNotFoundError, RegistryValueNotFoundError):
```
Schamper, line 223:
```suggestion
    except RegistryValueNotFoundError:
```

dissect/target/plugins/os/windows/network.py
Schamper: I don't really agree. Why does it need to be in the unit test? If you base the source of the data in the unit test to use this, and later on an error is introduced in this plugin, the unit test will still pass while in reality the plugin will no longer function. Not _everything_ has to be deduplicated just because it can be :smile:.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,365 @@
...
+
+                # Yield a record for each non-empty configuration
+                for conf in config:
+                    if conf:
+                        # Create a copy of device_info to avoid overwriting
+                        record_info = device_info.copy()
+                        record_info.update(conf)
+                        yield WindowsInterfaceRecord(
+                            **record_info,
+                            source="registry",
```
Schamper, line 279: Source should be the registry key path I believe.

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,365 @@
...
+                            last_connected=None,
```
Schamper, line 280: This is nowhere?

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,365 @@
...
+                            enabled=None,
```
Schamper, line 281: What's this?

dissect/target/plugins/os/windows/network.py
```diff
@@ -284,64 +297,69 @@ def _extract_network_device_config(self, interface_id: str) -> dict[str, str | l
...
+            dhcp_config.setdefault("subnetmask", []).append(dhcp_subnetmask)
+
+        dhcp_domain = _try_value(keys, "DhcpDomain")
+        if dhcp_domain not in ["", None]:
+            dhcp_config.setdefault("search_domain", []).append(dhcp_domain)
+
+        if len(dhcp_config) > 0:
+            dhcp_enable = _try_value(keys, "EnableDHCP")
+            if dhcp_enable not in ["", None, 1]:
+                dhcp_config.setdefault("dhcp", False)
```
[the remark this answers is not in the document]
Schamper: I don't really agree, this highly assumes a certain naming scheme out of our control.

tests/plugins/os/windows/test_network.py
```diff
@@ -0,0 +1,322 @@
...
+from dissect.target.target import Target
+
+
+@dataclass
+class MockRegVal:
+    name: str
+    value: str | int
+
+
+class MockRegistry:
```
Schamper, line 19: Can you take a note out of other plugins instead? This is a bit overkill. While nice, it's best of this was implemented as something general usable for all unit tests, not just this one. For now I'd stay stick with the (arguably more annoying and worse) registry mapping.

Schamper (changes requested):

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+class WindowsNetworkPlugin(NetworkPlugin):
+    REGISTRY_KEY_INTERFACE = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces\\"
+    REGISTRY_KEY_CONTROLSET = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}"
+    REGISTRY_KEY_CONNECTION = (
+        "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Network\\{4D36E972-E325-11CE-BFC1-08002BE10318}\\"
+    )
+
+    @internal
```
pyrco, line 233:
```suggestion
```

pyrco (changes requested):

cecinestpasunepipe (approved):

dissect/target/plugins/os/windows/network.py
```diff
@@ -0,0 +1,373 @@
...
+        REGISTRY_KEY_INTERFACE (str): Interface parameters for TCP/IP.
+        REGISTRY_KEY_CONTROLSET (str): Control set for network class.
+        REGISTRY_KEY_CONNECTION (str): Network connection settings.
+    """
+
+    REGISTRY_KEY_INTERFACE = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces\\"
+    REGISTRY_KEY_CONTROLSET = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}"
+    REGISTRY_KEY_CONNECTION = (
+        "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Network\\{4D36E972-E325-11CE-BFC1-08002BE10318}\\"
+    )
```
Schamper, line 241: Still same comment. This is not configurable and I think it distracts from the code to have these as constants up here as you only ever use them once.

Schamper (changes requested):

dissect/target/plugins/os/windows/network.py
cecinestpasunepipe: > Why does it need to be in the unit test?
Because it makes both the code and the test more readable

> the unit test will still pass
Only if there is a typo in the plugin constant, however if you replaced both the test will still fail to detect the regression, so what is the benefit?

dissect/target/helpers/record.py
```diff
@@ -144,14 +144,14 @@ def DynamicDescriptor(types):  # noqa
 )
 
 COMMON_INTERFACE_ELEMENTS = [
+    ("stringlist", "sources"),
```
Schamper, line 147: I'd just let the source be the single string to `HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}\\{key_name}` for simplicity. As that's the leading device key, and the rest is just additional metadata.

Schamper (changes requested):

pyrco (approved):

dissect/target/helpers/record.py
```diff
@@ -144,14 +144,14 @@ def DynamicDescriptor(types):  # noqa
 )
 
 COMMON_INTERFACE_ELEMENTS = [
+    ("string", "source"),
```
Schamper, line 147: Can you move this to the bottom in a new PR? @narimantos

PR #841, opened 2024-09-04
Add json output to target-query --list

### LINKED ISSUE #887: Add json output to `target-query --list`

JSCU-CNI:
This PR adds JSON output support to `target-query --list` using the `--json` flag. Fixes #887.


EinatFox:
Hi @JSCU-CNI , please open an issue and attache this PR so we can schedule a review into our sprints.

dissect/target/plugins/general/loaders.py
```diff
@@ -10,7 +12,8 @@ def check_compatible(self) -> None:
         pass
 
     @export(output="none")
-    def loaders(self):
+    @arg("--json", dest="output_json", action="store_true")
+    def loaders(self, output_json: bool = False) -> None:
```
Schamper, line 16: Why not output to just `json`? Does it conflict with the `--json` in the "main" argparse instance?

Schamper (changes requested):

dissect/target/plugins/general/loaders.py
Schamper: It doesn't really work as expected now anyways because of that. `target-query <target> -f loaders --json` doesn't work, only `target-query -l --json` works as expected. Changing the argument in the plugin to something else (`--as-json`) works as a bypass.

I'd rather fix it up in `query.py` so that it's possible to "re-use" arguments across plugins and the main CLI. That's a bit more complicated though, and probably also easier after #889. I'm not sure if you're willing to try and fix that in the current code, or rather use a workaround for now (I'd probably prefer `--as-json`) with some comments to linked issues on GitHub that will describe how to properly fix this.

Schamper: Can you make it consistently use `as_json` instead of `output_json` too, to keep it a bit in sync?

Schamper (approved):

PR #842, opened 2024-09-04
Fix 3.12 ET and datetime deprecation warnings

JSCU-CNI:
This PR, together with https://github.com/fox-it/flow.record/pull/136 should fix all pytest deprecation warnings when running CPython 3.12.

### DIFF
--- a/dissect/target/plugins/os/windows/adpolicy.py
+++ b/dissect/target/plugins/os/windows/adpolicy.py
@@ -70,8 +70,7 @@ class ADPolicyPlugin(Plugin):
                 tree = ElementTree.fromstring(xml)
                 for task in tree.findall(".//{*}Task"):
                     # https://github.com/python/cpython/issues/83122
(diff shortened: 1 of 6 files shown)


dissect/target/plugins/os/windows/adpolicy.py
```diff
@@ -69,7 +69,11 @@ def _xmltasks(self, policy_dir):
                 xml = task_file.read_text()
                 tree = ElementTree.fromstring(xml)
                 for task in tree.findall(".//{*}Task"):
-                    properties = task.find("Properties") or task
+                    # https://github.com/python/cpython/issues/83122
+                    properties = task.find("Properties")
+                    if properties is None:
```
Schamper, line 74:
```suggestion
                    if (properties := task.find("Properties")) is None:
```
Does this work?

Schamper (changes requested):

PR #847, opened 2024-09-11
Dis 3281/improved security scheme

### LINKED ISSUE #846: Targetd: Improved security scheme
In case of an advanced hacker, Targetd needs an extra security layer to make sure agents cannot get compromised. We already have seceurity to some extend (certificates, ACL) but an active hacker can still emit false information on behalf of other hosts (although this requires massive effort). The proposed security scheme mitigates this.

1. Agent generates random username and random password when registering as a service and stores this in the registry (and processes this as an argument on the cli)
1. Agent connects using random username and password, fails
1. Agent retries with group username or password (if extra security is not activated, flow ends here)
1. Controller sends MQTT-command “callid“
1. Agents sends username,  password, and hostname in “registration“ message
1. Controller outputs users to be added to ACL (manually, can be automated later)

How it works:

After deploying agents, controller gives agents a limited time frame to register to the broker with proper credentials. After registering the agents can be used. This gives an active hacker only the time between deployment and registration to hack an agent and fake the registration response. It only gives the hacker the limited registration time frame to emit the correct response. 

To-do:

1.  Add random username/password to start_stop_service() in util.c
1. Connect with new user/pass
1. Implement register command in agent and loader

twiggler:
This is a companion PR of the agent and cannot be tested stand-alone.

To test together with the agent,

1. Generate an agent
2. Deploy it on a machine.
3. Start the broker
4. Run the command below

```
target-query all -L mqtt --mqtt-port=443 --mqtt-broker=0.0.0.0 --mqtt-ca=domain.crt --mqtt-crt=/tmp/cert.crt --mqtt-key=/tmp/private.key --mqtt-case=case1 --mqtt-peers=100 --mqtt-username=ag_case1_roel --mqtt-password --mqtt-diag --mqtt-command callid  -f hostname
```

closes #846

### DIFF
--- a/dissect/target/loaders/mqtt.py
+++ b/dissect/target/loaders/mqtt.py
@@ -341,8 +341,12 @@ class Broker:
             self.mqtt_client.subscribe(f"{self.case}/{host}/DISKS")
             self.mqtt_client.subscribe(f"{self.case}/{host}/READ/#")
             if self.command is not None:
+                self.mqtt_client.subscribe(f"{self.case}/{host}/CALLID")
                 self.mqtt_client.publish(f"{self.case}/{host}/COMM", self.command.encode("utf-8"))
             time.sleep(1)
+
+    def _on_callId(self, hostname: str, payload: bytes) -> None:
+        print(f"CALLID: {payload.decode("utf-8")}")
 
     def _on_log(self, client: mqtt.Client, userdata: Any, log_level: int, message: str) -> None:
         log.debug(message)
@@ -365,6 +369,8 @@ class Broker:
             self._on_read(hostname, tokens, msg.payload)
         elif response == "ID":
             self._on_id(hostname, msg.payload)
+        elif response == "CALLID":
+            self._on_callId(hostname, msg.payload)
 
     def seek(self, host: str, disk_id: int, offset: int, flength: int, optimization_strategy: int) -> None:
         length = int(flength / self.factor)
--- a/dissect/target/loaders/mqtt.py
+++ b/dissect/target/loaders/mqtt.py
@@ -345,7 +345,7 @@ class Broker:
                 self.mqtt_client.publish(f"{self.case}/{host}/COMM", self.command.encode("utf-8"))
             time.sleep(1)
 
-    def _on_callId(self, hostname: str, payload: bytes) -> None:
+    def _on_call_id(self, hostname: str, payload: bytes) -> None:
         try:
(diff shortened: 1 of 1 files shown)


dissect/target/loaders/mqtt.py
```diff
@@ -341,9 +341,20 @@ def _on_id(self, hostname: str, payload: bytes) -> None:
...
 
+    def _on_callId(self, hostname: str, payload: bytes) -> None:
+        try:
+            decoded_payload = payload.decode("utf-8")
+        except UnicodeDecodeError as e:
+            log.error(f"Failed to decode payload for hostname {hostname}: {e}")
+            return
+
+        # The payload with the username and password is comma separated
+        print(f'CALLID: "{hostname}",{decoded_payload}')
```
cecinestpasunepipe, line 356: please leave out the prefix here

dissect/target/loaders/mqtt.py
```diff
@@ -341,9 +341,20 @@ def _on_id(self, hostname: str, payload: bytes) -> None:
             self.mqtt_client.subscribe(f"{self.case}/{host}/DISKS")
             self.mqtt_client.subscribe(f"{self.case}/{host}/READ/#")
             if self.command is not None:
+                self.mqtt_client.subscribe(f"{self.case}/{host}/CALLID")
                 self.mqtt_client.publish(f"{self.case}/{host}/COMM", self.command.encode("utf-8"))
             time.sleep(1)
 
+    def _on_callId(self, hostname: str, payload: bytes) -> None:
```
cecinestpasunepipe, line 348:
```suggestion
    def _on_call_id(self, hostname: str, payload: bytes) -> None:
```
please use snake_case

dissect/target/loaders/mqtt.py
```diff
@@ -365,6 +376,8 @@ def _on_message(self, client: mqtt.Client, userdata: Any, msg: mqtt.client.MQTTM
             self._on_read(hostname, tokens, msg.payload)
         elif response == "ID":
             self._on_id(hostname, msg.payload)
+        elif response == "CALLID":
+            self._on_callId(hostname, msg.payload)
```
cecinestpasunepipe, line 380: change function call here please to match updated version

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #848, opened 2024-09-13
Add birthtime, blocksize and blks to stat output for NTFS

Miauwkeru:
st_block uses the realsize if defined, oterwise it uses the size
the real size on disk is only defined if the file is not a resident file.

st_birthtime and st_birthtime_ns now use the creation_time from a MFT Record

Other timestamps where incorrectly used, and have been changed:
st_mtime: changed from using last_change_time to last_modification_time

### DIFF
--- a/dissect/target/filesystems/btrfs.py
+++ b/dissect/target/filesystems/btrfs.py
@@ -181,6 +181,9 @@ class BtrfsFilesystemEntry(FilesystemEntry):
 
         # Add block information of the filesystem
         st_info.st_blksize = entry.btrfs.sector_size
-        st_info.st_blocks = math.ceil(entry.size / st_info.st_blksize)
+
+        st_info.st_blocks = 0
+        if not self.is_dir():
+            st_info.st_blocks = (st_info.st_blksize // 512) * math.ceil(st_info.st_size / st_info.st_blksize)
 
         return st_info
--- a/tests/filesystems/test_btrfs.py
+++ b/tests/filesystems/test_btrfs.py
@@ -10,9 +10,9 @@ from dissect.target.filesystems.btrfs import BtrfsFilesystemEntry
 @pytest.mark.parametrize(
     "sector_size, filesize, expected_blocks",
     [
-        (0x1000, 0x343, 0x1),
-        (0x1000, 0x1000, 0x1),
-        (0x1000, 0x1001, 0x2),
+        (0x1000, 0x343, 0x8),
+        (0x1000, 0x1000, 0x8),
+        (0x1000, 0x1001, 0x10),
     ],
     ids=["lower", "equal", "greater"],
 )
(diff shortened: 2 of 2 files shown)


Schamper:
image

Schamper:
The entire filesystem layer is designed to be consistent with Python since the beginning of Dissect.

Schamper:
https://github.com/python/cpython/blob/3.12/Modules/posixmodule.c#L2142-L2145
https://docs.python.org/3/library/os.html#os.DirEntry.stat (deprecation note)

The only thing missing currently is setting `st_birthtime` and `st_birthtime_ns` to their proper values on Windows.

Btrfs and other filesystems don't have this.

Schamper:
> st_birthtime gets set here right?: https://github.com/python/cpython/blob/aee219f4558dda619bd86e4b0e028ce47a5e4b77/Python/fileutils.c#L1122

In our code.

pyrco:
Yes, it seems to be correct now.

Both `st_birthtime` and `st_ctime` are now set to the `CreationTime` attribute, as is done by Python up to and including Py3.12.

The change for `st_mtime` to the `LastModificationTime` (which is called `LastWriteTime` in the Windows `fileapi.h`, and is the name Python uses) is also in line with how python handles this timestamp.

Horofic:
Ack on the timestamps. Is @Poeloe still on for a code-review?

dissect/target/filesystems/btrfs.py
```diff
@@ -181,6 +181,9 @@ def lstat(self) -> fsutil.stat_result:
 
         # Add block information of the filesystem
         st_info.st_blksize = entry.btrfs.sector_size
-        st_info.st_blocks = math.ceil(entry.size / st_info.st_blksize)
+
+        st_info.st_blocks = 0
+        if not self.is_dir():
+            st_info.st_blocks = math.ceil(st_info.st_blksize / 512) * math.ceil(st_info.st_size / st_info.st_blksize)
```
pyrco, line 187: I'm not sure this double ceiling is correct, it could lead to a really inflated number of blocks, e.g.:

```
st_size: 256
st_blksize: 128
->
math.ceil(128/512) * math.ceil(256/128) = math.ceil(1/4) * math.ceil(2) = 1 * 2 = 2
```
Inflates the apparent size by a factor of 4

```
st_size: 1024
st_blksize: 768
->
math.ceil(768/512) * math.ceil(1024/768) = math.ceil(3/2) * math.ceil(4/3) = 2 * 2 = 4
```
Inflates the apparent size by a factor of 2

If I read and interpret the various sources of what `st_blocks` should be it seems to me it represents the size *on disk* in blocks (which are 512 bytes in our implementation).
The POSIX/opengroup for example says: "Number of blocks allocated for this object."

The on disk is important in the sense that if files that are e.g. sparse their size on disk is smaller than their actual size, the same holds for example for files that are compressed on the filesystem layer.

So I think we can just do a `math.ceil(st_info.st_size / 512)` here. It also means that for symbolic links `st_blocks` should be 0, unless the symbolic link data is not (just) contained in the inode structure.

pyrco (changes requested):

pyrco (approved):

PR #850, opened 2024-09-16
Improve argument parsing of the mqtt loader

### LINKED ISSUE #849: Targetd: Improve argument parser MQTT -loader and remove old Targetd-loader
The argument parser is a bit rough, adjust as follows:

- emit error if peers < 1
- case must be a string (and not contain MQTT chars) with same constraints at case naming in acquire deploy
- check if broker is a valid IP
- key must be a path type
- crt must be a path type
- ca must be a path type
- command must be a string type
- username must be a string type
- check if port no. is valid
- Improve help strings

Also, for some reason why decided to keep targetd 1.0 as a loader. This loader can be safely removed.

Make sure that removing the loader does not affect other loaders in bootstrapping.

Adjust tests if necessary.

twiggler:
This PR improved upon the argument parsing of the mqtt loader.
More specifically, 

1. It checks for invalid values.  For example, check that the specified number of peers is strictly positive
2. It adds more helpful error messages.

Additionally,

1. It deletes the legacy targetd loader (#849 was about _two_ things, now not entirely sure how they are related).
2. Fixes a typing issue.
3. Removes the hostname prefix of `CALLID` messages 
4. Simplify mocking of MQTTClient

Closes #849


dissect/target/loaders/mqtt.py
```diff
@@ -423,13 +426,147 @@ def connect(self) -> None:
...
 
 
-@arg("--mqtt-peers", type=int, dest="peers", help="minimum number of peers to await for first alias")
-@arg("--mqtt-case", dest="case", help="case name (broker will determine if you are allowed to access this data)")
-@arg("--mqtt-port", type=int, dest="port", help="broker connection port")
-@arg("--mqtt-broker", dest="broker", help="broker ip-address")
-@arg("--mqtt-key", dest="key", help="private key file")
-@arg("--mqtt-crt", dest="crt", help="client certificate file")
-@arg("--mqtt-ca", dest="ca", help="certificate authority file")
+def strictly_positive(value: str) -> int:
```
twiggler, line 429: For reviewer: I think these parse functions are better moved to another module so that they don´t distract from the main functionality.

Any suggestions?

dissect/target/loaders/mqtt.py
```diff
@@ -423,13 +426,147 @@ def connect(self) -> None:
...
+@arg("--mqtt-peers", type=strictly_positive, dest="peers", help="minimum number of peers to await for first alias")
+@arg(
+    "--mqtt-case",
+    type=case,
+    dest="case",
+    help="case name (broker will determine if you are allowed to access this data)",
+)
+@arg("--mqtt-port", type=port, dest="port", help="broker connection port")
+@arg("--mqtt-broker", type=host_name_or_ip_address, dest="broker", help="broker ip-address")
+@arg("--mqtt-key", type=file_exists, dest="key", help="private key file")
```
twiggler, line 567: For reviewer: In other files these typically have `type=path`.

The advantage of file_exists is that you we get an exception immediately, and not in a deep stack trace where the error might be more difficult to comprehend. 

Let me know what you think.

dissect/target/loaders/mqtt.py
```diff
@@ -280,7 +283,7 @@ class Broker:
     factor = 1
 
     def __init__(
-        self, broker: Broker, port: str, key: str, crt: str, ca: str, case: str, username: str, password: str, **kwargs
+        self, broker: str, port: str, key: str, crt: str, ca: str, case: str, username: str, password: str, **kwargs
```
twiggler, line 286: For reviewer: was typo 00psie

dissect/target/loaders/mqtt.py
```diff
@@ -423,13 +426,147 @@ def connect(self) -> None:
...
+def ip_address(value: str) -> bool:
+    """Checks if the value is a valid IP address."""
+    try:
+        ipaddress.ip_address(value)
+        return True
+    except ValueError:
+        return False
+
+
+def host_name_or_ip_address(value: str) -> str:
```
twiggler, line 485: For reviewer: alternatively you could write a parser combinator 

`def or_parser(parser1: Callable[[str], T1], parser2: Callable[[str], T2]) -> Callable[[str], T1 | T2]` 

but this is easier to understand, and there are probably python libraries which do that much better.

dissect/target/helpers/targetd.py
twiggler, line 1: For reviewer: deleted because the targetd loader is gone.

tests/loaders/test_mqtt.py
```diff
@@ -68,21 +69,10 @@ def read(self, *args) -> MockSeekMessage | None:
...
 
-        yield mock_paho
-
-
-@pytest.fixture
-def mock_client(mock_paho: MagicMock) -> Iterator[MagicMock]:
-    mock_client = MQTTMock()
-    mock_paho.mqtt.client.Client.return_value = mock_client
-    yield mock_client
+    monkeypatch.setattr(mqtt, "Client", MQTTMock)
```
twiggler, line 75: For reviewer: this seems simpler,  since we do not yet need to manipulate the mock instance of `MQTTMock` inside the test.

The other approach also required us to do our imports from `dissect.target.loaders.mqtt` inside the test.

dissect/target/loaders/mqtt.py
```diff
@@ -423,16 +425,149 @@ def connect(self) -> None:
...
+    Returns:
+        int: The validated integer value.
+
+    Raises:
+        argparse.ArgumentTypeError: If the value is not a strictly positive integer.
+    """
+    try:
+        strictly_positive_value = int(value)
+        if strictly_positive_value < 1:
+            raise argparse.ArgumentTypeError("Value must be strictly positive.")
```
pyrco, line 446:
```suggestion
            raise argparse.ArgumentTypeError("Value must be larger then or equal to 1.")
```
I'd phrase it like this.

tests/loaders/test_mqtt.py
pyrco, line 75: Could you move the import to the generic imports and do the `monkeypatch()` using a `with ...:` in `test_remote_loader_stream()`?

pyrco (changes requested):
Probably better to remove the host/ip checks. I don't think it is worth the aded complexity as the broker connection could fail due to a number of reasons anyway.

tests/loaders/test_mqtt.py
twiggler: - Moved the import to module scope
- Moved applying of the mock to the test.
- Apparently, it is unnecessary to use a context manager. From the monkeypatch documentation:

https://docs.pytest.org/en/stable/how-to/monkeypatch.html

"All modifications will be undone after the requesting test function or fixture has finished. The raising parameter determines if a KeyError or AttributeError will be raised if the target of the set/deletion operation does not exist."

pyrco (approved):

PR #852, opened 2024-09-17
Add Unix trash plugin

JSCU-CNI:
Since we [heard](https://github.com/fox-it/dissect.target/pull/715) that this will be the year of the Linux Desktop, we added support for parsing GNOME `.Trash` folder artifacts on Unix systems.


dissect/target/plugins/os/unix/trash.py
```diff
@@ -0,0 +1,124 @@
...
+                if (path := user_details.home_path.joinpath(trash_path)).exists():
+                    yield user_details, path
+
+    def check_compatible(self) -> None:
+        if not self.trashes:
+            raise UnsupportedPluginError("No Trash folder(s) found")
+
+    @export(record=TrashRecord)
+    @alias(name="recyclebin")
+    def trash(self) -> Iterator[TrashRecord]:
```
Horofic, line 60: Please add some record field documentation. Could you also document some of the caveats here as well? For example the timestamp technicality for expunged files.

Horofic (changes requested):

tests/plugins/os/unix/test_trash.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_gnome_trash(target_unix_users: Target, fs_unix: VirtualFilesystem) -> None:
+    """test if GNOME Trash plugin finds all deleted files including recursively deleted folders and expunged items."""
+
+    fs_unix.map_file_fh("etc/hostname", BytesIO(b"hostname"))
+    fs_unix.map_dir("home/user/.local/share/Trash", absolute_path("_data/plugins/os/unix/trash"))
+    target_unix_users.add_plugin(UnixPlugin)
+    target_unix_users.add_plugin(GnomeTrashPlugin)
+
```
Horofic, line 18:
```suggestion

    assert hasattr(target_unix_users, "recyclebin")

```

Horofic: Forgot about that one, even though I recently used it. Feel free to change it to `target_unix_users.has_function("recyclebin")`

dissect/target/plugins/os/unix/trash.py
Horofic: Something along the lines of what is shown below. This will be used in the plugin documentation and will show up when you use `target-query -f trash -h`.

https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/apps/texteditor/windowsnotepad.py#L307-#L316

Horofic: Yes, that would be nice!

dissect/target/plugins/os/unix/trash.py
```diff
@@ -0,0 +1,121 @@
...
+    PATHS = [
+        # Default $XDG_DATA_HOME/Trash
+        ".local/share/Trash",
+    ]
+
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self.trashes = list(self.garbage_collector())
+
+    def garbage_collector(self) -> Iterator[tuple[UserDetails, TargetPath]]:
```
Horofic, line 47:
```suggestion
    def _garbage_collector(self) -> Iterator[tuple[UserDetails, TargetPath]]:
```
Although I love the name, this is somewhat more consistent with the rest of these type of functions.

Horofic (changes requested):

dissect/target/tools/utils.py
```diff
@@ -91,6 +91,10 @@ def generate_argparse_for_unbound_method(
         raise ValueError(f"Value `{method}` is not an unbound plugin method")
 
     desc = method.__doc__ or docs.get_func_description(method, with_docstrings=True)
+
+    if "\n" in desc:
+        desc = desc.split("\n")[0] + "\n" + textwrap.dedent("\n".join(desc.split("\n")[1:]))
```
Horofic, line 96:
```suggestion
        desc = inspect.cleandoc(desc)
```
This should do the same :).

Horofic (changes requested):

Horofic (approved):

PR #857, opened 2024-09-19
Small log improvements

### LINKED ISSUE #880: `modify_record` noisey when using `--hash` with filesystem plugins
Some plugins generate records with paths pointing to non existing files, which causes log spam in combination with `--hash`.

### LINKED ISSUE #881: Dissect attempts to load `.targetcfg.py` from child paths

JSCU-CNI:
This PR addresses two small issues:

* The hash record modifier could generate a lot of warnings when combining the walkfs plugin with `--hash` as a lot of files can be broken. this log is now demoted to INFO for `FileNotFoundError`s as that exception is not really interesting in those instances.
* When entering a child target using `--children` or target-shell's `enter` the path of said child was appended to the locations of where to look for targetcfg files. That obviously does not work as that path is virtual inside the main target. This is now fixed.

Fixes #880 and fixes #881.


EinatFox:
Hi @JSCU-CNI , please open an issue and attache this PR so we can schedule a review into our sprints.

dissect/target/helpers/record_modifier.py
```diff
@@ -94,6 +94,13 @@ def modify_record(target: Target, record: Record, modifier_function: ModifierFun
...
             _record = modifier_function(field_name, resolved_path)
+
+        except FileNotFoundError as e:
+            target.log.info(
+                "Unable to modify record '%s' with function '%s': %s", record._desc.name, modifier_function.__name__, e
+            )
+            target.log.debug("", exc_info=e)
+
         except FilesystemError as e:
             target.log.warning(
```
Schamper, line 105:
```suggestion
        except FilesystemError as e:
            level = logging.INFO if isinstance(e, FileNotFoundError) else logging.WARNING
            target.log.log(
                level,
```
Maybe a bit more concise. I played around with this or setting `target.log.warning`/`target.log.info` to a variable but I'm not sure if I have a specific preference.

Schamper (changes requested):

Schamper (approved):

PR #858, opened 2024-09-19
Correctly mount file mount points in Overlay2Filesystem

### LINKED ISSUE #856: Docker MountPoint files incorrectly mapped as directories in Overlay2Filesystem
Running `target-query -f walkfs --hash --children target.img` crashes if the target contains docker children with file mount points. It seems we incorrectly map files as directories in Overlay2Filesystem.

Eventually `scandir` in `filesystems/dir.py` will error with a `NotADirectoryError` on the path that was added as a layer since it is actually a file.

JSCU-CNI:
This PR fixes #856.


Schamper:
Could you add a unit test? 😄

Schamper (approved):

Schamper (dismissed):

Schamper (approved):

PR #861, opened 2024-09-23
Unfold ADS macb records

cecinestpasunepipe:
Avoids hiding several ADS MACBs behind the same TS.

### DIFF
--- a/dissect/target/plugins/filesystem/ntfs/mft.py
+++ b/dissect/target/plugins/filesystem/ntfs/mft.py
@@ -152,7 +152,7 @@ class MftPlugin(Plugin):
         "--macb",
         group="fmt",
         action="store_true",
-        help="compacts MFT timestamps into MACB bitfield (format: standard|ads/filename)",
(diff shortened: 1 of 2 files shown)


Schamper:
Can you add a regression test?

dissect/target/plugins/filesystem/ntfs/mft.py
```diff
@@ -151,7 +152,7 @@ def check_compatible(self) -> None:
         "--macb",
         group="fmt",
         action="store_true",
-        help="compacts the MFT entry timestamps into aggregated records with MACB bitfield",
+        help="compacts MFT timestamps into MACB bitfield (format: standard|ads / filename)",
```
Poeloe, line 155: Do you think this maybe makes the format more explanatory?
```suggestion
        help="compacts MFT timestamps into MACB bitfield (format: MACB[standard|ads]/MACB[filename])",
```

Poeloe (approved):
LGTM!

PR #869, opened 2024-10-01
Move helpers/network_manager.py to plugins/os/unix/linux/network_manager.py

Miauwkeru:
- **Move network_manager.py to os/unix**
- **Update the typing on network_managers.py**


dissect/target/plugins/os/unix/network_managers.py
pyrco, line 1: I think this file should be moved under the `./linux/` subdir as it is very Linux specific.

pyrco (changes requested):

pyrco (approved):

pyrco (approved):

PR #871, opened 2024-10-01
Add support for runtime aliases to target_shell

cecinestpasunepipe:
(DIS-2965)

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -236,6 +236,14 @@ class ExtendedCmd(cmd.Cmd):
     def complete_man(self, *args) -> list[str]:
         return cmd.Cmd.complete_help(self, *args)
 
+    def do_unalias(self, aliased: str) -> bool:
+        """delete runtime alias"""
+        if aliased in self._runtime_aliases:
+            del self._runtime_aliases[aliased]
+        else:
+            print(f"alias {aliased} not found")
+        return False
+
     def do_alias(self, line: str) -> bool:
         """create a runtime alias"""
         args = list(shlex.shlex(line, posix=True))
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -249,15 +249,27 @@ class ExtendedCmd(cmd.Cmd):
     def do_alias(self, line: str) -> bool:
         """create a runtime alias"""
         args = list(shlex.shlex(line, posix=True))
-        if len(args) > 2 and len(args) % 3 == 0:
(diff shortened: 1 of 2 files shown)


twiggler (changes requested):
QA:

When entering the unknown command `ee`, 

**Actual** I am getting the following error when entering an unknown command:

image
 
 
**Expected**:

```
*** Unknown syntax: ee
```

dissect/target/tools/shell.py
```diff
@@ -230,6 +236,27 @@ def do_man(self, line: str) -> bool:
     def complete_man(self, *args) -> list[str]:
         return cmd.Cmd.complete_help(self, *args)
 
+    def do_unalias(self, aliased: str) -> bool:
+        """delete runtime alias"""
+        if aliased in self._runtime_aliases:
```
twiggler, line 241: Perhaps add a small comment to each case.

tests/tools/test_shell.py
```diff
@@ -270,3 +270,14 @@ def test_shell_cmd_alias(monkeypatch: pytest.MonkeyPatch, capsys: pytest.Capture
...
     ll_out, _ = run_target_shell(monkeypatch, capsys, target_path, "ll")
     assert ls_la_out == ll_out
+
+
+def test_shell_cmd_alias_runtime(monkeypatch: pytest.MonkeyPatch, capsys: pytest.CaptureFixture) -> None:
+    """test if alias commands call their parent attribute correctly."""
+    target_path = absolute_path("_data/tools/info/image.tar")
+
+    # 'list' and 'ls' should return the same output after runtime aliasing
+    list_out, _ = run_target_shell(monkeypatch, capsys, target_path, "alias list=ls\nlist")
```
twiggler, line 280: Perhaps add a test for

1. Output of alias with 1 argument.
1. Running alias without any arguments

dissect/target/tools/shell.py
```diff
@@ -230,6 +236,27 @@ def do_man(self, line: str) -> bool:
...
+            del self._runtime_aliases[aliased]
```
twiggler, line 242: There is slightly different behavior when using alias with more than 3 arguments. 
I think we ignore them, but bash accepts them

twiggler (approved):

twiggler:
@cecinestpasunepipe One small QA issue

When I run `alias "./"="echo roel"`,  the alias is defined (although it cannot be used):

image

Not that important but a bit weird.

twiggler (approved):

PR #892, opened 2024-10-08
Explicitly mark home directories as having a posix or windows flavor

### LINKED ISSUE #891: Explicit path comparison for user home directories
Ensure that the paths to the home directory on both Unix and Windows are a `posix_path` and `windows_path`, respectively. This could cause problems, such as manifested in #866.
 
Also, adjust the unit test for the home directory on OSX  by removing the explicit type check on `posix_path`.

twiggler:
- Explicitly mark home directories as either `posix_path` or `windows_path`
- Add missing unit test for parsing of users on windows and unix

Closes #891

### DIFF
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -6,6 +6,8 @@ import uuid
 from struct import unpack
 from typing import Iterator, Optional, Union
 
+from flow.record.fieldtypes import posix_path
+
 from dissect.target.filesystem import Filesystem
 from dissect.target.helpers.fsutil import TargetPath
 from dissect.target.helpers.record import UnixUserRecord
@@ -71,7 +73,7 @@ class UnixPlugin(OSPlugin):
                         uid=pwent.get(2),
                         gid=pwent.get(3),
                         gecos=pwent.get(4),
-                        home=self.target.fs.path(pwent.get(5)),
+                        home=posix_path(pwent.get(5)),
                         shell=pwent.get(6),
                         source=passwd_file,
                         _target=self.target,
@@ -115,7 +117,7 @@ class UnixPlugin(OSPlugin):
 
                 yield UnixUserRecord(
                     name=user["name"],
-                    home=user["home"],
+                    home=posix_path(user["home"]),
                     shell=user["shell"],
                     source="/var/log/syslog",
                     _target=self.target,
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -4,7 +4,7 @@ import logging
 import re
 import uuid
 from struct import unpack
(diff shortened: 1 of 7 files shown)


dissect/target/plugins/os/unix/_os.py
```diff
@@ -6,6 +6,8 @@
 from struct import unpack
 from typing import Iterator, Optional, Union
```
Schamper, line 7: Could you update type hints in this file to modern variations using `|`?

dissect/target/plugins/os/windows/_os.py
```diff
@@ -4,6 +4,8 @@
 import struct
 from typing import Any, Iterator, Optional
```
Schamper, line 5: Could you update type hints in this file to modern variations using `|`?

Schamper (changes requested):

tests/plugins/os/unix/bsd/citrix/test__os.py
```diff
@@ -51,15 +53,15 @@ def test_citrix_os(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
...
 
     assert users[6].name == "root"  # User entry for /root, from the config
-    assert users[6].home == "/root"
+    assert users[6].home == posix_path("/root")
     assert users[6].shell is None
 
     assert users[7].name == "root"  # User entry for /root, from /etc/passwd
-    assert users[7].home == "/root"
+    assert users[7].home == posix_path("/root")
     assert users[7].shell == "/usr/bin/bash"
```
Schamper, line 67: Misc comment that may needs to be it's own PR, but I think `shell` should also be a proper path.

dissect/target/plugins/os/unix/_os.py
```diff
@@ -4,7 +4,7 @@
 import re
 import uuid
 from struct import unpack
-from typing import Iterator, Optional, Union
+from typing import Iterator, Union
```
Schamper, line 7: Can you also replace `Union`?

Schamper (changes requested):
Only small question, looks good otherwise.

tests/plugins/os/unix/bsd/citrix/test__os.py
twiggler: Yeah, probably.

It is marked as a string in the record definition. 
Because it is a string, at least it won´t suffer from the issue we had where the path got converted to the flavor of the  host OS. 
But it is semantically clearer to mark it as a path.

Will think about it some more and create a ticket.

dissect/target/plugins/os/unix/_os.py
twiggler: Replaced

Schamper (approved):
If you make the title slightly more descriptive this looks good to me!

PR #897, opened 2024-10-09
Add Python 3.13 compatibility

### LINKED ISSUE #895: CPython 3.13 support
CPython 3.13.0 was [released](https://www.python.org/downloads/release/python-3130/) on 2024-10-07. This ticket can be used to track progress in making dissect compatible with Python 3.13.

JSCU-CNI:
This PR aims to implement a TargetPath 3.13 variant. Fixes #895 and currently WIP.

- [x] Add 3.13 variant
- [x] Fix `scandir`
- [x] Fix `glob`
- [x] Fix `rglob`
- [x] Do we still need flavours?
- [x] ...

Resources:
- https://docs.python.org/3.13/whatsnew/3.13.html#pathlib
- https://github.com/python/cpython/blob/v3.13.0/Lib/pathlib/_local.py
- https://github.com/python/cpython/blob/v3.12.7/Lib/pathlib.py
- https://github.com/python/cpython/blob/v3.13.0/Lib/glob.py
- https://github.com/python/cpython/blob/v3.12.7/Lib/glob.py


Schamper:
Fixed up the rest, I believe only `flow.record` needs to add compatibility with 3.13 (`path` type).

Schamper:
Depends on https://github.com/fox-it/flow.record/pull/154

Schamper:
I didn't run the entire test suite yet with the flow.record fixes, so that probably still requires some looking into 😄

Schamper:
I saw that deprecation warning as well when looking at the pathlib changes, but the alternative is not really a good fit for Dissect. We don't really use that function anyway beside a unit test, so maybe remove it from the unit test and add a `fsutil.isreserved()` for completeness sake?

Schamper:
We have to bump that version anyway, so we'll just have to wait for a new release.

tests/test_container.py
```diff
@@ -1,6 +1,7 @@
 import struct
```
Horofic, line 1: For the Union types in `test_open_inputs()`
```suggestion
from __future__ import annotations

import struct
```

Horofic (changes requested):

dissect/target/helpers/compat/path_313.py
```diff
@@ -0,0 +1,399 @@
...
+            path_args.append(arg)
+
+        super().__init__(*path_args)
+        self._fs = fs
+        self.parser = _DissectParser(alt_separator=fs.alt_separator, case_sensitive=fs.case_sensitive)
+
+    def with_segments(self, *pathsegments) -> TargetPath:
+        return type(self)(self._fs, *pathsegments)
+
+    # NOTE: This is copied from pathlib/_local.py but turned into an instance method so we get access to the correct flavour
```
Horofic, line 118: Linter fix
```suggestion
    # NOTE: This is copied from pathlib/_local.py 
    # but turned into an instance method so we get access to the correct flavour
```

Horofic (approved):

Horofic (approved):

Schamper:
Can be merged after a new release of `flow.record`.

Schamper (approved):

PR #901, opened 2024-10-15
Add support for ISO 8601 formatted auth log timestamps

JSCU-CNI:
This PR builds on top of #860 and introduces support for ISO 8601 formatted timestamps in `auth.log` files which can (finally :partying_face:) be found in newer Debian and Ubuntu systems. Fixes #903.

PR #860, opened 2024-09-20
Refactor AuthLogPlugin

Poeloe:
Fixes #286

PR #907, opened 2024-10-16
Add support for ISO 8601 timestamps in syslogs

### LINKED ISSUE #909: Syslog plugin broken on recent Debian and Ubuntu distros

JSCU-CNI:
This PR adds support for ISO 8601 timestamps in unix syslogs introduced in recent versions of Debian and Ubuntu distributions. The helper functions have been copied from #901. Once #860 is merged in main I propose to let the auth plugin import those functions from the helper file. Fixes #909.

### DIFF
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -12,7 +12,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import BinaryIO, Iterator, Optional, TextIO, Union
 
-from dissect.regf import regf
+from dissect.regf import c_regf, regf
 
 from dissect.target.exceptions import (
     RegistryError,
@@ -31,16 +31,33 @@ ValueType = Union[int, str, bytes, list[str]]
 
 
 class RegistryValueType(IntEnum):
-    NONE = regf.REG_NONE
-    SZ = regf.REG_SZ
-    EXPAND_SZ = regf.REG_EXPAND_SZ
-    BINARY = regf.REG_BINARY
-    DWORD = regf.REG_DWORD
-    DWORD_BIG_ENDIAN = regf.REG_DWORD_BIG_ENDIAN
-    MULTI_SZ = regf.REG_MULTI_SZ
-    FULL_RESOURCE_DESCRIPTOR = regf.REG_FULL_RESOURCE_DESCRIPTOR
-    RESOURCE_REQUIREMENTS_LIST = regf.REG_RESOURCE_REQUIREMENTS_LIST
-    QWORD = regf.REG_QWORD
+    """Registry value types as defined in ``winnt.h``.
+
+    Resources:
+        - https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types
+        - https://github.com/fox-it/dissect.regf/blob/main/dissect/regf/c_regf.py
+    """
+
+    NONE = c_regf.REG_NONE
+    SZ = c_regf.REG_SZ
+    EXPAND_SZ = c_regf.REG_EXPAND_SZ
+    BINARY = c_regf.REG_BINARY
+    DWORD = c_regf.REG_DWORD
+    DWORD_BIG_ENDIAN = c_regf.REG_DWORD_BIG_ENDIAN
+    LINK = c_regf.REG_LINK
+    MULTI_SZ = c_regf.REG_MULTI_SZ
+    RESOURCE_LIST = c_regf.REG_RESOURCE_LIST
+    FULL_RESOURCE_DESCRIPTOR = c_regf.REG_FULL_RESOURCE_DESCRIPTOR
+    RESOURCE_REQUIREMENTS_LIST = c_regf.REG_RESOURCE_REQUIREMENTS_LIST
+    QWORD = c_regf.REG_QWORD
+
+    @classmethod
+    def _missing_(cls, value: int) -> IntEnum:
+        # Allow values other than defined members
+        member = int.__new__(cls, value)
+        member._name_ = None
+        member._value_ = value
+        return member
 
 
 class RegistryHive:
--- a/dissect/target/loaders/itunes.py
+++ b/dissect/target/loaders/itunes.py
@@ -163,8 +163,10 @@ class ITunesBackup:
 
     def files(self) -> Iterator[FileInfo]:
         """Iterate all the files in this backup."""
-        for row in self.manifest_db.table("Files").rows():
-            yield FileInfo(self, row.fileID, row.domain, row.relativePath, row.flags, row.file)
+
+        if table := self.manifest_db.table("Files"):
+            for row in table.rows():
+                yield FileInfo(self, row.fileID, row.domain, row.relativePath, row.flags, row.file)
 
 
 class FileInfo:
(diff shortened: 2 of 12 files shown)


dissect/target/plugins/os/unix/log/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+import re
+from datetime import datetime
+from typing import Iterator
+
+from dissect.target.helpers.fsutil import TargetPath, open_decompress
+
+log = logging.getLogger(__name__)
+
+_RE_TS = r"^[A-Za-z]{3}\s*[0-9]{1,2}\s[0-9]{1,2}:[0-9]{2}:[0-9]{2}"
```
Poeloe, line 11:
```suggestion
_RE_TS = r"^[A-Za-z]{3}\s*\d{1,2}\s\d{1,2}:\d{2}:\d{2}"
```

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -75,16 +76,26 @@ def messages(self) -> Iterator[MessagesRecord]:
                 yield from self._parse_cloud_init_log(log_file, tzinfo)
                 continue
 
-            for ts, line in year_rollover_helper(log_file, RE_TS, DEFAULT_TS_LOG_FORMAT, tzinfo):
-                daemon = dict(enumerate(RE_DAEMON.findall(line))).get(0)
-                pid = dict(enumerate(RE_PID.findall(line))).get(0)
-                message = dict(enumerate(RE_MSG.findall(line))).get(0, line)
+            if is_iso_fmt(log_file):
+                iterable = iso_readlines(log_file)
```
Poeloe, line 81:
```suggestion
```

dissect/target/plugins/os/unix/log/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+    with open_decompress(file, "rt") as fh:
+        for line in fh:
+            if not (match := RE_TS_ISO.match(line)):
+                log.warning("No timestamp found in one of the lines in %s!", file)
+                log.debug("Skipping line: %s", line)
+                continue
+
+            try:
+                ts = datetime.strptime(match[0], "%Y-%m-%dT%H:%M:%S.%f%z")
+
```
Poeloe, line 33:
```suggestion
```

dissect/target/plugins/os/unix/log/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+log = logging.getLogger(__name__)
+
+_RE_TS = r"^[A-Za-z]{3}\s*[0-9]{1,2}\s[0-9]{1,2}:[0-9]{2}:[0-9]{2}"
+_RE_TS_ISO = r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}\+\d{2}:\d{2}"
+
+RE_TS = re.compile(_RE_TS)
+RE_TS_ISO = re.compile(_RE_TS_ISO)
+RE_LINE = re.compile(
+    rf"(?P<ts>{_RE_TS}|{_RE_TS_ISO})\s(?P<hostname>\S+)\s(?P<service>\S+?)(\[(?P<pid>\d+)\])?:\s(?P<message>.+)$"
```
Poeloe, line 17: Some small changes to make the code below more compact:

- Start regex with matching on the similar ending of both timestamps
- Do match on the hostname, but do not capture it since it's not used
- Rename service to daemon, such that it matches the Record Descriptor
- Strip spaces left and right for the message
- Make the regex verbose to be able to provide comments

```suggestion
   RE_LINE = re.compile(
    r"""
    \d{2}:\d{2}\s                           # First match on the similar ending of the different timestamps
    (?:\S+)\s                               # The hostname, but do not capture it
    (?P<daemon>\S+?)(\[(?P<pid>\d+)\])?:    # The daemon with optionally the PID between brackets
    \s*(?P<message>.+?)\s*$                 # The log message stripped from spaces left and right
    """,
    re.VERBOSE,
)
```

and make sure you use `RE_LINE.search(line)` instead of `RE_LINE.match(line)` in `messages.py`

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -75,16 +76,26 @@ def messages(self) -> Iterator[MessagesRecord]:
...
 
+            else:
+                iterable = year_rollover_helper(log_file, RE_TS, DEFAULT_TS_LOG_FORMAT, tzinfo)
+
+            for ts, line in iterable:
+                match = RE_LINE.match(line)
```
Poeloe, line 86:
```suggestion
                match = RE_LINE.search(line)
```

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -75,16 +76,26 @@ def messages(self) -> Iterator[MessagesRecord]:
...
                     ts=ts,
-                    daemon=daemon,
-                    pid=pid,
-                    message=message,
+                    daemon=values["service"],
+                    pid=values["pid"],
+                    message=values["message"],
                     source=log_file,
                     _target=self.target,
                 )
```
Poeloe, line 101:
```suggestion
                    source=log_file,
                    **values,
                    _target=self.target,
                )
```

Poeloe (changes requested):
Since the ISO timestamp format support has been merged to main for the `AuthLogPlugin`, it might be good to use this new helper code there as well to prevent code duplication.

@Schamper because I don't have a particular opinion about the placement of this `helpers.py` file, do you agree on its location?

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -75,16 +80,23 @@ def messages(self) -> Iterator[MessagesRecord]:
...
 
-            for ts, line in year_rollover_helper(log_file, RE_TS, DEFAULT_TS_LOG_FORMAT, tzinfo):
-                daemon = dict(enumerate(RE_DAEMON.findall(line))).get(0)
-                pid = dict(enumerate(RE_PID.findall(line))).get(0)
-                message = dict(enumerate(RE_MSG.findall(line))).get(0, line)
+            if is_iso_fmt(log_file):
+                iterable = iso_readlines(log_file)
+
+            else:
+                iterable = year_rollover_helper(log_file, RE_TS, DEFAULT_TS_LOG_FORMAT, tzinfo)
```
Poeloe, line 87: Would it still work if you use `RE_TS` from `helpers.py` here, instead of the separately defined one in `messages.py`? I would namely prefer to make use of the regexes in the `helpers.py` file wherever possible.

dissect/target/plugins/os/unix/log/messages.py
```diff
@@ -75,16 +80,23 @@ def messages(self) -> Iterator[MessagesRecord]:
                 yield from self._parse_cloud_init_log(log_file, tzinfo)
                 continue
 
-            for ts, line in year_rollover_helper(log_file, RE_TS, DEFAULT_TS_LOG_FORMAT, tzinfo):
-                daemon = dict(enumerate(RE_DAEMON.findall(line))).get(0)
-                pid = dict(enumerate(RE_PID.findall(line))).get(0)
-                message = dict(enumerate(RE_MSG.findall(line))).get(0, line)
```
Poeloe, line 81: Removing these lines would also mean that the separately defined `RE_*` variables can be removed, right?

Poeloe (changes requested):
Could you also remove the `RE_TS_ISO` variable from the `auth.py` file and import the `RE_LINE` and `RE_TS` variable from the `helpers.py` file instead of defining it separately in the `auth.py` file?

Poeloe (approved):
Looks good!

PR #911, opened 2024-10-17
Fix circular imports during loader registration 

### LINKED ISSUE #910: Fix circular imports error during Loader registration
During `Loader` registration some loaders incorrectly fail to import due to `ImportError`'s caused by circular imports, such as the `RemoteLoader` (remote.py).

```python
\*\*\* ImportError: Failed to lazily import dissect.target.loaders.remote: cannot import name 'Target' from partially initialized module 'dissect.target.target' (most likely due to a circular import) (/projects/dissect/dissect.target/dissect/target/target.py)
  File "/usr/local/Cellar/python@3.9/3.9.15/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pdb.py", line 366, in displayhook
    self.message(repr(obj))
  File "/projects/dissect/dissect.target/dissect/target/helpers/lazy.py", line 66, in __repr__
    return f"<lazyattr {self.module._module_name}.{self.attr} loaded={self.realattr is not None}>"
  File "/projects/dissect/dissect.target/dissect/target/helpers/lazy.py", line 51, in realattr
    self._realattr = getattr(self.module._module, self.attr)
  File "/projects/dissect/dissect.target/dissect/target/helpers/lazy.py", line 38, in __getattr__
    self._error()
  File "/projects/dissect/dissect.target/dissect/target/helpers/lazy.py", line 32, in _error
    raise ImportError(f"Failed to lazily import {self.module._module_name}: {self.exc}")  # noqa
```

Horofic:
(DIS-3463)

Fixes #910

### DIFF
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -116,6 +116,7 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
         module = ".".join([MODULE_PATH, module_name])
     else:
         module = module_name
+
     loader = getattr(import_lazy(module), class_name)
     LOADERS.append(loader)
     LOADERS_BY_SCHEME[module_name] = loader
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -116,7 +116,6 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
         module = ".".join([MODULE_PATH, module_name])
     else:
         module = module_name
-
     loader = getattr(import_lazy(module), class_name)
     LOADERS.append(loader)
     LOADERS_BY_SCHEME[module_name] = loader
--- a/dissect/target/loaders/ab.py
+++ b/dissect/target/loaders/ab.py
@@ -1,10 +1,12 @@
+from __future__ import annotations
+
 import hashlib
 import io
 import posixpath
 import shutil
 import struct
 from pathlib import Path
-from typing import BinaryIO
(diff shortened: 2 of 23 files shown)


dissect/target/loader.py
```diff
@@ -116,7 +116,6 @@ def register(module_name: str, class_name: str, internal: bool = True) -> None:
         module = ".".join([MODULE_PATH, module_name])
     else:
         module = module_name
-
     loader = getattr(import_lazy(module), class_name)
```
Schamper, line 119:
```suggestion

    loader = getattr(import_lazy(module), class_name)
```

dissect/target/loaders/ab.py
```diff
@@ -21,7 +23,10 @@
 from dissect.target.helpers import keychain
 from dissect.target.loader import Loader
 from dissect.target.plugins.os.unix.linux.android._os import AndroidPlugin
-from dissect.target.target import Target
+
+if TYPE_CHECKING:
+    from dissect.target import Target
```
Schamper, line 28:
```suggestion
    from dissect.target.target import Target
```
I'd make it the full import everywhere.

Schamper (changes requested):
Can you diagnose the root cause of the circular import? Because this seems to have always worked fine, so the PR title is not necessarily correct. This is bypassing the circular import issue with moving all `Target` imports to type checking.

PR #915, opened 2024-10-23
Restructure editor plugin namespace

JSCU-CNI:
This PR restructures the `editor` plugin namespace. The previous implementation deviated from our internal structure for this category of plugins. This patch paves the road for open sourcing more editor plugins.


dissect/target/plugins/apps/editor/windowsnotepad.py
```diff
@@ -94,15 +91,27 @@
 };
 """
 
-WINDOWS_SAVED_TABS_EXTRA_FIELDS = [("datetime", "modification_time"), ("digest", "hashes"), ("path", "saved_path")]
+GENERIC_TAB_CONTENTS_RECORD_FIELDS = [
+    ("string", "editor"),
+    ("string", "content"),
+    ("path", "path"),
+    ("string", "deleted_content"),
+    ("path", "source"),
```
Horofic, line 99:
```suggestion
    ("string", "editor"),
    ("string", "content"),
    ("path", "path"),
    ("string", "deleted_content"),
    ("path", "source"),
```
Some of these seem to be duplicate from `COMMON_EDITOR_FIELDS`. Could you perhaps that record descriptor here?

dissect/target/plugins/apps/editor/editor.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @export
+    def extensions(self) -> None:
+        """Yields installed extensions."""
+        raise NotImplementedError
+
+    @export
+    def history(self) -> None:
+        """Yields history of files."""
+        raise NotImplementedError
```
Horofic, line 23: Any reason to include these here? This does not seem to consistent with most of the other 'base' namespaced plugins, such as with the remote access and locate base plugins.

Horofic (changes requested):

dissect/target/plugins/apps/editor/editor.py
Horofic: Check. Would you mind adding the alias? Then ill approve this PR

Horofic (approved):

PR #917, opened 2024-10-25
Add blksize, nblocks, and birthtime_ns to stat for FFS

### LINKED ISSUE #818: Add `st_blocks` and `st_blksize` to stat output for FFS
As discussed in https://github.com/fox-it/dissect.target/pull/812#discussion_r1716827893.

twiggler:
On Linux, stat always returns the default block size of 4096
We are returning the actual block size of the filesystem, similar to stat on  BSD

Note: st_blocks * 512 can be lower than st_blksize because FFS employs fragments

Closes #818

### DIFF
--- a/tests/filesystems/test_ffs.py
+++ b/tests/filesystems/test_ffs.py
@@ -1,0 +1,67 @@
+from datetime import datetime
+from typing import Iterator
+from unittest.mock import Mock, patch
+
+import pytest
+
+from dissect.target.filesystems.ffs import FfsFilesystem, FfsFilesystemEntry
+
+NANOSECONDS_IN_SECOND = 1_000_000_000
+
+
+@pytest.fixture
+def ffs_fs() -> Iterator[FfsFilesystem]:
+    with patch("dissect.ffs.ffs.FFS"):
+        ffs_fs = FfsFilesystem(Mock())
+        ffs_fs.ffs.block_size = 32 * 1024
+        yield ffs_fs
+
+
+@pytest.fixture
+def ffs_fs_entry(ffs_fs: FfsFilesystem) -> Iterator[FfsFilesystemEntry]:
+    atime = datetime(2024, 10, 1, 12, 0, 0)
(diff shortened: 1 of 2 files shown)


tests/filesystems/test_ffs.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pytest
+
+from dissect.target.filesystems.ffs import FfsFilesystem, FfsFilesystemEntry
+
+NANOSECONDS_IN_SECOND = 1_000_000_000
+
+
+@pytest.fixture
+def ffs_fs() -> Iterator[FfsFilesystem]:
```
Poeloe, line 13: I would place this fixture in the `conftest.py` file, rename it to `fs_ffs` and place it at the other `fs_*` fixtures there

tests/filesystems/test_ffs.py
```diff
@@ -0,0 +1,67 @@
...
+    with patch("dissect.ffs.ffs.FFS"):
+        ffs_fs = FfsFilesystem(Mock())
+        ffs_fs.ffs.block_size = 32 * 1024
+        yield ffs_fs
+
+
+@pytest.fixture
+def ffs_fs_entry(ffs_fs: FfsFilesystem) -> Iterator[FfsFilesystemEntry]:
```
Poeloe, line 21: Maybe just also place this fixture in the `conftest.py`, but I'll leave that up to you :smile:

Poeloe (changes requested):

tests/filesystems/test_ffs.py
Schamper: I wouldn't, it's only used here.

Schamper (approved):

tests/filesystems/test_ffs.py
twiggler: @Poeloe The `ffs_fs ` fixture is only used in this test file, so I agree with Schamper that it makes more sense to keep it here. I also kept the fixture in the test file for xfs and jffs2.

Poeloe (approved):

PR #921, opened 2024-10-30
Improve FortiManager OS support

JSCU-CNI:
This PR improves FortiManager OS support in dissect.


dissect/target/plugins/os/unix/_os.py
```diff
@@ -301,37 +317,36 @@ def _parse_os_release(self, glob: str | None = None) -> dict[str, str]:
...
+        bits = unpack("B", fh.read(1))[0]
+        fh.seek(18)
+        # ELF - e_machine
+        arch = unpack("H", fh.read(2))[0]
+        arch = ARCH_MAP.get(arch)
+
+        if bits == 1:  # 32 bit system
+            return f"{arch}_32-{os}"
+        else:
+            return f"{arch}-{os}"
```
Horofic, line 349: Would this work instead?
```suggestion
        if not isinstance(path, TargetPath):
            path = self.target.fs.path(path)
        
        if (path := self.target.fs.path(path)).exists():
            fh = path.open("rb")
            fh.seek(4)
            # ELF - e_ident[EI_CLASS]
            bits = unpack("B", fh.read(1))[0]
            fh.seek(18)
            # ELF - e_machine
            arch = unpack("H", fh.read(2))[0]
            arch = ARCH_MAP.get(arch)

            if bits == 1:  # 32 bit system
                return f"{arch}_32-{os}"
            else:
                return f"{arch}-{os}"
```

Horofic (changes requested):

dissect/target/plugins/os/unix/_os.py
Schamper: Keeping the loop in should be fine, but it can be written a bit nicer without the `CANDIDATE_FOUND` and more Pythonic with `None` checks. Maybe introduce a new variable `pathobj` (or swapping it around and introduce a `pathstr`, set `path` to `None` and check for a `if path is None or not path.exists()`)?

The loop should also probably start with `self.target.fs` as the first filesystem to try and check, followed by the others. I'm not sure if our "custom list" at `.filesystems` supports the `[self.target.fs, *self.target.filesystems]` syntax though. The check in the loop can also be simplified with a `fs.path(path).exists()`.

dissect/target/plugins/os/unix/_os.py
```diff
@@ -301,37 +320,30 @@ def _parse_os_release(self, glob: str | None = None) -> dict[str, str]:
...
+        Resources:
+            - https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#ISA
+        """
+
+        if not isinstance(path, TargetPath):
+            for fs in [self.target.fs, *self.target.filesystems]:
+                if (path := fs.path(path)).exists():
+                    break
+
+        if path is None or not path.exists():
```
Schamper, line 335:
```suggestion
        if not path.exists():
```
I think path will always be a Path object with the current code.

Schamper (changes requested):

Schamper (dismissed):

dissect/target/plugins/os/unix/_os.py
```diff
@@ -301,37 +319,30 @@ def _parse_os_release(self, glob: str | None = None) -> dict[str, str]:
...
+
+        fh = path.open("rb")
+        fh.seek(4)  # ELF - e_ident[EI_CLASS]
+        bits = fh.read(1)[0]
+
+        fh.seek(18)  # ELF - e_machine
+        e_machine = int.from_bytes(fh.read(2), "little")
+        arch = ARCH_MAP.get(e_machine, "unknown")
+
+        return f"{arch}_32-{os}" if bits == 1 and not arch[-2:] == "32" else f"{arch}-{os}"
```
Horofic, line 345:
```suggestion
        if not isinstance(path, TargetPath):
            path = self.target.fs.path(path)

        for fs in [self.target.fs, *self.target.filesystems]:
            if (path := fs.path(path)).exists():
                fh = path.open("rb")
                fh.seek(4)  # ELF - e_ident[EI_CLASS]
                bits = fh.read(1)[0]

                fh.seek(18)  # ELF - e_machine
                e_machine = int.from_bytes(fh.read(2), "little")
                arch = ARCH_MAP.get(e_machine, "unknown")

                return f"{arch}_32-{os}" if bits == 1 and not arch[-2:] == "32" else f"{arch}-{os}"
```

Horofic: This should do it. Also, thanks for adding the test!

Horofic (changes requested):

dissect/target/plugins/os/unix/_os.py
Schamper: I prefer the original code, your suggestion is not really how `Path` objects are supposed to be used.

Horofic: Made a small edit in the suggestion. I dont see why doing the `.exists()` check twice is desirable. 
Only doing the path exists check in the loop seems fine to me and avoids the above. There can be instances where `Path` objects are passed to the check in the loop now, but that seems negligible.

Horofic: Fair point! I did not consider this 😄. After this 101 the original solution makes more sense. So consider this resolved.

dissect/target/plugins/os/windows/_os.py
```diff
@@ -12,6 +12,14 @@
 from dissect.target.plugin import OperatingSystem, OSPlugin, export
 from dissect.target.target import Target
 
+ARCH_MAP = {
+    "x86": 32,
+    "IA64": 64,
+    "ARM64": 64,
+    "EM64T": 64,
+    "AMD64": 64,
+}
```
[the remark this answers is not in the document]
Schamper: Would you mind creating a ticket for that?

Schamper (approved):

Horofic (approved):

PR #935, opened 2024-11-05
Add ZIP loader

### LINKED ISSUE #934: Cannot process ZIP based acquire collects
Currently there is no loader for ZIP based Acquire collects. So if you run acquire like `acquire -ot zip` and follow this by `target-query -f mft <filename>` this won't work.

Matthijsy:
This PR adds a loader for ZIP files, which is mainly useful for ZIP based acquire collects. 

I don't have experience with creating loaders, and mainly took the logic from the TarLoader. So if there are things that can be done better would be happy to improve!

fixes #934


Horofic:
First of all, thanks for you contribution @Matthijsy. I think this is a good first attempt :)! 

As you kind of pointed out, the `ZIPloader` mainly took the logic from the `TarLoader` with some minor tweaks. I think a better way forward would be to create an `AcquireLoader` that works on both `Zip`, `Tar` and `Dir` based Acquire collections! Kind of like the `VelociraptorLoader` found [here](https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/velociraptor.py). Here it does some specific checks to see if a collection is 
is a Velociraptor collection.

Let me know what you think about this!

dissect/target/loaders/zip.py
```diff
@@ -0,0 +1,97 @@
+import logging
```
Horofic, line 1: Could you add tests for this new functionality? This could be done later, if you decide to go on the `AcquireLoader` route.

Horofic (changes requested):

Matthijsy:
@Horofic Thank you for the feedback, that indeed sounds like a good plan. I have now changed it to a dedicated `AcquireLoader`. This way it seems to work for both the `zip` and `directory` types. However, the directory mapping for the `tar` version does not work yet. When I test this using `target-shell` the root directories look like this

```
c:
fs0
fs1
```

I suspect that the `find_and_map_dirs` does not handle the tar format correctly, but I cannot figure out how to add support for that. Could you give me some pointers in how to do that?

Furthermore, I suspect this will not work anymore for the legacy acquire format as described in the `TarLoader` logic. Is this still something that should be supported?

Matthijsy:
I have also added (or mainly re-used) a few test cases now. Only the `test_tar_sensitive_drive_letter` and `test_tar_anonymous_filesystems` don't work anymore. I don't get the cases they try to test, is that something related to Acquire of for the TarLoader?

dissect/target/loaders/acquire.py
```diff
@@ -0,0 +1,56 @@
...
+        self.root = _get_root(path)
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        root = _get_root(path)
+
+        if not root:
+            return False
+
+        return root.joinpath(FILESYSTEMS_ROOT).exists() or root.joinpath(FILESYSTEMS_LEGACY_ROOT).exists()
```
Schamper, line 45: This will be very slow on large tar files, where there will be a double performance penalty. First the entire tar file will be parsed just for this check, then if it exists, it will be parsed again in `__init__`. However, if it doesn't exist, it will still be parsed _again_ in the actual tar loader.

tests/loaders/test_acquire.py
```diff
@@ -0,0 +1,70 @@
...
+#     loader = AcquireLoader(Path(tar_file))
+#     loader.map(target_default)
+#
+#     # mounts = $fs$/fs0, $fs$/fs1 and /
+#     assert len(target_default.fs.mounts) == 3
+#     assert "$fs$/fs0" in target_default.fs.mounts.keys()
+#     assert "$fs$/fs1" in target_default.fs.mounts.keys()
+#     assert "/" in target_default.fs.mounts.keys()
+#     assert target_default.fs.get("$fs$/fs0/foo").open().read() == b"hello world\n"
+#     assert target_default.fs.get("$fs$/fs1/bar").open().read() == b"hello world\n"
```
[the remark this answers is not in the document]
Horofic: I did some testing and it seems we are not doing everything correctly at the moment here. When running `target-shell` on a Windows Acquire collection using the `main` branch a file listing looks of `$fs/` as follows:
```bash
fs0
fs1
fs2
```
When doing the same on this branch it looks like this:
```bash
fs0
```
So we end up missing some for some volumes here. This is the case for `.tar`, `.zip` and `dir` Acquire collection.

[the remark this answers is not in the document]
Matthijsy: I am trying to fix that, but it seems that on master `$fs$/fs0` gets mapped as `/fs1` instead of `/fs0`. Is that on purpose, or is that a bug?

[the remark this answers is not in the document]
Matthijsy: I have added support for it, they test is changed a little bit as we don't create dedicated mounts for everything anymore but via the LayeredFileSystem, so the files are accessible in the same way. You can now still read the files via `$fs$/fs0/foo` and via `fs1/foo`

tests/loaders/test_acquire.py
```diff
@@ -0,0 +1,70 @@
...
+#     # Initialize our own WindowsPlugin to override the detection
+#     target_bare._os_plugin = WindowsPlugin.create(target_bare, target_bare.fs.mounts["c:"])
+#     target_bare._init_os()
+#
+#     # sysvol is now added
+#     assert sorted(target_bare.fs.mounts.keys()) == ["/", "c:", "sysvol"]
+#
+#     # WindowsPlugin sets the case sensitivity to False
+#     assert target_bare.fs.get("C:/test.file").open().read() == b"hello_world"
+#     assert target_bare.fs.get("c:/test.file").open().read() == b"hello_world"
```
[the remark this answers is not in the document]
Matthijsy: Thank you, this should be supported as well now

dissect/target/loaders/acquire.py
Matthijsy: Yes I agree, but is there a way to prevent that? As we do need to check if the `fs` directory exists in the file, since otherwise the tar is not an acquire collect and thus need to be handled by the `TarLoader`.

Horofic (changes requested):

dissect/target/loaders/acquire.py
Schamper: "Ideally" (not really, since it's only for performance and the code path itself will be more confusing) it's a subpath in the `zip` and `tar` loaders. So you piggy back on the detection logic of those, and upon mapping you do a fast check to see if you need to divert logic to Acquire logic (which could exist in `loaders/acquire.py`.

At least, that's the "best" idea I can come up with right now.

dissect/target/loaders/dir.py
```diff
@@ -130,12 +139,21 @@ def find_dirs(path: Path) -> tuple[str, list[Path]]:
         for p in path.iterdir():
             # Look for directories like C or C:
             if p.is_dir() and (is_drive_letter_path(p) or p.name in ("sysvol", "$rootfs$")):
-                dirs.append(p)
+                if p.name == "sysvol":
+                    dirs.append(('c', p))
+                else:
+                    dirs.append((p.name[0], p))
```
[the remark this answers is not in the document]
Matthijsy: I agree, based on the tests I found that the `kape` and `tanium` loader broke by this change, and I have fixed them in the latest commit. Can we trust on the unittests for this, or should we do additional checks somehow?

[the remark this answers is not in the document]
Horofic: This should be fine. Could you change the function's return type and docstring to reflect the change you made.

dissect/target/loaders/dir.py
```diff
@@ -130,12 +140,21 @@ def find_dirs(path: Path) -> tuple[str, list[Path]]:
...
 
                 if not os_type:
                     os_type = os_type_from_path(p)
 
-        if not os_type:
+            if p.name == "$fs$":
+                dirs.append(('$fs$', p))
```
Horofic, line 152:
```suggestion
                dirs.append(("$fs$", p))
```

dissect/target/loaders/dir.py
```diff
@@ -130,12 +140,21 @@ def find_dirs(path: Path) -> tuple[str, list[Path]]:
         for p in path.iterdir():
             # Look for directories like C or C:
             if p.is_dir() and (is_drive_letter_path(p) or p.name in ("sysvol", "$rootfs$")):
-                dirs.append(p)
+                if p.name == "sysvol":
+                    dirs.append(('c', p))
```
Horofic, line 144:
```suggestion
                    dirs.append(("c", p))
```

Horofic (changes requested):

dissect/target/loaders/acquire.py
Schamper: Can you base it on https://github.com/fox-it/dissect.target/pull/1000?

dissect/target/loaders/acquire.py
```diff
@@ -0,0 +1,53 @@
...
+    def detect(tarfile: tf.TarFile) -> bool:
+        for member in tarfile.getmembers():
+            if member.name.startswith((f"/{FILESYSTEMS_ROOT}/", f"{FILESYSTEMS_ROOT}/",
+                                           f"/{FILESYSTEMS_LEGACY_ROOT}/", f"{FILESYSTEMS_LEGACY_ROOT}/")):
+                return True
+        return False
+
+    def map(self, target: Target) -> None:
+        tar_fs = TarFilesystem(tarfile=self.tar).path()
+        map_acquire(target, tar_fs)
```
Schamper, line 33: This combined with the change in `dissect/target/loaders/dir.py` will still iterate the tarfile twice. Why not just move the logic from `dissect/target/loaders/tar.py` in here?

It may be a good idea to add a small benchmark test for loading a tar file in a separate PR before merging this one. You can take a look at https://github.com/fox-it/dissect.target/pull/1129 for inspiration.

Schamper (changes requested):

dissect/target/loaders/acquire.py
Matthijsy: I just tested it a bit with the benchmark, and indeed the old implementation is quicker. I have moved the logic from the tar loader into the acquire loader to make sure it stays efficient. @Schamper do you agree with this approach?

Schamper: Yes, efficiency is more important in this case than saving a few lines of code.

dissect/target/loaders/dir.py
Schamper, line 1: Are all the changes in this file still necessary with the revised approach?

dissect/target/loaders/kape.py
Schamper, line 1: Same here, as a result of the `dir.py` comment/changes.

dissect/target/loaders/tanium.py
Schamper, line 1: Same here, as a result of the `dir.py` comment/changes.

dissect/target/loaders/zip.py
```diff
@@ -0,0 +1,73 @@
...
+        """Only to be called internally by :class:`ZipLoader`."""
+        raise NotImplementedError
+
+    def map(self, target: target.Target) -> None:
+        """Only to be called internally by :class:`ZipLoader`."""
+        raise NotImplementedError
+
+class GenericZipSubLoader(ZipSubLoader):
+    """Generic zip sub loader.
+    """
```
Schamper, line 36:
```suggestion
    """Generic zip sub loader."""
```

tests/loaders/test_acquire.py
```diff
@@ -0,0 +1,94 @@
...
+    [
+        ("_data/loaders/acquire/test-windows-fs-c-relative.tar", TarLoader),
+        ("_data/loaders/acquire/test-windows-fs-c.zip", ZipLoader),
+    ],
+)
+@pytest.mark.benchmark
+def test_benchmark(benchmark: BenchmarkFixture, target_default: Target, archive: str, loader: Loader) -> None:
+    file = Path(absolute_path(archive))
+
+    benchmark(lambda: loader(file).map(target_default))
```
Schamper, line 94: Could you make this small change a separate PR? That way we can see the difference in performance. I'll approve that one super quick 😉.

Schamper (changes requested):
Also the linter is giving some errors.

PR #870, opened 2024-10-01
Fix inconsistency of target.ips output

JSCU-CNI:
Fixes #877.

PR #936, opened 2024-11-06
Fix a regression bug inside windows network_manager

Miauwkeru:
Found a regression inside #870 in the case a registry key doesn't exist or is empty.
I made a small fix for it here, however I also noticed the types aren't correct either. Will save that for another moment

### DIFF
--- a/dissect/target/filesystems/fat.py
+++ b/dissect/target/filesystems/fat.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 import stat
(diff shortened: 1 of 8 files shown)


twiggler (approved):
The fix looks ok. 
Since it is nice to get this in the next release I approve it.

Can you create a ticket to:
- add/adjust the  unit test for when `extract_network_device_config` returns nothing
- Make the return type of  _extract_network_device_config more meaningful, for example by adding data classes

Issue #790, opened 2024-08-06
Add handling for non-ascii characters PR#270 


EinatFox:
This issue is to follow up on PR #270

PR #938, opened 2024-11-07
Make WER file field names with non-ascii characters usable

pyrco:
We use WER field names as the field names for the records the WER plugin creates. This means they have certain limitations in what characters can be used. Previously all characters in the set [a-zA-Z_] were removed. This patch encodes them as _<utf-8_sequence_in_hex>.


pyrco:
This improves on https://github.com/fox-it/dissect.target/pull/544

It also makes https://github.com/fox-it/dissect.target/pull/270 redundant as it accomplishes the same thing in a slightly different way.

When this ticket is merged the issue https://github.com/fox-it/dissect.target/issues/790 can be closed.

pyrco:
This PR uncovered a bug in flow.record's field name validation which is fixed in https://github.com/fox-it/flow.record/pull/151

cecinestpasunepipe (approved):

PR #941, opened 2024-11-12
Add `st_blocks` and `st_blksize` to stat output for FAT

cecinestpasunepipe:
DIS-3368

### DIFF
--- a/dissect/target/filesystems/fat.py
+++ b/dissect/target/filesystems/fat.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 import stat
 from typing import BinaryIO, Iterator, Optional, Union
 
@@ -116,7 +117,7 @@ class FatFilesystemEntry(FilesystemEntry):
             ]
         )
 
-        st_info.st_blocks = self.entry.nblocks
-        st_info.st_blksize = self.entry.blksize
+        st_info.st_blocks = math.ceil( self.entry.size / self.entry.fs.cluster_size)
+        st_info.st_blksize = self.entry.fs.cluster_size
         st_info.st_birthtime = creation_time
         return st_info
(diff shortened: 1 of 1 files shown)


dissect/target/filesystems/fat.py
```diff
@@ -100,16 +101,23 @@ def stat(self, follow_symlinks: bool = True) -> fsutil.stat_result:
...
+                self.entry.size,
+                self.entry.atime.replace(tzinfo=self.fs.tzinfo).timestamp(),
+                self.entry.mtime.replace(tzinfo=self.fs.tzinfo).timestamp(),
+                creation_time,
+            ]
+        )
+
+        st_info.st_blocks = math.ceil( self.entry.size / self.entry.fs.cluster_size)
+        st_info.st_blksize = self.entry.fs.cluster_size
+        st_info.st_birthtime = creation_time
```
Schamper, line 122:
```suggestion
        st_info = fsutil.stat_result(
            [
                (stat.S_IFDIR if self.is_dir() else stat.S_IFREG) | 0o777,
                self.entry.cluster,
                id(self.fs),
                1,
                0,
                0,
                self.entry.size,
                self.entry.atime.replace(tzinfo=self.fs.tzinfo).timestamp(),
                self.entry.mtime.replace(tzinfo=self.fs.tzinfo).timestamp(),
                self.entry.ctime.replace(tzinfo=self.fs.tzinfo).timestamp(),
            ]
        )

        st_info.st_blocks = math.ceil( self.entry.size / self.entry.fs.cluster_size)
        st_info.st_blksize = self.entry.fs.cluster_size
```
FAT does not have an explicit birthtime. That field should only be filled if the filesystem has an explicit timestamp for it.

Schamper (changes requested):

Schamper (approved):

PR #949, opened 2024-11-18
Add a dependency on dissect.fve for the full extra

pyrco:


Zawadidone:
@pyrco does this `platform_python_implementation != 'PyPy'` mean that `dissect.fve` does not work with PyPy, as part of `dissect.target`?

Zawadidone:
Okay thanks for the explanation.

PR #952, opened 2024-11-19
Strip trailing dir separator for translated file paths in ITunesLoader

### LINKED ISSUE #947: iTunes DOMAIN_TRANSLATION path not visible in `target-shell`
When mounting a iTunes backup with `target-shell` the DOMAIN_TRANSLATION paths do not show up when using `ls`, `ls -R`, or `tree`:

image

I did notice when I add the following code to the [translate_file_path](https://github.com/fox-it/dissect.target/blob/1a2c8feaac89ccda33847cc119fced392b41ab40/dissect/target/loaders/itunes.py#L283) function:

```
    if not relative_path:
        relative_path = Path(domain_path).name
```

It is a bit better, but now the current path also always shows up as a child directory:
image

I'm fine with making a PR for a fix but I'm not 100% sure where it should be implemented.

Poeloe:
Fixes #947


tests/loaders/test_itunes.py
```diff
@@ -0,0 +1,25 @@
...
+        pytest.param("ProtectedDomain", "", "ProtectedDomain", id="ProtectedDomain"),
+        pytest.param(
+            "AppDomain-com.apple.freeform",
+            "",
+            "/private/var/mobile/Containers/Data/Application/com.apple.freeform",
+            id="AppDomain",
+        ),
+    ],
+)
+def test_translate_file_path(domain: str, relative_path: str, result):
```
Schamper, line 24:
```suggestion
def test_translate_file_path(domain: str, relative_path: str, result) -> None:
```

Schamper (changes requested):

tests/loaders/test_itunes.py
```diff
@@ -0,0 +1,25 @@
...
+        pytest.param("ProtectedDomain", "", "ProtectedDomain", id="ProtectedDomain"),
+        pytest.param(
+            "AppDomain-com.apple.freeform",
+            "",
+            "/private/var/mobile/Containers/Data/Application/com.apple.freeform",
+            id="AppDomain",
+        ),
+    ],
+)
+def test_translate_file_path(domain: str, relative_path: str, result) -> None:
```
Schamper, line 24:
```suggestion
def test_translate_file_path(domain: str, relative_path: str, result: str) -> None:
```

Schamper (dismissed):

Issue #951, opened 2024-11-19
HKLM\\SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces\\{ID}\\DhcpNameServer may contain multiple ips, separated by a space


wbi-ocd:
DhcpNameServer may contain a list of server, separated by a space, causing the following error (crash) when the ips plugins is used : 

```bash
ValueError: '172.16.0.3 172.16.0.4 does not appear to be an IPv4 or IPv6 address
```

Debug trace shows than these two IPS are considered as one element and not two element of the set 

```
│ in __init__:38                                                                                   │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │                  _classification = None                                                      │ │
│ │                       _generated = None                                                      │ │
│ │                          _source = PosixPath('CENSORED')       │ │
│ │                         _version = None                                                      │ │
│ │ _windows_network_interface__self = <repr-error "'windows_network_interface' object has no    │ │
│ │                                    attribute 'dns'">                                         │ │
│ │                             dhcp = True                                                      │ │
│ │                              dns = {'172.16.0.3 172.16.0.4'}   
``` 

A simple fix would be to split on space when the value contains space  :
https://github.com/fox-it/dissect.target/blob/d6eb12e18f6650c4d4f279620032b8ce463b07df/dissect/target/plugins/os/windows/network.py#L357

wbi-ocd:
I have made some test and the following line key may also cause bugs : 
* [            static_config["dns"].update(_get_config_value(key, "NameServer"))](https://github.com/fox-it/dissect.target/blob/d6eb12e18f6650c4d4f279620032b8ce463b07df/dissect/target/plugins/os/windows/network.py#L361C13-L361C78) :> could contains comma separated values (eg after running the following commands : `Set-DnsClientServerAddress -InterfaceIndex 13  -ServerAddresses('8.8.8.8','4.4.4.4')` the NameServer key will contains : `8.8.8.8,4.4.4.4`

This look like a regression related to https://github.com/fox-it/dissect.target/commit/90cbaac62b1f46461cbb270b2c9e9820097911ef#diff-a8cdba395abca9f908bbf562e9a7f94c37e2bed16994171b11e08e45e17333eaL311 , as a split was done before.

I will make a PR

Horofic:
This issue can indeed be closed. Ill raise comment about the release with the team!

PR #955, opened 2024-11-20
Fix error in windows network plugins

wbi-ocd:
Fix issue #951

### DIFF
--- a/dissect/target/plugins/os/windows/network.py
+++ b/dissect/target/plugins/os/windows/network.py
@@ -225,13 +225,13 @@ def _try_value(subkey: RegistryKey, value: str) -> str | list | None:
         return None
 
 
-def _get_config_value(key: RegistryKey, name: str, split_char: list[str] | None = None) -> set:
+def _get_config_value(key: RegistryKey, name: str, sep: str | None = None) -> set:
     value = _try_value(key, name)
     if not value or value in ("", "0.0.0.0", None, [], ["0.0.0.0"]):
         return set()
-    if split_char and isinstance(value, str):
-        split_regexp = '|'.join(re.escape(c) for c in split_char)
-        value = re.split(split_regexp, value)
+    if sep and isinstance(value, str):
+        re_sep = '|'.join(map(re.escape, sep))
+        value = re.split(re_sep, value)
     if isinstance(value, list):
         return set(value)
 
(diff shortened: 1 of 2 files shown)


wbi-ocd:
Patch has been applied

dissect/target/plugins/os/windows/network.py
```diff
@@ -224,11 +225,13 @@ def _try_value(subkey: RegistryKey, value: str) -> str | list | None:
         return None
 
 
-def _get_config_value(key: RegistryKey, name: str) -> set:
+def _get_config_value(key: RegistryKey, name: str, split_char: list[str] | None = None) -> set:
```
Schamper, line 228:
```suggestion
def _get_config_value(key: RegistryKey, name: str, sep: str | None = None) -> set:
```
Rename to something that's also used in Python standard library, plus we can just make this a string since you're treating it like a list of single character strings anyway. This is also in line how Pythons own "split" is used.

dissect/target/plugins/os/windows/network.py
```diff
@@ -224,11 +225,13 @@ def _try_value(subkey: RegistryKey, value: str) -> str | list | None:
...
     value = _try_value(key, name)
     if not value or value in ("", "0.0.0.0", None, [], ["0.0.0.0"]):
         return set()
-
+    if split_char and isinstance(value, str):
+        split_regexp = '|'.join(re.escape(c) for c in split_char)
+        value = re.split(split_regexp, value)
```
Schamper, line 234:
```suggestion
    if sep and isinstance(value, str):
        re_sep = '|'.join(re.escape(c) for c in sep)
        value = re.split(re_sep, value)
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -354,11 +357,11 @@ def _extract_network_device_config(self, interface_id: str) -> list[dict[str, se
             dhcp_config["ip"].update(_get_config_value(key, "DhcpIPAddress"))
             dhcp_config["subnetmask"].update(_get_config_value(key, "DhcpSubnetMask"))
             dhcp_config["search_domain"].update(_get_config_value(key, "DhcpDomain"))
-            dhcp_config["dns"].update(_get_config_value(key, "DhcpNameServer"))
+            dhcp_config["dns"].update(_get_config_value(key, "DhcpNameServer", [" ", ","]))
```
Schamper, line 360:
```suggestion
            dhcp_config["dns"].update(_get_config_value(key, "DhcpNameServer", " ,"))
```

dissect/target/plugins/os/windows/network.py
```diff
@@ -354,11 +357,11 @@ def _extract_network_device_config(self, interface_id: str) -> list[dict[str, se
...
 
             # Extract static configuration from the registry
             static_config["gateway"].update(_get_config_value(key, "DefaultGateway"))
-            static_config["dns"].update(_get_config_value(key, "NameServer"))
+            static_config["dns"].update(_get_config_value(key, "NameServer", [" ", ","]))
```
Schamper, line 364:
```suggestion
            static_config["dns"].update(_get_config_value(key, "NameServer", " ,"))
```

Schamper (changes requested):

wbi-ocd:
Changes applied

Schamper (approved):

wbi-ocd:
Done

PR #956, opened 2024-11-21
Allow RegfValue.type to fall back to int

JSCU-CNI:
This PR fixes edge cases where `RegfValue.type` cannot be translated to the `RegistryValueType` enum.

### DIFF
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -12,7 +12,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import BinaryIO, Iterator, Optional, TextIO, Union
 
-from dissect.regf import regf
+from dissect.regf import c_regf, regf
 
 from dissect.target.exceptions import (
     RegistryError,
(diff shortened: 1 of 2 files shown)


Horofic:
Would you mind adding a test case for this behaviour in `test_regutil`?

Not sure, but should this not return `REG_NONE` in this case? Took a cursory peek here https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types.

Schamper:
Using `_missing_` is probably nicer: https://docs.python.org/3/library/enum.html#enum.Enum._missing_

dissect/target/helpers/regutil.py
```diff
@@ -707,7 +724,7 @@ def value(self) -> ValueType:
         return self.kv.value
 
     @property
-    def type(self) -> RegistryValueType:
+    def type(self) -> RegistryValueType | int:
```
Schamper, line 727:
```suggestion
    def type(self) -> RegistryValueType:
```
No longer necessary.

Schamper (changes requested):

Schamper:
> Added in [308fda1](https://github.com/fox-it/dissect.target/pull/956/commits/308fda1cb84d9ac2dd63087bb8af0032664740f8). Perhaps the Enum should be moved to the `dissect.regf` project?

Maybe, I'm fine with it in here for now since this enum serves more as an abstraction layer than a "parsing" enum.

Schamper (approved):

Schamper (approved):

PR #965, opened 2024-12-04
Add Unix DHCP lease file parser for `target.ips`

### LINKED ISSUE #960: Get DHCP lease ips from /var/lib On Linux based operating systems
On various Linux based network managers, the DHCP leases are stored on various files.

For example, in NetworkManager, The DHCP lease ip are stored in `/var/lib/NetworkManager/*.lease`.
In dhclient, these are stored in `/var/lib/dhcp/dhclient.leases`.

Those are sample of a better solution to finding Target Ip, instead of parsing syslog, which in some cases could be a more resource intensive task.

JSCU-CNI:
This PR open sources a function called `parse_unix_dhcp_leases` which aids in finding DHCP IP addresses on Linux targets (`target.ips`). It does not implement DHCP parsing for the new `linux.network` plugin and thus partially fixes #960.


tests/plugins/os/unix/test_ips.py
```diff
@@ -246,3 +246,58 @@ def test_regression_ips_unique_strings(target_unix: Target, fs_unix: VirtualFile
...
+    fs_unix.map_file_fh("/var/lib/dhcp/dhclient.leases", BytesIO(textwrap.dedent(lease2).encode()))
+    fs_unix.map_file_fh("/var/lib/dhclient/dhclient.eth0.leases", BytesIO(textwrap.dedent(lease3).encode()))
+
+    syslog = "Apr  4 13:37:04 localhost dhclient[4]: bound to 1.3.3.7 -- renewal in 1337 seconds."
+    fs_unix.map_file_fh("/var/log/syslog", BytesIO(textwrap.dedent(syslog).encode()))
+
+    target_unix.add_plugin(LinuxPlugin)
+
+    # tests if we did not call :func:`parse_unix_dhcp_log_messages` since :func:`parse_unix_dhcp_leases` has results.
+    assert len(target_unix.ips) == 4
```
Horofic, line 302: This assertions currently fails because `target_unix.ips` return 3 entries instead of 4

Horofic (changes requested):

Horofic (approved):

PR #547, opened 2024-02-20
Move lru_cache definitions to __init__

Miauwkeru:
Using the lru_cache decorators on class methods, the ones that have a reference to `self`,
will also cache self. So we move it to the __init__ of the class
    
(DIS-2913)

PR #966, opened 2024-12-05
Microsoft Office add-in detection 

twiggler:
Detects:
- Persistence through startup folders (templates, executables)
- COM addins 
- VSTO addins by parsing ClickOnce Deployment manifests
- Web addins by inspecting WEF cache and parsing manifests

Limitations:
- Office for Mac is not supported
- Supports Microsoft Office 2016 and later
- Manifests which are wof compressed are unreadable

Bonus:
- Fix issue in resolver where user environment was not taken into account when expanding paths
- Fix issue where partial path resolution resolved to directories.
- Added convenience functions for querying registry

### DIFF
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -179,7 +179,7 @@ class RegistryKey:
         except RegistryValueNotFoundError:
             if default is self.__marker:
                 raise
-            return VirtualValue(VirtualHive(), value, default)
+            return VirtualValue(self.hive, value, default)
 
     def _value(self, value: str) -> RegistryValue:
         raise NotImplementedError()
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -161,28 +161,27 @@ class RegistryKey:
         """Returns a list of subkeys from this key."""
         raise NotImplementedError()
 
-    def value(self, value: str) -> RegistryValue:
+    __marker = object()
+
+    def value(self, value: str, default: ValueType = __marker) -> RegistryValue:
         """Returns a specific value from this key.
 
         Args:
             value: The name of the value to retrieve.
+            default: An optional argument that specifies the default return value in case the value cannot be found.
 
         Raises:
-            RegistryValueNotFoundError: If this key has no value with the requested name.
-        """
-        raise NotImplementedError()
-
-    def value_or_default(self, value: str, default: ValueType = None) -> RegistryValue:
-        """Returns a specific value from this key, or a default value if it does not exist.
-
-        Args:
-            value: The name of the value to retrieve.
-            default: The default value to return if the value does not exist.
+            RegistryValueNotFoundError: If this key has no value with the requested name and default is not defined.
         """
         try:
-            return self.value(value)
+            return self._value(value)
         except RegistryValueNotFoundError:
+            if default is self.__marker:
+                raise
             return VirtualValue(VirtualHive(), value, default)
+
+    def _value(self, value: str) -> RegistryValue:
+        raise NotImplementedError()
 
     def values(self) -> list[RegistryValue]:
         """Returns a list of all the values from this key."""
@@ -412,7 +411,7 @@ class VirtualKey(RegistryKey):
 
         return res.values()
 
-    def value(self, value: str) -> RegistryValue:
+    def _value(self, value: str) -> RegistryValue:
         try:
             return self._values[value.lower()]
         except KeyError:
@@ -600,7 +599,7 @@ class KeyCollection(RegistryKey):
 
         return ret.values()
 
-    def value(self, value: str) -> ValueCollection:
+    def _value(self, value: str) -> ValueCollection:
         ret = ValueCollection()
         for key in self:
             try:
@@ -710,7 +709,7 @@ class RegfKey(RegistryKey):
     def subkeys(self) -> list[RegistryKey]:
         return [RegfKey(self.hive, k) for k in self.key.subkeys()]
 
-    def value(self, value: str) -> RegistryValue:
+    def _value(self, value: str) -> RegistryValue:
         try:
             return RegfValue(self.hive, self.key.value(value))
         except regf.RegistryValueNotFoundError as e:
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -26,7 +26,7 @@ GLOB_MAGIC_REGEX = re.compile(r"[*?[]")
 KeyType = Union[regf.IndexLeaf, regf.FastLeaf, regf.HashLeaf, regf.IndexRoot, regf.NamedKey]
 """The possible key types that can be returned from the registry."""
 
-ValueType = Union[int, str, bytes, list[str]]
+ValueType = Union[int, str, bytes, list[str], None]
 """The possible value types that can be returned from the registry."""
 
 
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -26,7 +26,7 @@ GLOB_MAGIC_REGEX = re.compile(r"[*?[]")
 KeyType = Union[regf.IndexLeaf, regf.FastLeaf, regf.HashLeaf, regf.IndexRoot, regf.NamedKey]
 """The possible key types that can be returned from the registry."""
 
-ValueType = Union[int, str, bytes, list[str], None]
+ValueType = Union[int, str, bytes, list[str]]
 """The possible value types that can be returned from the registry."""
 
 
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -10,7 +10,7 @@ from enum import IntEnum
 from functools import cached_property
 from io import BytesIO
 from pathlib import Path
-from typing import BinaryIO, Iterator, Optional, TextIO, Union
+from typing import BinaryIO, Iterator, NewType, Optional, TextIO, Union
 
 from dissect.regf import c_regf, regf
 
@@ -161,9 +161,10 @@ class RegistryKey:
         """Returns a list of subkeys from this key."""
         raise NotImplementedError()
 
-    __marker = object()
+    Marker = NewType("Marker", object)
+    __marker = Marker(object())
 
-    def value(self, value: str, default: ValueType = __marker) -> RegistryValue:
+    def value(self, value: str, default: ValueType | Marker = __marker) -> RegistryValue:
         """Returns a specific value from this key.
 
         Args:
--- a/dissect/target/helpers/regutil.py
+++ b/dissect/target/helpers/regutil.py
@@ -26,7 +26,7 @@ GLOB_MAGIC_REGEX = re.compile(r"[*?[]")
 KeyType = Union[regf.IndexLeaf, regf.FastLeaf, regf.HashLeaf, regf.IndexRoot, regf.NamedKey]
 """The possible key types that can be returned from the registry."""
 
-ValueType = Union[int, str, bytes, list[str]]
+ValueType = Union[int, str, bytes, list[str], None]
 """The possible value types that can be returned from the registry."""
 
 
@@ -171,6 +171,18 @@ class RegistryKey:
             RegistryValueNotFoundError: If this key has no value with the requested name.
         """
         raise NotImplementedError()
+
+    def value_or_default(self, value: str, default: ValueType = None) -> RegistryValue:
+        """Returns a specific value from this key, or a default value if it does not exist.
+
+        Args:
+            value: The name of the value to retrieve.
+            default: The default value to return if the value does not exist.
+        """
+        try:
+            return self.value(value)
+        except RegistryValueNotFoundError:
+            return VirtualValue(VirtualHive(), value, default)
 
     def values(self) -> list[RegistryValue]:
         """Returns a list of all the values from this key."""
--- a/dissect/target/helpers/utils.py
+++ b/dissect/target/helpers/utils.py
@@ -40,7 +40,7 @@ def to_list(value: T | list[T] | None) -> list[T]:
     """
     if value is None:
         return []
-    elif not isinstance(value, list):
+    if not isinstance(value, list):
         return [value]
 
     return value
--- a/dissect/target/helpers/utils.py
+++ b/dissect/target/helpers/utils.py
@@ -29,8 +29,8 @@ def findall(buf: bytes, needle: bytes) -> Iterator[int]:
 T = TypeVar("T")
 
 
-def to_list(value: T | list[T]) -> list[T]:
-    """Convert a single value or a list of values to a list.
+def to_list(value: T | list[T] | None) -> list[T]:
+    """Convert a single value or a list of values to a list. A value of `None` is converted to an empty list.
 
     Args:
         value: The value to convert.
@@ -38,8 +38,11 @@ def to_list(value: T | list[T]) -> list[T]:
     Returns:
         A list of values.
     """
-    if not isinstance(value, list):
+    if value is None:
+        return []
+    elif not isinstance(value, list):
         return [value]
+
     return value
 
 
--- a/dissect/target/helpers/utils.py
+++ b/dissect/target/helpers/utils.py
@@ -30,7 +30,7 @@ T = TypeVar("T")
 
 
 def to_list(value: T | list[T] | None) -> list[T]:
-    """Convert a single value or a list of values to a list. A value of `None` is converted to an empty list.
+    """Convert a single value or a list of values to a list. A value of ``None`` is converted to an empty list.
 
     Args:
         value: The value to convert.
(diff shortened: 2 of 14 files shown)


dissect/target/helpers/regutil.py
```diff
@@ -172,6 +172,18 @@ def value(self, value: str) -> RegistryValue:
         """
         raise NotImplementedError()
 
+    def value_or_default(self, value: str, default: ValueType = None) -> RegistryValue:
```
Schamper, line 175: Can you roll this into the `value` method so it's more in line with standard library Python code?

dissect/target/helpers/utils.py
```diff
@@ -29,17 +29,20 @@ def findall(buf: bytes, needle: bytes) -> Iterator[int]:
 T = TypeVar("T")
 
 
-def to_list(value: T | list[T]) -> list[T]:
-    """Convert a single value or a list of values to a list.
+def to_list(value: T | list[T] | None) -> list[T]:
+    """Convert a single value or a list of values to a list. A value of `None` is converted to an empty list.
```
Schamper, line 33:
```suggestion
    """Convert a single value or a list of values to a list. A value of ``None`` is converted to an empty list.
```

dissect/target/plugins/os/windows/registry.py
```diff
@@ -282,11 +286,31 @@ def key(self, key: Optional[str] = None) -> KeyCollection:
 
         return res
 
+    @internal
+    def key_or_empty(self, key: Optional[str] = None) -> KeyCollection:
```
Schamper, line 290: Same with this one.

Schamper (changes requested):

dissect/target/helpers/utils.py
```diff
@@ -29,17 +29,20 @@ def findall(buf: bytes, needle: bytes) -> Iterator[int]:
 T = TypeVar("T")
 
 
-def to_list(value: T | list[T]) -> list[T]:
-    """Convert a single value or a list of values to a list.
+def to_list(value: T | list[T] | None) -> list[T]:
```
twiggler, line 32: To Michaël: For our usecases, your implementation won in the end.

Horofic (changes requested):

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,319 @@
...
+
+        ns = {"": "http://schemas.microsoft.com/office/appforoffice/1.1"}
+
+        manifest_tree: Element = ElementTree.fromstring(manifest_path.read_text("utf-8-sig"))
+
+        source_location_elements = manifest_tree.findall(".//SourceLocation", ns)
+        source_locations = [source_location.get("DefaultValue") for source_location in source_location_elements]
+
+        display_name_element = manifest_tree.find(".//DisplayName", ns)
+        display_name = display_name_element.get("DefaultValue") if display_name_element is not None else None
```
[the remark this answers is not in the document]
twiggler: No, see https://docs.python.org/3/reference/datamodel.html#object.__delitem__

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,319 @@
...
+                yield self._parse_web_addin_manifest(manifest_file)
+            except Exception as e:
+                self.target.log.warning("Error parsing web-addin manifest %s", manifest_file)
+                self.target.log.debug("", exc_info=e)
+
+    @export(record=OfficeNativeAddinRecord)
+    def native(self) -> Iterator[OfficeNativeAddinRecord]:
+        """List all native (COM / vsto) add-ins by parsing the registry and manifest files."""
+
+        addin_path_tuples = itertools.product(self.HIVES, [self.OFFICE_KEY], self.OFFICE_COMPONENTS, [self.ADD_IN_KEY])
```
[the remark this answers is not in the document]
twiggler: Tried but it gets messier IMO

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # Typically, all components share the same root.
+        # Curiously enough, the "Common" component has no InstallRoot defined.
+        key = f"HKLM\\{self.OFFICE_KEY}\\16.0\\{component}\\InstallRoot"
+        if office_key := self.target.registry.value_or_empty(key, "Path"):
+            return office_key.value
+
+        return self.OFFICE_DEFAULT_ROOT
+
+    def _machine_startup_folders(self) -> Iterable[str]:
```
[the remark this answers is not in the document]
twiggler: > Can't OFFICE_COMPONENTS be reused here? 
We could eagerly create a map from component to root directory, but I am unsure if that is better. 

Added a union of string literals to prevent mistakes.

> I think this can also be consolidated in just the _office_install_root function

This got a bit messy.

Added memoization to avoid looking up multiple times.

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+
+        See https://learn.microsoft.com/en-us/visualstudio/vsto/registry-entries-for-vsto-add-ins?view=vs-2022#LoadBehavior # noqa: E501
+        """
+
+        load_behavior = addin.value("LoadBehavior", None).value
+
+        if load_behavior is None:
+            return None
+        elif load_behavior == 0:
+            return NativePluginStatus(loaded=False, load_behavior=LoadBehavior.Manual)
```
Horofic, line 399: Wouldn't a `Dataclass` be a better approach for this? With the initial `loaded` and `load_behaviour` set to `None`. 

This this way you always return something, and avoids the `if None` checking in the `native()` function and potentially somewhere else.

Horofic (changes requested):

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+        yield fsutil.join(self._office_install_root("Word"), "Templates", alt_separator="\\")
+        yield fsutil.join(self._office_install_root("Word"), "Document Themes", alt_separator="\\")
+
+    def _parse_plugin_status(self, addin: KeyCollection) -> NativePluginStatus | None:
+        """Parse the registry value which controls if the add-in autostarts.
+
+        See https://learn.microsoft.com/en-us/visualstudio/vsto/registry-entries-for-vsto-add-ins?view=vs-2022#LoadBehavior # noqa: E501
+        """
+
+        load_behavior = addin.value("LoadBehavior", None).value
```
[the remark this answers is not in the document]
twiggler: No, it will return ``None`` because we are returning a RegistryValue with a value of ``None``.

(I do see that the ValueType type union was missing ``None``, added it, but that didn't cause runtime-errors)

dissect/target/plugins/apps/productivity/msoffice.py
twiggler: That would imply that either one of `loaded` and `load_behavior` can be ``None``, which cannot happen; either both values are present, or they are both absent.

So I think it is better to leave as-is.

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+            return set()
+
+        dependent_assemblies: set[self.Assembly] = set()
+        dependent_assembly_elements = manifest_tree.findall(".//dependentAssembly", self.XML_NAMESPACE)
+        for dependent_assembly_element in dependent_assembly_elements:
+            dependent_assemblies |= self._parse_dependent_assembly(dependent_assembly_element, manifest_path.parent)
+
+        return dependent_assemblies
+
+    def _parse_dependent_assembly(self, dependent_assembly: Element, cwd: Path) -> set[Assembly]:
```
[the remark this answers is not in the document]
twiggler: AFAIK The function can return an empty set, but never a set with None values.

For maximum clarity I added a type hint to the empty sets

dissect/target/helpers/regutil.py
```diff
@@ -161,15 +161,27 @@ def subkeys(self) -> list[RegistryKey]:
...
         Raises:
-            RegistryValueNotFoundError: If this key has no value with the requested name.
+            RegistryValueNotFoundError: If this key has no value with the requested name and default is not defined.
         """
+        try:
+            return self._value(value)
+        except RegistryValueNotFoundError:
+            if default is self.__marker:
+                raise
+            return VirtualValue(VirtualHive(), value, default)
```
Schamper, line 182: Wouldn't we still want the actual hive to be linked?

dissect/target/loaders/cb.py
```diff
@@ -161,7 +161,7 @@ def subkey(self, subkey: str) -> CbRegistryKey:
     def subkeys(self) -> list[CbRegistryKey]:
         return list(map(self.subkey, self.data["sub_keys"]))
 
-    def value(self, value: str) -> str:
+    def _value(self, value: str) -> str:
```
Schamper, line 164: Maybe update these return types while we're at it?

dissect/target/loaders/smb.py
```diff
@@ -308,7 +308,7 @@ def subkeys(self) -> list[SmbRegistryKey]:
 
         return subkeys
 
-    def value(self, value: str) -> str:
+    def _value(self, value: str) -> str:
```
Schamper, line 311: Maybe update these return types while we're at it?

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # Typically, all components share the same root.
+        # Curiously enough, the "Common" component has no InstallRoot defined.
+        key = f"HKLM\\{self.OFFICE_KEY}\\16.0\\{component}\\InstallRoot"
+        try:
+            return self.target.registry.value(key, "Path").value
+        except RegistryError:
+            return self.OFFICE_DEFAULT_ROOT
+
+    @lru_cache(16)
```
Schamper, line 379: Can you initialise this in the `__init__`? Search for other uses of `lru_cache` in the Dissect project for examples.

If you want context, hit me up offline.

Schamper (changes requested):

dissect/target/plugins/apps/productivity/msoffice.py
Horofic: That's true, but you end up setting them to `None` anyway in the `OfficeNativeAddinRecord` anyway. So I thought the suggested approach would be a bit nicer.

dissect/target/plugins/apps/productivity/msoffice.py
Horofic: Posted this in an earlier comment. Some more context can also be found here https://github.com/fox-it/dissect.target/pull/547!

dissect/target/plugins/apps/productivity/msoffice.py
twiggler: I appreciate the suggestion and agree that it is  clunky at the call-site.

However, I think it comes down to Python lacking an optional chaining operator such as in `status?.loaded`, 
and I would rather not trade invariants to compensate for that.

dissect/target/plugins/apps/productivity/msoffice.py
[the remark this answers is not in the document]
twiggler: >  Though, I wasn't aware it returned a RegisteryValue with a None value in it

Yeah IMO the pattern is different in that `value(.)` always is wrapped in the `RegistryValue` context. This incidentally also causes a stutter when using the API, i.e. `value(.).value`.

Interestingly how this relates to your suggestion here https://github.com/fox-it/dissect.target/pull/966#discussion_r1914865963, i.e. the difference between a value wrapped in an optional, and an optional embedded in a value.

I guess this can be resolved now.

dissect/target/loaders/cb.py
twiggler: I love types, so happy to do it.

But is there light at the end of the tunnel? 

Unfortunately, a lot of the type annotations are incorrect. As I am used to inferring the behavior of a function by its type signature, this has put me on the wrong foot on multiple occasions. The type annotations are incorrect because the code is not type checked, either in CI or locally. I think one of the problems with enforcing type checking is that core dependencies such as flow record and cstruct use dynamic typing. 
I believe @Miauwkeru mentioned a solution for somehow adding type information to cstruct. Or we could use code generation at "compile-time" instead of during run-time. 

Happy to help out with this (also in my own time). However,  I also think a principled decision needs to be made to stop relying on techniques which use dynamic types.

Schamper: The `.pyi` generation for cstruct was indeed my idea, of which Miauwkeru made a POC. Happy to hear other suggestions of yours.

However, 95% of code in `dissect.target` does not interact with `flow.record` or `dissect.cstruct`, and we still want accurate type information there. I agree that it's best if we can check that in CI or locally. So I think the best way forward for the time being is to figure out if there's some way to do that and ignore all `flow.record` and `dissect.cstruct` types while we work separately on a solution for those dynamic types.

twiggler: (I take it the return type can be covariant?)

Horofic (changes requested):

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+        """Returns all startup items found in Microsoft Office startup folders.
+
+        Office startup folders are specific directories where Microsoft Office looks add-ins, macros, templates, or custom scripts.
+        These are used to automatically load when the corresponding Office application starts up.
+        These folders allow users and administrators to automate launching add-ins, executing scripts, or applying custom settings.
+
+        References:
+            - https://pentestlab.blog/2019/12/11/persistence-office-application-startup/
+
+        Yields a OfficeNativeAddinRecord with fields:
```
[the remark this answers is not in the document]
Horofic: bump

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+        VSTO is a set of tools provided by Microsoft to create Office add-ins using the .NET Framework.
+        VSTO plugins are more modern than COM plugins and leverage managed code. They are typically developed in C# or VB.NET using Visual Studio.
+
+        Both COM and VSTO add-ins are registered in the Windows registry, where they are associated with specific Office applications and configured to load automatically or on demand.
+
+        References:
+            - https://learn.microsoft.com/en-us/office/dev/add-ins/overview/office-add-ins
+            - https://learn.microsoft.com/en-us/visualstudio/vsto/registry-entries-for-vsto-add-ins
+
+        Yields a OfficeNativeAddinRecord with fields:
```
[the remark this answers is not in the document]
Horofic: bump

dissect/target/plugins/apps/productivity/msoffice.py
```diff
@@ -0,0 +1,413 @@
...
+
+        Office Web Add-ins are web-based applications that extend the functionality of Office applications like Word, Excel, and Outlook.
+        These add-ins can interact with the content in Office documents and provide additional features and capabilities.
+        The WEF folder contains cached data and manifests for Office Web Add-ins.
+        The manifest includes information about the add-ins, such as their source locations, display names, and other metadata.
+
+        References:
+            - https://learn.microsoft.com/en-us/office/dev/add-ins/overview/office-add-ins
+
+        Yields a OfficeWebAddinRecord with fields:
```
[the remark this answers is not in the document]
Horofic: bump

dissect/target/plugins/os/windows/registry.py
```diff
@@ -299,10 +303,10 @@ def iterkeys(self, keys: Union[str, list[str]]) -> Iterator[KeyCollection]:
             yield key
 
     @internal
-    def keys(self, keys: Union[str, list[str]]) -> Iterator[KeyCollection]:
+    def keys(self, keys: Union[str, list[str]]) -> Iterator[RegistryKey]:
```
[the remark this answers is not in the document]
Horofic: bump

dissect/target/plugins/os/windows/registry.py
```diff
@@ -316,6 +320,19 @@ def keys(self, keys: Union[str, list[str]]) -> Iterator[KeyCollection]:
             except HiveUnavailableError:
                 pass
 
+    @internal
+    def values(self, keys: Union[str, list[str]], value: str) -> Iterator[RegistryValue]:
```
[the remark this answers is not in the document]
Horofic: bump

tests/plugins/filesystem/test_resolver.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
Horofic: bump

Schamper (approved):

PR #967, opened 2024-12-06
Remove paho from mqqt unit test

### LINKED ISSUE #926: Remove dependency on `paho-mqtt` in unit test
https://github.com/fox-it/dissect.target/blob/c9b4aa10846aeabcafcf3ca3bc4baa88c58da669/tests/loaders/test_mqtt.py

It's super annoying and unnecessary that this test requires `paho-mqtt`. It uses nothing from the library. It should be more similar to the Carbon Black and SMB unit tests and properly mock it.

twiggler:
We remove an import at the top, but now have to import the unit under test in the test itself.

Closes #926

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,7 +80,7 @@ full = [
     "zstandard",
 ]
 dev = [
-    "dissect.target[full,mqtt,yara]",
+    "dissect.target[full,yara]",
     "dissect.archive[dev]>=1.0.dev,<2.0.dev",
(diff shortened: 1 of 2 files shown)


twiggler:
> It might also be an idea to remove mqtt from line 83 of the pyproject.toml as there now shouldn't be a dependency requiring it anymore during tests no?

@Miauwkeru It is no longer required for the test, but it should still be when using the `mqtt` loader right?

PR #971, opened 2024-12-10
Add warning for OBJECT_UNUSED in journal plugin

JSCU-CNI:
This PR adds an additional check for ObjectType `OBJECT_UNUSED` when iterating to a new journal entry array. It is possible that space is allocated in a journal file which is not yet used to store the next entry array (e.g. the object resides in memory and is not yet flushed to disk). The plugin would then attempt to read this empty data and return a `ValueError` since it is not expecting a zero byte.

We have changed this to a warning and added a break to stop iterating since this ObjectType can only be encountered at the end of a journal file.

### DIFF
--- a/dissect/target/plugins/os/windows/regf/cam.py
+++ b/dissect/target/plugins/os/windows/regf/cam.py
@@ -1,0 +1,118 @@
+from typing import Iterator
+
+from dissect.util.ts import wintimestamp
+from flow.record.fieldtypes import windows_path
+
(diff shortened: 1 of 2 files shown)


Schamper:
Would it be possible to add a unit test?

Schamper:
Would it be possible to fabricate one with some hex editing or a cstruct write?

Schamper (approved):

PR #972, opened 2024-12-10
Prevent enumerating entire non-ISO formatted syslog files in `is_iso_fmt`

JSCU-CNI:
This PR prevents `is_iso_fmt` from enumerating entire non-ISO formatted auth or syslogs.


dissect/target/plugins/os/unix/log/helpers.py
```diff
@@ -22,12 +22,17 @@
 )
 
 
-def iso_readlines(file: Path) -> Iterator[tuple[datetime, str]]:
+def iso_readlines(file: Path, max_lines: int | None = None) -> Iterator[tuple[datetime, str]]:
     """Iterator reading the provided log file in ISO format. Mimics ``year_rollover_helper`` behaviour."""
     with open_decompress(file, "rt") as fh:
-        for line in fh:
+        for i, line in enumerate(fh):
+            if isinstance(max_lines, int) and i > max_lines:
```
Schamper, line 29:
```suggestion
            if max_lines is not None and i > max_lines:
```

Schamper (changes requested):
Can you also add a unit test for this specific edge case

dissect/target/plugins/os/unix/log/helpers.py
```diff
@@ -22,12 +22,17 @@
...
+                log.debug("Stopping iso_readlines enumeration in %s: max_lines=%s was reached", file, max_lines)
+                break
+
             if not (match := RE_TS_ISO.match(line)):
-                log.warning("No timestamp found in one of the lines in %s!", file)
+                if not max_lines:
+                    log.warning("No timestamp found in one of the lines in %s!", file)
```
Poeloe, line 35: Is it desired that you now only get debug messages and not a warning when the `max_lines` value is reached?

Schamper:
@Poeloe

Poeloe (approved):

PR #973, opened 2024-12-10
Improve LayerFilesystem iterdir and scandir

JSCU-CNI:
This PR prevents `NotADirectoryError` `DirectoryFilesystemEntry` exceptions from crashing plugins when iterating `LayerFilesystem` entries using `recurse`, `iterdir` or `scandir`.

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -840,11 +840,12 @@ class FilesystemEntry:
 class VirtualDirectory(FilesystemEntry):
     """Virtual directory implementation. Backed by a dict."""
 
-    def __init__(self, fs, path):
+    def __init__(self, fs, path, stat: fsutil.stat_result | None = None):
         super().__init__(fs, path, None)
         self.up = None
         self.top = None
         self.entries = {}
+        self._stat = stat
 
     def __getitem__(self, item) -> FilesystemEntry:
         if not self.fs.case_sensitive:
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/filesystem/walkfs.py
```diff
@@ -37,6 +37,15 @@ def check_compatible(self) -> None:
     @arg("--walkfs-path", default="/", help="path to recursively walk")
     def walkfs(self, walkfs_path: str = "/") -> Iterator[FilesystemRecord]:
         """Walk a target's filesystem and return all filesystem entries."""
+
+        if not self.target.fs.path(walkfs_path).exists():
```
Schamper, line 41: Would be nicer to not have to call path on this twice.

dissect/target/filesystem.py
```diff
@@ -1535,6 +1535,9 @@ def iterdir(self) -> Iterator[str]:
         yielded = {".", ".."}
         selfentry = self._resolve()
         for fsentry in selfentry.entries:
+            if not fsentry.is_dir():
```
Schamper, line 1538: Can you add unit tests?

Schamper (changes requested):

Schamper:
@JSCU-CNI can you still add a unit test that showcases the originally faulty behavior so that I can try some other ways to fix this?

PR #978, opened 2024-12-19
Make Mssql errorlog plugin more robust

### LINKED ISSUE #977: Mssql errorlog plugin fails unexpectedly during compatibility check
Mssql `errorlog` plugin fails unexpectedly during compatibility check in case the `"HKLM\\SOFTWARE\\Microsoft\\Microsoft SQL Server` key is not present on the target with a  `RegistryKeyNotFoundError`

Horofic:
Fixes #977

### DIFF
--- a/dissect/target/plugins/os/windows/log/mssql.py
+++ b/dissect/target/plugins/os/windows/log/mssql.py
@@ -90,7 +90,7 @@ class MssqlPlugin(Plugin):
                     buf += line
 
     def _find_instances(self) -> set[str, TargetPath]:
-        instances: set = set()
+        instances = set()
 
         for subkey in self.target.registry.glob_ext(self.MSSQL_KEY_GLOB):
             if "." in subkey.name:
--- a/dissect/target/plugins/os/windows/log/mssql.py
+++ b/dissect/target/plugins/os/windows/log/mssql.py
@@ -90,13 +90,7 @@ class MssqlPlugin(Plugin):
                     buf += line
 
     def _find_instances(self) -> set[str, TargetPath]:
-        instances = set()
-
-        for subkey in self.target.registry.glob_ext(self.MSSQL_KEY_GLOB):
-            instances.add(
-                (
-                    subkey.name,
-                    self.target.fs.path(subkey.subkey("SQLServerAgent").value("ErrorLogFile").value).parent,
-                )
-            )
-        return instances
+        return {
+            (subkey.name, self.target.fs.path(subkey.subkey("SQLServerAgent").value("ErrorLogFile").value).parent)
+            for subkey in self.target.registry.glob_ext(self.MSSQL_KEY_GLOB)
+        }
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/log/mssql.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, target: Target):
 
     def check_compatible(self) -> None:
         if not self.instances:
-            raise UnsupportedPluginError("System does not seem to be running SQL Server")
+            raise UnsupportedPluginError("No Microsoft SQL Server instances have been found.")
```
Schamper, line 47:
```suggestion
            raise UnsupportedPluginError("No Microsoft SQL Server instances have been found")
```

dissect/target/plugins/os/windows/log/mssql.py
```diff
@@ -89,12 +89,12 @@ def errorlog(self) -> Iterator[MssqlErrorlogRecord]:
 
                     buf += line
 
-    def _find_instances(self) -> list[str, TargetPath]:
-        instances = []
+    def _find_instances(self) -> set[str, TargetPath]:
+        instances: set = set()
```
Schamper, line 93:
```suggestion
        instances = set()
```

dissect/target/plugins/os/windows/log/mssql.py
```diff
@@ -89,12 +89,12 @@ def errorlog(self) -> Iterator[MssqlErrorlogRecord]:
...
 
-        for subkey in self.target.registry.key(self.MSSQL_KEY).subkeys():
-            if subkey.name.startswith("MSSQL") and "." in subkey.name:
-                instances.append(
+        for subkey in self.target.registry.glob_ext(self.MSSQL_KEY_GLOB):
```
Schamper, line 95: I think using `keys` instead of `key` also would've resolved this.

Schamper (changes requested):

dissect/target/plugins/os/windows/log/mssql.py
Horofic: Thats correct. Would you prefer that? In this case the globing works nice with the MSSQL<pattern>.<instance_name> subkeys.

Schamper: I think since you're doing additional filtering after the glob anyway, might as well check everything without the glob. Probably a bit faster too.

If you could do this with just a glob, then that'd probably be nicer.

dissect/target/plugins/os/windows/log/mssql.py
```diff
@@ -89,15 +89,14 @@ def errorlog(self) -> Iterator[MssqlErrorlogRecord]:
...
+        instances = set()
 
-        for subkey in self.target.registry.key(self.MSSQL_KEY).subkeys():
-            if subkey.name.startswith("MSSQL") and "." in subkey.name:
-                instances.append(
-                    (
-                        subkey.name,
-                        self.target.fs.path(subkey.subkey("SQLServerAgent").value("ErrorLogFile").value).parent,
-                    )
+        for subkey in self.target.registry.glob_ext(self.MSSQL_KEY_GLOB):
```
Schamper, line 95: Maybe make this a comprehension, not sure if it'll become too large.

Schamper (changes requested):

Schamper (approved):

PR #980, opened 2025-01-08
Improve ApachePlugin log file discovery

### LINKED ISSUE #1013: webserver.apache : Support Include/IncludeOptionnal directive in apache configurations files
Currently, webserver.apache parser try to read configuration from apache in order to add possible access/error logs path (https://github.com/fox-it/dissect.target/blob/6fb0e5aa54575f63529d5e5c1c8d1c319b3d28b4/dissect/target/plugins/apps/webserver/apache.py#L219
 
If vhost configuration made in dedicated files using the [Include](https://httpd.apache.org/docs/2.4/mod/core.html#include) or [IncludeOptional](https://httpd.apache.org/docs/2.4/mod/core.html#includeoptional) they are ignored.

It could be possible to:
* Retrieve ServerRoot (used as base fore relative path)
* Retrieve included path using Include(Optionnal) Directive
* Parse identified vhost configuration files included in order to extract potential access/error logs path.

I can make a PR if you want.

JSCU-CNI:
This PR improves the log file discovery of the ApachePlugin on targets.

* We now parse `Include` and `IncludeOptional` directives in all apache `.conf` files (mods, sites and confs) and extract `ErrorLog` and `CustomLog` from those configuration files.
* We now substitute Apache `envvars` from those log file location directives (e.g. `${APACHE_LOG_DIR}`).

Fixes #1013.


dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,66 +216,153 @@ class ApachePlugin(WebserverPlugin):
...
+
+            elif "Include" in line:
+                match = RE_CONFIG_INCLUDE.match(line)
+                if not match:
+                    self.target.log.warning("Unexpected Apache 'Include' configuration in %s: '%s'", path, line)
+                    continue
+
+                directive = match.groupdict()
+
+                if "*" in directive["location"]:
```
william-billaud, line 292: Why do not use glob directly?

As python glob rely on fnmatch and apache documentation explicitly mark the location as being fnmatch style (https://httpd.apache.org/docs/2.4/en/mod/core.html#include) 
> Shell-style (fnmatch()) wildcard characters can be used in the filename`

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,66 +216,153 @@ class ApachePlugin(WebserverPlugin):
...
-                        pattern_to_use = RE_CONFIG_ERRORLOG_DIRECTIVE
-                    else:
-                        set_to_update = access_log_paths
-                        pattern_to_use = RE_CONFIG_CUSTOM_LOG_DIRECTIVE
+        return self.access_paths, self.error_paths
 
-                    match = pattern_to_use.match(line)
-                    if not match:
-                        self.target.log.warning("Unexpected Apache log configuration: %s (%s)", line, path)
+    def _process_conf_file(self, path: Path) -> None:
```
william-billaud, line 264: Maybe it's would be better to add either a `max_recurse` integer or to create a set of already processed file in order to prevent infinite recursion?

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,166 @@ class ApachePlugin(WebserverPlugin):
...
+                seen.add(path)
+
+        # Check all .conf files inside the server root
+        if self.server_root:
+            for path in self.server_root.rglob("*.conf"):
+                if path not in seen:
+                    self._process_conf_file(path, seen=None)
+                    seen.add(path)
+
+    def _process_conf_file(self, path: Path, seen: set[Path] | None = set()) -> None:
```
Schamper, line 269:
```suggestion
    def _process_conf_file(self, path: Path, seen: set[Path] | None = None) -> None:
```
The caching issue is because you initialize the `seen` argument with a `set()` default, which is a mutable object. This mutable object will then be used across invocations of different instances of the class. The correct way to use this pattern is to initialize it as `None` in the default argument, and then initialize it to a new `set` in the function body.

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,166 @@ class ApachePlugin(WebserverPlugin):
...
+        and ``OptionalInclude`` directives. Populates ``self.access_paths`` and ``self.error_paths``.
+        """
 
-                    if not line or ("CustomLog" not in line and "ErrorLog" not in line):
-                        continue
+        self.target.log.debug("Processing conf file: %s", path)
+        self.target.log.debug("Current seen conf files: %s", seen)
+
+        if not seen:
+            seen = set()
```
Schamper, line 278:
```suggestion
        seen = seen or set()
```

Schamper (changes requested):
See the comment, I think this makes the change to a factory redundant.

tests/plugins/apps/webserver/test_citrix.py
```diff
@@ -75,16 +78,18 @@ def test_error_logs(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
...
-    assert len(error_log_paths) == 2
+    assert len(target_citrix.citrix.error_paths) == 2
 
 
 def test_access_logs_webserver_namespace(target_citrix: Target, fs_bsd: VirtualFilesystem) -> None:
     data_file = absolute_path("_data/plugins/apps/webserver/citrix/httpaccess.log")
     fs_bsd.map_file("var/log/httpaccess.log", data_file)
 
-    target_citrix.add_plugin(ApachePlugin, check_compatible=False)
+    with pytest.raises(UnsupportedPluginError):
```
Schamper, line 90: Can you add a `match=""` with the expected error message?

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
-        if not len(self.access_log_paths) and not len(self.error_log_paths):
-            raise UnsupportedPluginError("No Apache directories found")
+        if not self.access_paths and not self.error_paths:
+            raise UnsupportedPluginError("No Apache log files found")
 
-    def get_log_paths(self) -> tuple[list[Path], list[Path]]:
-        """
-        Discover any present Apache log paths on the target system.
+        if self.target.os == OperatingSystem.CITRIX:
+            raise UnsupportedPluginError(f"Use {self.target.os} Apache plugin instead")
```
Schamper, line 234: I meant more something that actually included the function name to use instead 😄 now I still don't know what to replace in my `target-query -f` command.

Schamper (changes requested):

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+            if (path := self.target.fs.path(config)).exists() and path not in seen:
+                self._process_conf_file(path)
+                seen.add(path)
+
+        # Check all .conf files inside the server root
+        if self.server_root:
+            for path in self.server_root.rglob("*.conf"):
+                if path not in seen:
+                    self._process_conf_file(path)
+                    seen.add(path)
```
Schamper, line 267:
```suggestion
            if (path := self.target.fs.path(config)).exists() and path not in seen:
                self._process_conf_file(path, seen)

        # Check all .conf files inside the server root
        if self.server_root:
            for path in self.server_root.rglob("*.conf"):
                if path not in seen:
                    self._process_conf_file(path, seen)
```
Would this work? I think theoretically you would process the same file twice with this if you don't pass the `seen` here, if one of these base files has a reference back to itself (or another default file path).

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+
+        seen.add(path)
+
+        for line in path.open("rt"):
+            line = line.strip()
+
+            if not line:
+                continue
+
+            elif "ServerRoot" in line:
```
Schamper, line 287:
```suggestion
            if not (line := line.strip()):
                continue

            if "ServerRoot" in line:
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+                    continue
+                location = match.groupdict().get("location")
+                self.server_root = self.target.fs.path(location)
+
+            elif "CustomLog" in line or "ErrorLog" in line:
+                self._process_conf_line(path, line)
+
+            elif "Include" in line:
+                match = RE_CONFIG_INCLUDE.match(line)
+                if not match:
```
Schamper, line 300:
```suggestion
                if not (match := RE_CONFIG_INCLUDE.match(line)):
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+
+        for line in path.open("rt"):
+            line = line.strip()
+
+            if not line:
+                continue
+
+            elif "ServerRoot" in line:
+                match = RE_CONFIG_ROOT.match(line)
+                if not match:
```
Schamper, line 289:
```suggestion
                if not (match := RE_CONFIG_ROOT.match(line)):
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+                elif (include_path := self.target.fs.path(location)).exists():
+                    self._process_conf_file(include_path, seen)
+
+                else:
+                    self.target.log.warning("Unable to resolve Apache Include in %s: %r", path, line)
+
+    def _process_conf_line(self, path: Path, line: str) -> None:
+        """Parse and resolve the given ``CustomLog`` or ``ErrorLog`` directive found in a Apache ``.conf`` file."""
+        line = line.strip()
+
```
Schamper, line 331:
```suggestion
```
Line is already stripped.

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+        if "ErrorLog" in line:
+            pattern = RE_CONFIG_ERRORLOG_DIRECTIVE
+            set_to_update = self.error_paths
+
```
Schamper, line 335:
```suggestion
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+        else:
+            pattern = RE_CONFIG_CUSTOM_LOG_DIRECTIVE
+            set_to_update = self.access_paths
+
+        match = pattern.match(line)
+        if not match:
```
Schamper, line 341:
```suggestion
        if not (match := pattern.match(line)):
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+            return
+
+        location = match.groupdict().get("location")
+        custom_log = self.target.fs.path(location)
+
+        if env_var_match := RE_ENV_VAR_IN_STRING.match(location):
+            envvar_directive = env_var_match.groupdict()
+            env_var = envvar_directive["env_var"]
+            apache_log_dir = self.env_vars.get(env_var)
+            if apache_log_dir is None:
```
Schamper, line 352:
```suggestion
            if (apache_log_dir := self.env_vars.get(env_var)) is None:
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+        if not match:
+            self.target.log.warning("Unexpected Apache 'ErrorLog' or 'CustomLog' configuration in %s: %r", path, line)
+            return
+
+        location = match.groupdict().get("location")
+        custom_log = self.target.fs.path(location)
+
+        if env_var_match := RE_ENV_VAR_IN_STRING.match(location):
+            envvar_directive = env_var_match.groupdict()
+            env_var = envvar_directive["env_var"]
```
Schamper, line 350:
```suggestion
            env_var = env_var_match.groupdict()["env_var"]
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+                self.target.log.warning("%s does not exist, cannot resolve '%s' in %s", env_var, custom_log, path)
+                return
+
+            custom_log = self.target.fs.path(location.replace(f"${{{env_var}}}", apache_log_dir.replace("$SUFFIX", "")))
+
+        set_to_update.update(path for path in custom_log.parent.glob(f"*{custom_log.name}*"))
 
-        return sorted(access_log_paths), sorted(error_log_paths)
+    @cached_property
+    def env_vars(self) -> dict:
```
Schamper, line 361:
```suggestion
    def env_vars(self) -> dict[str, str]:
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,161 @@ class ApachePlugin(WebserverPlugin):
...
+        variables = {}
+        for envvar_file in self.DEFAULT_ENVVAR_PATHS:
+            if (file := self.target.fs.path(envvar_file)).exists():
+                for line in file.read_text().split("\n"):
```
Schamper, line 365: Is this guaranteed to use `\n` and never `\r\n`? Otherwise `.splitlines()` might be better?

Schamper (changes requested):

dissect/target/plugins/apps/webserver/apache.py
Schamper: What about this one?

Schamper (changes requested):

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,152 @@ class ApachePlugin(WebserverPlugin):
...
+            custom_log = self.target.fs.path(location.replace(f"${{{env_var}}}", apache_log_dir.replace("$SUFFIX", "")))
+
+        set_to_update.update(path for path in custom_log.parent.glob(f"*{custom_log.name}*"))
+
+    @cached_property
+    def env_vars(self) -> dict[str, str]:
+        variables = {}
+        for envvar_file in self.DEFAULT_ENVVAR_PATHS:
+            if (file := self.target.fs.path(envvar_file)).exists():
+                for line in file.open("rt").readlines():
```
Schamper, line 355:
```suggestion
                for line in file.read_text().splitlines():
```

dissect/target/plugins/apps/webserver/apache.py
```diff
@@ -187,65 +217,152 @@ class ApachePlugin(WebserverPlugin):
...
+                    line = line.strip()
+                    key, _, value = line.partition("=")
```
Schamper, line 357:
```suggestion
                    key, _, value = line.strip().partition("=")
```

Schamper (changes requested):

dissect/target/plugins/apps/webserver/apache.py
Schamper: There's a flaw in your current logic that allows for files to be processed twice, this fixes that.

Say you have the following configuration:
```
/etc/apache2/apache2.conf
    # Normal config
/etc/httpd.conf
    # References /etc/apache2/apache2.conf
```

With your current logic, `/etc/apache2/apache2.conf` will be processed twice.

Take this code as an example:

```python
def _process_conf_file(path, seen=None, content=None):
    seen = set() if seen is None else seen
    if path in seen:
        return
    seen.add(path)

    print("processing", path)

    for key, value in content.items():
        _process_conf_file(key, seen, value)


MOCK_DATA = {
    "/path/a": {
        "/path/a/sub": {},
        "/path/a/sub/sub": {},
    },
    "/path/b": {
        "/path/b/sub": {},
        "/path/a": {},
    },
}

for path, value in MOCK_DATA.items():
    _process_conf_file(path, None, value)
```

`content` is just to disregard processing and emulate a recursive parse of a next file. Running this prints:
```
processing /path/a
processing /path/a/sub
processing /path/a/sub/sub
processing /path/b
processing /path/b/sub
processing /path/a  # <-- processed twice
```
Simply changing the loop to this fixes it:
```python
seen = set()
for path, value in MOCK_DATA.items():
    _process_conf_file(path, seen, value)
```
```
processing /path/a
processing /path/a/sub
processing /path/a/sub/sub
processing /path/b
processing /path/b/sub
```

You can simplify it even further, there's no reason to check `path not in seen` in this function since you can let `_process_conf_file` handle that check. There's one change needed in `_process_conf_file` to make it work properly: `seen = set() if seen is None else seen`.

Schamper (approved):

PR #987, opened 2025-01-14
Don't attempt to load child targets when initial target loading fails

JSCU-CNI:
This PR makes sure `Target.open_all()` does not call `target.open_children()` if no `target` was loaded. This can occur when specifying a bogus target path in combination with `--children` to e.g. `target-query`.


dissect/target/target.py
```diff
@@ -310,7 +311,7 @@ def _find(find_path: Path, parsed_path: Optional[urllib.parse.ParseResult]):
                     except Exception as e:
                         getlogger(sub_entry).error("Failed to load target with loader %s", ldr, exc_info=e)
 
-                    if include_children:
+                    if target and include_children:
```
Schamper, line 314:
```suggestion
                    if target is not None and include_children:
```
Bit clearer perhaps.

Schamper (changes requested):

Schamper:
I proposed an alternative fix with https://github.com/fox-it/dissect.target/pull/987/commits/ffa0c9c890a8759d01130960e8148353cc7500f8

Schamper (approved):

PR #993, opened 2025-01-16
Add Kape VHDX loader

qmadev:
This PR adds a new loader to map a Kape VHDX file to a `Target`. The previous implementation only worked on directories created by Kape, but there is also an option to store the collected data in a VHDX file.

Never wrote a loader before so I'm not sure if this is the way to go.


Schamper:
Thanks for the contribution! I've not looked at the Kape VHDX format myself yet, but how similar is the directory structure to that of a Kape "directory"? If it matches, the logic could be simplified and re-use the existing Kape directory loading code. (Loaders and the directory filesystem also work on `TargetPath` objects, paths that come from within a Dissect opened filesystem)

tests/loaders/test_kape.py
```diff
@@ -37,3 +39,16 @@ def test_kape_loader(target_bare, tmp_path):
...
     assert len(target_bare.fs.mounts) == 5
     assert len(list(target_bare.fs.mounts["c:"].ntfs.usnjrnl.records())) == 1
+
+
+def test_kape_vhdx_loader(target_bare: Target):
+    t = Path(absolute_path("_data/loaders/kape/test.vhdx"))
+
+    assert KapeLoader.detect(t)
+
+    loader = KapeLoader(t)
```
Schamper, line 49:
```suggestion
def test_kape_vhdx_loader(target_bare: Target) -> None:
    p = Path(absolute_path("_data/loaders/kape/test.vhdx"))

    assert KapeLoader.detect(p)

    loader = KapeLoader(p)
```

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
     """
 
+    def __init__(self, path, **kwargs):
+        super().__init__(path)
+
+        if path.is_dir():
+            self.map = self.map_dir
+        if path.suffix == ".vhdx":
+            self.map = self.map_vhdx
+
```
Schamper, line 65:
```suggestion
```
Ideally just do this in code in the `.map()` method instead of swapping out methods.

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
+    container = VhdxContainer(path)
+    volume_system = volume.open(container)
+    for vol in volume_system.volumes:
+        yield filesystem.open(vol)
+
+
+def is_drive_letter(path: str) -> bool:
+    return len(path) == 1 or (len(path) == 2 and path[1] == ":")
+
+
```
Schamper, line 31:
```suggestion
```
This is now unused.

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
+        if path.is_dir():
+            return is_valid_kape_dir(path)
+        if path.suffix == ".vhdx":
+            return is_valid_kape_vhdx(path)
 
-        return False
+    def map_vhdx(self, target: Target) -> None:
+        for fs in open_vhdx(self.path):
+            self.path = fs.path()
+            self.map_dir(target)
```
Schamper, line 76:
```suggestion
```

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
 
-    def map(self, target: Target) -> None:
+    def map_dir(self, target: Target) -> None:
```
Schamper, line 78:
```suggestion
    def map(self, target: Target) -> None:
```

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
     def detect(path: Path) -> bool:
-        os_type, dirs = find_dirs(path)
-        if os_type == OperatingSystem.WINDOWS:
-            for dir_path in dirs:
-                for path in USNJRNL_PATHS:
-                    if dir_path.joinpath(path).exists():
-                        return True
+        if path.is_dir():
+            return is_valid_kape_dir(path)
+        if path.suffix == ".vhdx":
```
Schamper, line 70:
```suggestion
        if path.suffix.lower() == ".vhdx":
```

dissect/target/loaders/kape.py
```diff
@@ -15,25 +18,64 @@
...
+    def map_vhdx(self, target: Target) -> None:
+        for fs in open_vhdx(self.path):
+            self.path = fs.path()
+            self.map_dir(target)
 
-    def map(self, target: Target) -> None:
+    def map_dir(self, target: Target) -> None:
         find_and_map_dirs(
             target,
             self.path,
```
Schamper, line 81:
```suggestion
        path = self.path if self.path.is_dir() else next(open_vhdx(self.path)).path()

        find_and_map_dirs(
            target,
            path,
```

Schamper (approved):

PR #996, opened 2025-01-16
Fix crash in unix users plugins if GID or UID is empty

### LINKED ISSUE #995: unix : users plugin crash if UID/GID is empty in passwd file
If `/etc/passwd` file contains an entry without uid/GID (eg: /etc/passwd in compatibility mode), users plugins will crash, and also plugins that depends on this plugin (like commandhistory) 

Example the following entry

```
root:x:0:0:root:/root:/bin/bash
......
+@ngtest:x:::::
```

Will generate the following error

```
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/target.py", line 540, in add_plugin
    p = plugin_cls(self)
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/plugins/os/unix/history.py", line 46, in __init__
    self._history_files = list(self._find_history_files())
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/plugins/os/unix/history.py", line 55, in _find_history_files
    for user_details in self.target.user_details.all_with_home():
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/plugins/general/users.py", line 74, in all_with_home
    for user in self.target.users():
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/plugins/os/unix/_os.py", line 90, in users
    yield UnixUserRecord(
  File "/home/censored/Documents/tools/github/dissect.target/dissect/target/helpers/record.py", line 78, in __call__
    return super().__call__(*args, **kwargs)
  File "/home/censored/Documents/tools/github/dissect.target/venv/lib/python3.10/site-packages/flow/record/base.py", line 601, in __call__
    return self.recordType(*args, **kwargs)
  File "<string>", line 28, in __init__
  File "/home/censored/Documents/tools/github/dissect.target/venv/lib/python3.10/site-packages/flow/record/base.py", line 195, in __setattr__
    v = field_type(v)
ValueError: invalid literal for int() with base 10: ''
```

Information regarding compatibility mode could be found on the following resources:
* https://man.freebsd.org/cgi/man.cgi?query=passwd&sektion=5#end 
* https://man7.org/linux/man-pages/man5/nsswitch.conf.5.html

wbi-ocd:
This PR fix an issue while processing /etc/passwd entry where gid or uid id null.
This kind OS issue could also be fixed more generally in flow, by checking if entry is empty before coercing into an int, but I'm not sure if this would be the expected comportment for all entries.

fix #995

### DIFF
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -85,13 +85,12 @@ class UnixPlugin(OSPlugin):
                     current_user = (pwent.get(0), pwent.get(5), pwent.get(6))
                     if current_user in seen_users:
                         continue
-
                     seen_users.add(current_user)
                     yield UnixUserRecord(
                         name=pwent.get(0),
                         passwd=pwent.get(1),
-                        uid=pwent.get(2),
-                        gid=pwent.get(3),
+                        uid=pwent.get(2) if pwent.get(2) != "" else None,
+                        gid=pwent.get(3) if pwent.get(3) != "" else None,
                         gecos=pwent.get(4),
                         home=posix_path(pwent.get(5)),
                         shell=pwent.get(6),


dissect/target/plugins/os/unix/_os.py
```diff
@@ -75,7 +75,7 @@ def users(self, sessions: bool = False) -> Iterator[UnixUserRecord]:
         # Yield users found in passwd files.
         for passwd_file in PASSWD_FILES:
             if (path := self.target.fs.path(passwd_file)).exists():
-                for line in path.open("rt"):
+                for line in path.open("rt", errors="backslashreplace"):
```
Horofic, line 78:
```suggestion
                for line in path.open("rt", errors="surrogateescape"):
```

dissect/target/plugins/os/unix/_os.py
```diff
@@ -85,13 +85,12 @@ def users(self, sessions: bool = False) -> Iterator[UnixUserRecord]:
...
                         continue
-
                     seen_users.add(current_user)
                     yield UnixUserRecord(
                         name=pwent.get(0),
                         passwd=pwent.get(1),
-                        uid=pwent.get(2),
-                        gid=pwent.get(3),
+                        uid=pwent.get(2) if pwent.get(2) != "" else None,
+                        gid=pwent.get(3) if pwent.get(3) != "" else None,
```
Horofic, line 93:
```suggestion
                        uid=pwent.get(2) or None,
                        gid=pwent.get(3) or None,
```

Horofic (changes requested):

Horofic (approved):

PR #998, opened 2025-01-21
Fix file header detection in open_decompress

### LINKED ISSUE #999: Using open_decompress with fileobj can lead to incorrect file format detection
When using `fsutil.open_decompress` with a `fileobj`, it will start reading the magic bytes from the current position. This can lead to incorrect file format detection.

yunzheng:
Previously file header could be misread if fileobj position was not at start, leading to incorrect file format detection. Now explicitly seek to beginning before reading the magic file header bytes.

Also fixed the example in the docstring of `open_decompress` so that it renders it as a code block.

fixes #999

### DIFF
--- a/dissect/target/helpers/fsutil.py
+++ b/dissect/target/helpers/fsutil.py
@@ -512,10 +512,12 @@ def open_decompress(
         An binary or text IO stream, depending on the mode with which the file was opened.
 
     Example:
-        bytes_buf = open_decompress(Path("/dir/file.gz")).read()
+        .. code-block:: python
 
-        for line in open_decompress(Path("/dir/file.gz"), "rt"):
-            print(line)
+            bytes_buf = open_decompress(Path("/dir/file.gz")).read()
+
+            for line in open_decompress(Path("/dir/file.gz"), "rt"):
+                print(line)
     """
     if path and fileobj:
         raise ValueError("path and fileobj are mutually exclusive")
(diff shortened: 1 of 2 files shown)


dissect/target/helpers/fsutil.py
```diff
@@ -512,10 +512,12 @@ def open_decompress(
...
     Example:
-        bytes_buf = open_decompress(Path("/dir/file.gz")).read()
+        .. code-block:: python
 
-        for line in open_decompress(Path("/dir/file.gz"), "rt"):
-            print(line)
+            bytes_buf = open_decompress(Path("/dir/file.gz")).read()
+
+            for line in open_decompress(Path("/dir/file.gz"), "rt"):
+                print(line)
```
Schamper, line 520: Wouldn't just `Example::` achieve the same? (with a double `:`)

Schamper (changes requested):

dissect/target/helpers/fsutil.py
yunzheng: Also works, just renders it a bit differently

(Notice the capitalisation of Example)

Using `..code-block:: python`:

image

Using `Example::`:

image

yunzheng: The `Example::` does have the benefit that you don't have to indent it twice. Let me know what is preferred :)

Schamper: I prefer the first, because that capitalisation means that Sphinx has properly picked it up as an example block.

Schamper (approved):

dissect/target/helpers/fsutil.py
Schamper: I know, but I guess that's a bug in Sphinx?

PR #1000, opened 2025-01-22
Add container image support

JSCU-CNI:
This PR adds container image support to dissect in the form of a loader and filesystem. It implements the Docker image spec and OCI container image spec.

Example usage:
```bash
$ docker pull python:latest && docker image save python:latest -o image.tar
$ target-shell image.tar
debuerreotype:/$ ls
$fs$
bin
boot
dev
etc
home
lib
lib64
media
mnt
opt
proc
root
run
sbin
srv
sys
tmp
usr
var
debuerreotype:/$ python
In [1]: t.fs.mounts["/"]
Out[1]: <ContainerImageFilesystem path=/tmp/image.tar name=python:latest>

### CODE REVIEW

File: dissect/target/filesystems/containerimage.py
Comment (Schamper) [inline dissect/target/filesystems/containerimage.py:1]: I think this feature fits better when entirely implemented as a loader, rather than a loader and a separate filesystem. I don't think I see anything unique in the filesystem implementation that can't be handled in the loader, but correct me if I'm wrong.

Comment (Schamper) [reply dissect/target/filesystems/containerimage.py:1]: You can maintain those in the loader, which is accessible at `t._loader` I believe.

File: dissect/target/loaders/containerimage.py [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            self.config = json.loads(self.tar.path(self.manifest.get("Config")).read_text())
+        except Exception as e:
+            raise ValueError(f"Unable to read config inside docker image filesystem: {str(e)}")
+
+    @staticmethod
+    def detect(path: TargetPath) -> bool:
+        return (
+            TarLoader.detect(path)
+            and (root := set(map(str, TarFilesystem(path.open("rb")).path("/").iterdir())))
```
Schamper, line 68: Is there not a more efficient way to do this? This will be a very costly check on large tar files, which may not even be compatible with this loader.

This is a similar problem that we're dealing with in #935.

Schamper (changes requested):

dissect/target/loaders/containerimage.py
Schamper: I wanted to try an idea I had myself but unfortunately I got sick (and will have some days off after that), so perhaps someone on your side can shoot at this. I wrote down a little bit of this idea in #935 already:

> "Ideally" (not really, since it's only for performance and the code path itself will be more confusing) it's a subpath in the zip and tar loaders. So you piggy back on the detection logic of those, and upon mapping you do a fast check to see if you need to divert logic to Acquire logic (which could exist in loaders/acquire.py.
> 
> At least, that's the "best" idea I can come up with right now.

https://github.com/fox-it/dissect.target/pull/935/files#r1932313219

Basically I envisioned this loader (and the ones affected in #935) to not really work as independent loaders in isolation, but rather be an "extension" of `TarLoader` and `ZipLoader` in the sense that the code flow would look more like this:

1) `TarLoader.detect() -> True`
2) `TarLoader.map()`
3) In `TarLoader.map()`, the tarfile is loaded _once_ (!), either directly or by means of the `TarFilesystem`. 
4) A quick check here should determine what kind of tar file we're dealing with, so maybe the first few members of `getmembers()` or a few `exists` check on the `TarFilesystem`
5) Actual mapping logic is dispatched to the "subloader" if needed, or handled in the `TarLoader` in case it's just a regular tar file

Same logic but with zip files for the `ZipLoader`.

I don't think we should overcomplicate these things, but I do think the looming performance hit of this PR and #935 are big enough that we should come up with an alternative solution. I'm perfectly comfortable with some static imports and checks in `TarLoader` and `ZipLoader` respectively, as long as the end result is easy enough to understand and well documented.

dissect/target/filesystems/tar.py
```diff
@@ -33,15 +33,20 @@ class TarFilesystem(Filesystem):
     def __init__(
         self,
         fh: BinaryIO,
-        base: Optional[str] = None,
-        tarinfo: Optional[tarfile.TarInfo] = None,
+        base: str | None = None,
+        tarinfo: tarfile.TarInfo | None = None,
+        _tarfile: tarfile.TarFile | None = None,
```
Schamper, line 38:
```suggestion
        tarfile: tarfile.TarFile | None = None,
```
Don't specifically see a reason to prefix this with an underscore. Can you also add a docstring to this class that explains the arguments?

dissect/target/filesystems/tar.py
```diff
@@ -33,15 +33,20 @@ class TarFilesystem(Filesystem):
     def __init__(
         self,
         fh: BinaryIO,
-        base: Optional[str] = None,
-        tarinfo: Optional[tarfile.TarInfo] = None,
+        base: str | None = None,
```
Schamper, line 36:
```suggestion
        base: str | None = None,
        *,
```
Maybe to make the arguments a bit more explicit.

dissect/target/loaders/tar.py
```diff
@@ -34,14 +34,94 @@ def __init__(self, path: Path | str, **kwargs):
...
+
+    def map(self, target: target.Target) -> None:
+        for candidate in self.SUB_LOADERS:
+            if candidate.detect(self.tar):
+                self.subloader = candidate(self.tar)
+                self.subloader.map(target)
+                break
+
+
+TAR_EXT_COMP = (
```
Schamper, line 62: Can you move these constants to the top, above the `TarLoader` class?

dissect/target/loaders/tar.py
```diff
@@ -34,14 +34,94 @@ def __init__(self, path: Path | str, **kwargs):
...
+
+def is_tar_magic(path: Path, magics: tuple) -> bool:
+    with path.open("rb") as fh:
+        headers = [fh.read(6)]
+        fh.seek(257)
+        headers.append(fh.read(8))
+        for header in headers:
+            for magic in magics:
+                if len(magic) > len(header):
+                    continue
```
Schamper, line 93:
```suggestion
```
The check below it should fail anyway if this is true.

dissect/target/loader.py
```diff
@@ -101,6 +101,26 @@ def map(self, target: Target) -> None:
         raise NotImplementedError()
 
 
+class SubLoader:
```
Schamper, line 104:
```suggestion
T = TypeVar("T")


class SubLoader(Generic[T]):
```

dissect/target/loader.py
```diff
@@ -101,6 +101,26 @@ def map(self, target: Target) -> None:
...
+    """A base class for loading arbitary data and coupling it to a :class:`Target <dissect.target.target.Target>`.
+
+    Should not be called like a regular :class:`Loader`. For examples see :class:`TarLoader` and :class:`TarSubLoader` implementations.
+    """
+
+    def __init__(self, *args, **kwargs):
+        pass
+
```
Schamper, line 112:
```suggestion
```

dissect/target/loader.py
```diff
@@ -101,6 +101,26 @@ def map(self, target: Target) -> None:
...
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}>"
+
+    @staticmethod
+    def detect(*args) -> bool:
```
Schamper, line 117:
```suggestion
    def detect(value: T) -> bool:
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +34,94 @@ def __init__(self, path: Path | str, **kwargs):
                 "Consider uncompressing the archive before passing the tar file to Dissect."
             )
 
-        self.tar = tarfile.open(fileobj=path.open("rb"))
+        from dissect.target.loaders.containerimage import ContainerImageTarSubLoader
+
+        self.SUB_LOADERS = [
+            ContainerImageTarSubLoader,
+            GenericTarSubLoader,  # should be last
+        ]
```
Schamper, line 42: Maybe move this to a 
```python
__subloaders__ = [
    ContainerImageTarSubLoader,
    GenericTarSubLoader,
]
```
In the class? Can probably use `import_lazy` to avoid the circular import.

dissect/target/loaders/tar.py
```diff
@@ -34,14 +34,94 @@ def __init__(self, path: Path | str, **kwargs):
...
+
+        self.fh = path.open("rb")
+        self.tar = tarfile.open(mode="r:*", fileobj=self.fh)
```
Schamper, line 45:
```suggestion
        self.tar = tarfile.open(mode="r:*", fileobj=self.fh)
        
        self.subloader = None
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +34,94 @@ def __init__(self, path: Path | str, **kwargs):
...
+                if len(magic) > len(header):
+                    continue
+                if header[: len(magic)] == magic:
+                    return True
+    return False
+
 
-    def is_compressed(self, path: Path | str) -> bool:
-        return str(path).lower().endswith((".tar.gz", ".tgz"))
+class TarSubLoader(SubLoader):
```
Schamper, line 99:
```suggestion
class TarSubLoader(SubLoader[tarfile.TarFile]):
```

dissect/target/loaders/containerimage.py
```diff
@@ -0,0 +1,80 @@
...
+            raise ValueError(f"Unable to read manifest.json inside docker image filesystem: {str(e)}")
+
+        try:
+            self.config = json.loads(self.tarfs.path(self.manifest.get("Config")).read_text())
+        except Exception as e:
+            raise ValueError(f"Unable to read config inside docker image filesystem: {str(e)}")
+
+    @staticmethod
+    def detect(tar: tarfile.TarFile) -> bool:
+        # NOTE: tar.getmembers() / tar.getnames() is still inefficient for large tarballs, how should be prevent this?
```
Schamper, line 64: At least `getmembers()` is cached, so it'll get reused by the `GenericTarSubLoader`. Maybe it's a bit more efficient to do a loop over `getmembers()` and break out once you have seen enough filenames?

Schamper (changes requested):

dissect/target/filesystems/tar.py
Schamper: Can change that into `import tarfile as tf` or something similar?

dissect/target/loader.py
```diff
@@ -3,7 +3,7 @@
 import logging
 import urllib
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterator, Optional, Union
+from typing import TYPE_CHECKING, Generic, Iterator, Optional, TypeVar, Union
```
Schamper, line 6: Would you mind getting rid of `Optional` and `Union` in this file?

dissect/target/loader.py
```diff
@@ -101,6 +101,27 @@ def map(self, target: Target) -> None:
...
+    Should not be called like a regular :class:`Loader`. For examples see :class:`TarLoader`
+    and :class:`TarSubLoader` implementations.
+    """
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}>"
+
+    @staticmethod
+    def detect(value: T) -> bool:
+        raise NotImplementedError()
```
Schamper, line 119:
```suggestion
        raise NotImplementedError
```

dissect/target/loader.py
```diff
@@ -101,6 +101,27 @@ def map(self, target: Target) -> None:
...
+
+    def map(self, target: Target) -> None:
+        raise NotImplementedError()
```
Schamper, line 122:
```suggestion
        raise NotImplementedError
```

dissect/target/loaders/containerimage.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.tarfs = None
+        self.manifest = None
+        self.name = None
+        self.config = None
+
+        try:
+            self.tarfs = TarFilesystem(None, tarfile=tar)
+        except Exception as e:
+            raise ValueError(f"Unable to open {str(tar)} as TarFilesystem: {str(e)}")
```
Schamper, line 50:
```suggestion
            raise ValueError(f"Unable to open {str(tar)} as TarFilesystem: {str(e)}") from e
```

dissect/target/loaders/containerimage.py
```diff
@@ -0,0 +1,79 @@
...
+        try:
+            self.manifest = json.loads(self.tarfs.path("/manifest.json").read_text())[0]
+            self.name = self.manifest.get("RepoTags", [None])[0]
+        except Exception as e:
+            raise ValueError(f"Unable to read manifest.json inside docker image filesystem: {str(e)}")
```
Schamper, line 55:
```suggestion
            raise ValueError(f"Unable to read manifest.json inside docker image filesystem: {str(e)}") from e
```

dissect/target/loaders/containerimage.py
```diff
@@ -0,0 +1,79 @@
...
+
+        try:
+            self.config = json.loads(self.tarfs.path(self.manifest.get("Config")).read_text())
+        except Exception as e:
+            raise ValueError(f"Unable to read config inside docker image filesystem: {str(e)}")
```
Schamper, line 60:
```suggestion
            raise ValueError(f"Unable to read config inside docker image filesystem: {str(e)}") from e
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +66,66 @@ def __init__(self, path: Path | str, **kwargs):
...
+        return False
+
+    with path.open("rb") as fh:
+        headers = [fh.read(6)]
+        fh.seek(257)
+        headers.append(fh.read(8))
+        for header in headers:
+            for magic in magics:
+                if header[: len(magic)] == magic:
+                    return True
```
Schamper, line 99:
```suggestion
            if header.startswith(magics):
                return True
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +66,66 @@ def __init__(self, path: Path | str, **kwargs):
...
+
+    def map(self, target: target.Target) -> None:
+        for candidate in self.__subloaders__:
+            if candidate.detect(self.tar):
+                self.subloader = candidate(self.tar)
+                self.subloader.map(target)
+                break
+
+
+def is_tar_magic(path: Path, magics: tuple) -> bool:
```
Schamper, line 88:
```suggestion
def is_tar_magic(path: Path, magics: Iterable[bytes]) -> bool:
```
I think.

dissect/target/loaders/tar.py
```diff
@@ -34,14 +66,66 @@ def __init__(self, path: Path | str, **kwargs):
...
+    """Tar implementation of a :class:`SubLoader`."""
+
+    def __init__(self, tar: tf.TarFile, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.tar = tar
+
+    @staticmethod
+    def detect(tarfile: tf.TarFile) -> bool:
+        """Only to be called internally by :class:`TarLoader`."""
+        raise NotImplementedError()
```
Schamper, line 113:
```suggestion
        raise NotImplementedError
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +66,66 @@ def __init__(self, path: Path | str, **kwargs):
...
+
+    def map(self, target: target.Target) -> None:
+        """Only to be called internally by :class:`TarLoader`."""
+        raise NotImplementedError()
```
Schamper, line 117:
```suggestion
        raise NotImplementedError
```

dissect/target/loaders/tar.py
```diff
@@ -34,14 +66,66 @@ def __init__(self, path: Path | str, **kwargs):
...
+        self.tar = tf.open(mode="r:*", fileobj=self.fh)
+        self.subloader = None
 
     @staticmethod
     def detect(path: Path) -> bool:
-        return path.name.lower().endswith((".tar", ".tar.gz", ".tgz"))
+        return path.name.lower().endswith(TAR_EXT + TAR_EXT_COMP) or is_tar_magic(path, TAR_MAGIC + TAR_MAGIC_COMP)
+
+    def is_compressed(self, path: Path) -> bool:
+        return path.name.lower().endswith(TAR_EXT_COMP) or is_tar_magic(path, TAR_MAGIC_COMP)
```
Schamper, line 78: Can you move this to a normal function too? Like `is_tar_magic`.

dissect/target/loader.py
```diff
@@ -101,6 +101,27 @@ def map(self, target: Target) -> None:
...
+T = TypeVar("T")
+
+
+class SubLoader(Generic[T]):
+    """A base class for loading arbitary data and coupling it to a :class:`Target <dissect.target.target.Target>`.
+
+    Should not be called like a regular :class:`Loader`. For examples see :class:`TarLoader`
+    and :class:`TarSubLoader` implementations.
+    """
+
```
Schamper, line 113:
```suggestion

    def __init__(self, value: T):
        pass

```

Schamper (changes requested):

dissect/target/loaders/tar.py
```diff
@@ -2,46 +2,131 @@
...
+    return path.name.lower().endswith(TAR_EXT_COMP) or is_tar_magic(path, TAR_MAGIC_COMP)
 
-    def is_compressed(self, path: Path | str) -> bool:
-        return str(path).lower().endswith((".tar.gz", ".tgz"))
+
+class TarSubLoader(SubLoader[tf.TarFile]):
+    """Tar implementation of a :class:`SubLoader`."""
+
+    def __init__(self, tar: tf.TarFile, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
Schamper, line 108:
```suggestion
        super().__init__(tar, *args, **kwargs)
```
Should fix the error I think.

Schamper (changes requested):

Schamper (approved):

PR #1001, opened 2025-01-22
Normalize UTC timezone names for unix and windows targets

JSCU-CNI:
This PR fixes the timezone name for unix systems configured to use UTC and adds tests.


dissect/target/helpers/data/windowsZones.xml
```diff
@@ -266,7 +266,7 @@ Source:     https://github.com/unicode-org/cldr/blob/main/common/supplemental/wi
 			<mapZone other="Cape Verde Standard Time" territory="ZZ" type="Etc/GMT+1"/>
 
 			<!-- (UTC) Coordinated Universal Time -->
-			<mapZone other="UTC" territory="001" type="Etc/UTC"/>
+			<mapZone other="UTC" territory="001" type="UTC"/> <!--// Changed to UTC to be consistent across operating systems. //-->
```
Schamper, line 269: Probably better to handle this in code rather than modify the file? If this file requires periodic updates (according to the comment in the header) it's easy to forget this change.

Schamper (changes requested):

Schamper (approved):

PR #1007, opened 2025-01-29
Add retrieval of runtime information to plugin descriptors

Schamper:
Also add an initial example function for finding functions by record field types.

Depends on #763

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -82,6 +82,10 @@ class PluginDescriptor:
     functions: list[str]
     exports: list[str]
 
+    @property
+    def cls(self) -> type[Plugin]:
+        return load(self)
+
 
 @dataclass(frozen=True, eq=True)
 class FunctionDescriptor:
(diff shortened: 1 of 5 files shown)


Schamper:
That's an attribute that's known at registration time already, so I think we can just add that to the `FunctionDescriptor` dataclass directly.

Schamper:
Can you make a separate PR for that patch? I don't think it's dependent on anything in this PR.

Horofic (dismissed):

dissect/target/plugin.py
```diff
@@ -841,6 +866,43 @@ def find_functions(
     return result, invalid_functions
 
 
+def find_functions_by_record_field_type(
```
Horofic, line 869: After some testing using `target-shell`'s python mode on a fresh Linux virtual machine it seems that something along the lines of  `find_functions_by_record_field_type("path", target=t, compatibility=True)` slows things down considerably

The `seen` variable on line 954 in the `_filter_compatible()` function never gets populated. Populating this locally seems to fix the slow down issues.

Horofic (changes requested):

PR #1008, opened 2025-01-30
Improve compatibility checks on various plugins

JSCU-CNI:


dissect/target/plugins/filesystem/ntfs/usnjrnl.py
```diff
@@ -27,7 +28,8 @@ class UsnjrnlPlugin(Plugin):
     """NFTS UsnJrnl plugin."""
 
     def check_compatible(self) -> None:
-        pass
+        if not any([fs for fs in self.target.filesystems if fs.__type__ == "ntfs"]):
```
Schamper, line 31:
```suggestion
        if not any(fs for fs in self.target.filesystems if fs.__type__ == "ntfs"):
```

Schamper (changes requested):

Schamper (approved):

PR #1012, opened 2025-02-04
Add loader for VBK files

Matthijsy:
This PR adds support for handling VBK files. Currently we have to extract them using `vbk-extract` from `dissect.archive`, but by adding a loader for them we no longer have to go through this. This also gives the benefit is being able to extract/target specific hosts within the VBK.

For now I have only added support for child items which are vmx files, or direct backups of hosts (which in my case ended up in a disk with a certain identifier in the VBK). I coudn't figure a nice way to find the correct child items via a configuration file or something. If anyone thinks of a better way to identify the childs within a VBK I would be happy to implement!


dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+
+        def map_entries(directory, parent_path) -> None:
+            for entry in directory.iterdir():
+                entry_path = parent_path.joinpath(entry.name)
+                if entry.is_dir():
+                    map_entries(entry, entry_path)
+                else:
+                    self._fs.map_file_entry(entry_path.__str__(), VBKFilesystemEntry(self, str(entry_path), entry))
+
+        map_entries(self.vbk.get("/"), Path("/"))
```
Schamper, line 47: The VBK implementation already has it's own filesystem-like methods and path resolution, no need to walk the entire VBK and map everything into a virtual filesystem, just call the relevant methods on the VBK implementation.

You can look at other "real" filesystem implementations in `dissect.target` for inspiration.

dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+from dissect.target.exceptions import (
+    NotASymlinkError,
+)
+from dissect.target.filesystem import (
+    Filesystem,
+    FilesystemEntry,
+    VirtualFilesystem, )
+from dissect.target.helpers import fsutil
+
+log = logging.getLogger(__name__)
```
Schamper, line 19:
```suggestion
```
Unused.

dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+    @staticmethod
+    def _detect(fh: BinaryIO) -> bool:
+        raise TypeError("Detect is not allowed on VBKFilesystem class")
+
+    def get(self, path: str, relentry: FilesystemEntry = None) -> FilesystemEntry:
+        """Returns a VBKFilesystemEntry object corresponding to the given path."""
+        return self._fs.get(path, relentry=relentry)
+
+
+class VBKFilesystemEntry(FilesystemEntry):
```
Schamper, line 58:
```suggestion
class VbkFilesystemEntry(FilesystemEntry):
```

dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+                if entry.is_dir():
+                    map_entries(entry, entry_path)
+                else:
+                    self._fs.map_file_entry(entry_path.__str__(), VBKFilesystemEntry(self, str(entry_path), entry))
+
+        map_entries(self.vbk.get("/"), Path("/"))
+
+    @staticmethod
+    def _detect(fh: BinaryIO) -> bool:
+        raise TypeError("Detect is not allowed on VBKFilesystem class")
```
Schamper, line 51: Why not? The loader may do the heavy lifting most of the time, but denying this ability is needlessly limiting.

dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+    FilesystemEntry,
+    VirtualFilesystem, )
+from dissect.target.helpers import fsutil
+
+log = logging.getLogger(__name__)
+
+
+class VBKFilesystem(Filesystem):
+    """Filesystem implementation for VBK files.
+    """
```
Schamper, line 24:
```suggestion
    """Filesystem implementation for VBK files."""
```

dissect/target/filesystems/vbk.py
```diff
@@ -0,0 +1,105 @@
...
+
+    __type__ = "vbk"
+
+    def __init__(
+            self,
+            fh: BinaryIO,
+            *args,
+            **kwargs,
+    ):
```
Schamper, line 33:
```suggestion
    def __init__(self, fh: BinaryIO, *args, **kwargs):
```

dissect/target/loaders/vbk.py
```diff
@@ -0,0 +1,19 @@
+from dissect.target.filesystems.vbk import VBKFilesystem
+from dissect.target.loader import Loader
+
+
+class VBKLoader(Loader):
+    """Load Veaam Backup (VBK) files.
+    """
```
Schamper, line 7:
```suggestion
class VbkLoader(Loader):
    """Load Veaam Backup (VBK) files."""
```

dissect/target/loaders/vbk.py
Schamper, line 1: Can you add type hints?

dissect/target/loaders/vbk.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__(path, **kwargs)
+        self.path = path
+
+    @staticmethod
+    def detect(path):
+        return path.suffix.lower() == ".vbk"
+
+    def map(self, target):
+        fs = VBKFilesystem(self.path.open("rb"))
+        target.filesystems.add(fs)
```
Schamper, line 19: I don't think a VBK on itself should be a "valid" target, as it's more of a wrapper around one or more actual targets. Right now you're kind of abusing the child subsystem to achieve this, but ideally this is not needed and you can just target the VBK directly and get all the targets out that way.

Currently, I think the best way would be to utilize the `find_all` functionality of a loader, which allows you to divide a single "target path" into multiple:

https://github.com/fox-it/dissect.target/blob/a0a8c3e9207e46bad8b99bca65c60f2c869541c9/dissect/target/loader.py#L80
https://github.com/fox-it/dissect.target/blob/a0a8c3e9207e46bad8b99bca65c60f2c869541c9/dissect/target/target.py#L331

I think a nice way to handle this would be to generate file paths in `VbkLoader.find_all` that look like `/path/to/file.vbk/path/to/vm.vmx`. This is similar to how you can execute Python scripts in a ZIP file (`python file.zip/script.py`), so it's a familiar paradigm in the Python world. Then in the `__init__` and `map` of the `VbkLoader` you process that.

Another option would be to extend the loader subsystem with a way to indicate that you can map multiple targets from a single path, but that would be a more complicated change.

Schamper:
@Matthijsy do you have a small VBK test file that contains a `.vmx` for me to experiment with?

Schamper (approved):

Issue #920, opened 2024-10-29
Re-use arguments across plugins and target-query


Schamper:
Currently any argument that's used by a plugin will be unavailable for other plugins, as well as arguments used by `target-query` itself. This was observed in https://github.com/fox-it/dissect.target/pull/841.

Ideally there's no clashing and you can just freely reuse arguments, which is much more in line with expected behaviour.

PR #1022, opened 2025-02-13
Nfs filesystem

twiggler:
Implement Nfs filesystem in terms of experimental Nfsv3 Client.
- Automount the filesystem using mount information in `fstab`, but only when target is local.
- Bruteforce authentication by trying all users found on the target.

Also refactored client and serialization code.


dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+
+        current_handle = self._root_handle
+        for segment in path.split("/"):
+            result = self._client.lookup(segment, current_handle)
+            current_handle = result.object
+
+        return NfsFilesystemEntry(self, path, result.object, result.obj_attributes)
+
+
+class NfsFilesystemEntry(FilesystemEntry):
```
Schamper, line 59:
```suggestion
class NfsFilesystemEntry(FilesystemEntry):
    fs: NfsFilesystem
    entry: FileHandle3
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+    def __init__(
+        self, fs: NfsFilesystem, path: str, file_handle: FileHandle3, attributes: FileAttributes3 | None = None
+    ):
+        super().__init__(fs, path, file_handle)
+        self._fs = fs
+        self._file_handle = file_handle
```
Schamper, line 65:
```suggestion
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        self._backing_attributes = attributes
+
+    @property
+    def _attributes(self) -> FileAttributes3:
+        if self._backing_attributes is None:
+            self._backing_attributes = self._fs._client.getattr(self._file_handle)
```
Schamper, line 71:
```suggestion
            self._backing_attributes = self.fs._client.getattr(self._file_handle)
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        if follow_symlinks and self.is_symlink():
+            return self.readlink_ext().is_dir()
+
+        return self._attributes.type == FileType3.DIR
+
+    def is_symlink(self) -> bool:
+        return self._attributes.type == FileType3.LNK
+
+    def readlink(self) -> str:
+        return self._fs._client.readlink(self._file_handle)
```
Schamper, line 96:
```suggestion
        return self.fs._client.readlink(self._file_handle)
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def readlink_ext(self) -> NfsFilesystemEntry:
+        target = self._fs._client.readlink(self._file_handle)
```
Schamper, line 99:
```suggestion
        target = self.fs._client.readlink(self._file_handle)
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        return self.get(target)
+
+    def _iterdir(self) -> Iterator[EntryPlus3]:
+        if not self.is_dir():
+            raise NotADirectoryError(self.path)
+
+        yield from self._fs._client.readdir(self._file_handle).entries
```
Schamper, line 106:
```suggestion
        yield from self.fs._client.readdir(self._file_handle).entries
```

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+    def iterdir(self) -> Iterator[str]:
+        for entry in self._iterdir():
+            yield entry.name
+
+    def scandir(self) -> Iterator[FilesystemEntry]:
+        for entry in self._iterdir():
+            yield NfsFilesystemEntry(self.fs, entry.name, entry.handle, entry.attributes)
+
+    def open(self) -> NfsStream:
+        return NfsStream(self._fs._client, self._file_handle)
```
Schamper, line 117:
```suggestion
        return NfsStream(self.fs._client, self._file_handle)
```

dissect/target/helpers/nfs/mount_client.py
```diff
@@ -0,0 +1,56 @@
...
+class Client(AbstractContextManager):
+    DIR_COUNT = 4096  # See https://datatracker.ietf.org/doc/html/rfc1813#section-3.3.17
+    MAX_COUNT = 32768
+    READ_CHUNK_SIZE = 1024 * 1024
+
+    def __init__(self, rpc_client: SunRpcAbstractClient):
+        self._rpc_client = rpc_client
+
+    def __enter__(self) -> Client:
+        """Return `self` upon entering the runtime context."""
```
Schamper, line 36:
```suggestion
        """Return ``self`` upon entering the runtime context."""
```

dissect/target/helpers/nfs/nfs_client.py
Schamper, line 1: Maybe make a directory `dissect/target/helpers/nfs/client` and put a `nfs.py` and `mount.py` in there?

dissect/target/helpers/nfs/nfs_client.py
```diff
@@ -0,0 +1,164 @@
...
+class Client(AbstractContextManager):
+    DIR_COUNT = 4096  # See https://datatracker.ietf.org/doc/html/rfc1813#section-3.3.17
+    MAX_COUNT = 32768
+    READ_CHUNK_SIZE = 1024 * 1024
+
+    def __init__(self, rpc_client: SunRpcAbstractClient):
+        self._rpc_client = rpc_client
+
+    def __enter__(self) -> Client:
+        """Return `self` upon entering the runtime context."""
```
Schamper, line 72:
```suggestion
        """Return ``self`` upon entering the runtime context."""
```

dissect/target/helpers/nfs/serializer.py
```diff
@@ -130,14 +130,56 @@ def serialize(self, args: Read3args) -> bytes:
...
+
+
+# RdJ: Consider implementing in terms of a monadic bind, using generators
+class ResultDeserializer(XdrDeserializer[Union[ResultType, Nfs3Stat]]):
+    """A higher order deserializer that returns a result or an NFS status"""
+
+    def __init__(self, deserializer: XdrDeserializer[ResultType]):
+        self._deserializer = deserializer
+
+    def deserialize(self, payload: io.BytesIO) -> Union[ResultType, Nfs3Stat]:
```
Schamper, line 180:
```suggestion
    def deserialize(self, payload: io.BytesIO) -> ResultType | Nfs3Stat:
```

dissect/target/plugins/os/unix/_os.py
```diff
@@ -226,7 +249,12 @@ def _parse_hosts_string(self, paths: list[str] | None = None) -> dict[str, str]:
     def _add_mounts(self) -> None:
         fstab = self.target.fs.path("/etc/fstab")
 
-        for dev_id, volume_name, mount_point, _, options in parse_fstab(fstab, self.target.log):
+        for dev_id, volume_name, mount_point, fs_type, options in parse_fstab(fstab, self.target.log):
+            # Mount nfs, but only when target is local
+            if fs_type == "nfs" and str(self.target.path) == "local":
```
Schamper, line 254: Maybe nicer to do a type check on `self.target._loader` for `LocalLoader`?

dissect/target/plugins/os/unix/_os.py
```diff
@@ -265,6 +293,49 @@ def _add_mounts(self) -> None:
...
+            nfs_client = NfsClient.connect(address, nfs_port, auth_null(), FreePrivilegedPort)
+
+            # Try all users to see if we can access the share
+            for user in self.users():
+                if user.uid is None or user.gid is None:
+                    continue
+                auth = auth_unix("machine", user.uid, user.gid, [])
+                nfs_client.rebindAuth(auth)
+                try:
+                    self.target.log.info(f"Trying to read NFS share with uid {user.uid} and gid {user.gid}")
```
Schamper, line 324: Can you change all f-string log format messages to use `%s` style?

dissect/target/filesystems/nfs.py
twiggler: Can't, `FileSystemEntry` upcasts `fs` to `Filesystem`.

Schamper: What do you mean with "can't"? As in that a real type checker isn't satisfied with this? This works fine in e.g. `SmbFilesystem`.

twiggler: You're right, I misread that as being class variables, but they are actually typehints on instance variables
(https://peps.python.org/pep-0526/#class-and-instance-variable-annotations)

twiggler: (I still don´t get why `fs` and `entry` are defined on the `FilesystemEntry` though. `entry` is not used at all in `FilesystemEntry`, and `fs` primarily for the `alt_sepaparator`)

Schamper: It's for interactive usage in IPython shells, it gives a user access to the raw underlying filesystem object under a standardized attribute name.

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+            return stat.S_IFREG
+        elif type == FileType3.LNK:
+            return stat.S_IFLNK
+        else:
+            return 0o000000
+
+
+class NfsStream(AlignedStream):
+    def __init__(self, client: Client, file_handle: FileHandle3):
+        super().__init__(None, Client.READ_CHUNK_SIZE)
```
Schamper, line 162: Maybe nice to initialize the size with the size from the file attributes?

tests/filesystems/test_nfs.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.fixture
+def mock_nfs_client() -> Iterator[MagicMock]:
+    with patch("dissect.target.helpers.nfs.client.nfs.Client", autospec=True) as mock_client:
+        yield mock_client
+
+
+@pytest.fixture
+def nfs_filesystem(mock_nfs_client) -> NfsFilesystem:
```
Schamper, line 25: Can you add a type hint

tests/filesystems/test_nfs.py
```diff
@@ -0,0 +1,143 @@
...
+    client_factory = MagicMock(return_value=mock_nfs_client)
+    root_handle = FileHandle3(opaque=b"root_handle")
+    return NfsFilesystem(client_factory, root_handle)
+
+
+@pytest.fixture
+def nfs_filesystem_entry(nfs_filesystem) -> NfsFilesystemEntry:
```
Schamper, line 32: Can you add a type hint

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        hostname: str,
+        port: int,
+        auth: AuthScheme[Credentials, Verifier],
+        local_port: int | FreePrivilegedPortType = 0,
+    ) -> Client:
+        rpc_client = SunRpcClient.connect(hostname, port, auth, local_port)
+        return Client(rpc_client)
+
+    def rebind_auth(self, auth: AuthScheme[Credentials, Verifier]) -> None:
+        """Change the authentication scheme of the underlying sunrpc client"""
```
Schamper, line 91:
```suggestion
        """Change the authentication scheme of the underlying sunrpc client."""
```

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+
+        self._rpc_client = self._rpc_client.rebind_auth(auth)
+
+    def readdir(self, dir: FileHandle3) -> ReadDirResult | Nfs3Stat:
+        """Read the contents of a directory, including file attributes"""
```
Schamper, line 96:
```suggestion
        """Read the contents of a directory, including file attributes."""
```

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+
+            entries += result.entries
+            if result.eof or len(result.entries) == 0:
+                return ReadDirResult(result.dir_attributes, entries)
+
+            cookie = result.entries[-1].cookie
+            cookieverf = result.cookieverf
+
+    def readfile(self, handle: FileHandle3, offset: int = 0, size: int = -1) -> Iterator[bytes]:
+        """Read a file by its file handle"""
```
Schamper, line 119:
```suggestion
        """Read a file by its file handle."""
```

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        bytes_left = size
+        read_deserializer = NfsResultDeserializer(Read3ResultDeserializer())
+        while size == -1 or bytes_left > 0:
+            count = self.READ_CHUNK_SIZE if size == -1 else min(self.READ_CHUNK_SIZE, size)
+            params = Read3args(handle, offset, count)
```
Schamper, line 124: Should this not be `Read3Args`?

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+            if isinstance(result, Nfs3Stat):
+                raise NfsError("Failed to read file", result)
+            yield result.data
+            if result.eof:
+                return
+            offset += result.count
+            bytes_left -= result.count
+
+    def lookup(self, name: str, parent: FileHandle3) -> Lookup3resok:
+        """Lookup a file by name in a directory"""
```
Schamper, line 135:
```suggestion
        """Lookup a file by name in a directory."""
```

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+        args = DirOpArgs3(parent, name)
+        lookup_deserializer = NfsResultDeserializer(Lookup3ResultDeserializer())
+        result = self._rpc_client.call(LookupProc, args, DirOpArgs3Serializer(), lookup_deserializer)
+        if isinstance(result, Nfs3Stat):
+            raise NfsError("Failed to lookup file", result)
+
+        return result
+
+    def getattr(self, handle: FileHandle3) -> FileAttributes3:
+        """Get the attributes of a file by its file handle"""
```
Schamper, line 146:
```suggestion
        """Get the attributes of a file by its file handle."""
```

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+
+        attr_deserializer = NfsResultDeserializer(FileAttributesSerializer())
+        result = self._rpc_client.call(GetAttrProc, handle.opaque, OpaqueVarLengthSerializer(), attr_deserializer)
+        if isinstance(result, Nfs3Stat):
+            raise NfsError("Failed to get attributes", result)
+
+        return result
+
+    def readlink(self, handle: FileHandle3) -> str:
+        """Read the target of a symlink by its file handle"""
```
Schamper, line 156:
```suggestion
        """Read the target of a symlink by its file handle."""
```

dissect/target/helpers/nfs/nfs3.py
```diff
@@ -167,3 +164,22 @@ class Read3resok:
...
+
+
+@dataclass
+class DirOpArgs3:
+    handle: FileHandle3
+    filename: str
+
+
+@dataclass
+class Lookup3resok:
```
Schamper, line 176: Should anything in `resok` not be capitalized? I'm not entirely sure what `resok` means here.

dissect/target/helpers/nfs/nfs3.py
```diff
@@ -167,3 +164,22 @@ class Read3resok:
...
+    object: FileHandle3
+    obj_attributes: FileAttributes3 | None
+    dir_attributes: FileAttributes3 | None
+
+
+@dataclass
+class Readlink3resok:
```
Schamper, line 183: Same here.

dissect/target/plugins/os/unix/_os.py
```diff
@@ -266,6 +284,52 @@ def _add_mounts(self) -> None:
                     self.target.log.debug("Mounting %s (%s) at %s", fs, fs.volume, mount_point)
                     self.target.fs.mount(mount_point, fs)
 
+    def _add_nfs(self, address: str, exported_dir: str, mount_point: str) -> None:
```
Schamper, line 287: I'm wondering if it might be nicer to decouple all this initialization away from the `UnixPlugin`, perhaps into the `NfsFilesystem` itself (or that file). From a quick scan, it's mostly actions related to initializing the connection and setting up the `NfsFilesystem`, with the exception of the client factory.

My reasoning behind it is that, technically, Windows also supports NFS shares. Beside the client factory (much of which actually seems reusable too, only the providing of `uid`/`gid` is really specific to the `UnixPlugin`), I think pretty much all of this code could technically be reused there. Important to note that I'm not saying you should add NFS support for Windows in this PR, just that it might be beneficial to organize the code already in such a way that it's easier to support it.

I think something like a classmethod to initialize a `NfsFilesystem` from just a connection string (and auth information) would be pretty cool.

While thinking about this, eventually it'll also make sense to refactor the SMB code a bit to allow for more flexible initialization, perhaps also from just a connection string, since you can of course also have CIFS/SMB mounts on Unix.

Schamper (changes requested):

dissect/target/plugins/os/unix/_os.py
twiggler: Tried to implement; the factory method taking a  connection string can now be implemented in terms of the connect class method I added.

You can pass a `AuthScheme`, or a `auth_setter` which does disgusting things :).

What bothers me a bit is getting the root handle is eager because the  `mount ` method needs it, which seems to be a implicit coupling. I added a comment. We could add a `lazy_mount` at some point. 

In this case the laziness is required, because of the `users` method requiring a constructed os plugin. Technically we could rewrite users to output dataclasses, and then create a wrapper method which returns records, but that  doesn´t sound very appealing. Futher, I think it is generally speaking beneficial to only open a connection when required.

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,223 @@
...
+    """Filesystem implementation of a NFS share
+
+    The connection is lazily established to not waste resources.
+    Use the ``connect`` method to conveniently create a new instance.
+    """
+
+    __type__ = "nfs"
+
+    @classmethod
+    def connect(
```
Schamper, line 55: Can you move this below `__init__`?

dissect/target/filesystems/nfs.py
```diff
@@ -0,0 +1,223 @@
...
+    def detect(_: BinaryIO) -> bool:
+        raise TypeError("Detect is not allowed on a NfsFilesystem class")  # :/
+
+    def get(self, path: str) -> NfsFilesystemEntry:
+        path = fsutil.normalize(path, self.alt_separator).strip("/")
+
+        if not path:
+            return NfsFilesystemEntry(self, "/", self._root_handle)
+
+        current_handle = self._root_handle
```
Schamper, line 105: We use a `relentry` argument a lot in other filesystem implementations to save some performance on long relative path lookups.

dissect/target/helpers/nfs/client/mount.py
```diff
@@ -0,0 +1,56 @@
...
+from dissect.target.helpers.sunrpc.client import AbstractClient as SunRpcAbstractClient
+from dissect.target.helpers.sunrpc.client import Client as SunRpcClient
+from dissect.target.helpers.sunrpc.client import FreePrivilegedPortType, auth_null
+from dissect.target.helpers.sunrpc.serializer import StringSerializer
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+
+class MountError(Exception):
```
Schamper, line 17: Maybe make this a subclass of `dissect.target.exceptions.Error`.

dissect/target/helpers/nfs/client/nfs.py
```diff
@@ -0,0 +1,168 @@
...
+from dissect.target.helpers.sunrpc.serializer import OpaqueVarLengthSerializer
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+Credentials = TypeVar("Credentials")
+Verifier = TypeVar("Verifier")
+
+
+class NfsError(Exception):
```
Schamper, line 48: Same here.

dissect/target/helpers/nfs/serializer.py
```diff
@@ -119,25 +119,67 @@ def deserialize(self, payload: io.BytesIO) -> ReadDirPlusResult3:
...
+
+        return ReadlinkResult(attributes, target)
+
+
+ResultType = TypeVar("RT")
+
+
+# RdJ: Consider implementing in terms of a monadic bind, using generators
+class ResultDeserializer(XdrDeserializer[Union[ResultType, NfsStat]]):
+    """A higher order deserializer that returns a result or an NFS status"""
```
Schamper, line 175:
```suggestion
    """A higher order deserializer that returns a result or an NFS status."""
```

dissect/target/helpers/sunrpc/client.py
```diff
@@ -67,28 +76,94 @@ class IncompleteMessage(Exception):
...
+    def rebind_auth(self, auth: AuthScheme[ConCredentials, ConVerifier]) -> None:
+        pass
+
+
+class FreePrivilegedPortType:
+    pass
+
+
+"""Marker type indicating Client should search for a free privileged port"""
+FreePrivilegedPort = FreePrivilegedPortType()
```
Schamper, line 108: Is this marker really necessary? I think since `0` is generally accepted as serving a similar purpose (albeit not for privileged ports), within the context of this API + with a docstring describing it, it's equally suitable for this purpose.

Or is there something I'm not seeing? I think in the code there's a difference between `0` and `FreePrivilegedPort`, but I don't really see it clearly documented. In any case, I think I'm more in favor of a more "Pythonic" approach to this, which is to use a "magic value" instead of a special marker object. For example `-1`, or `None`, and then clearly document those values in the docstrings.

dissect/target/helpers/sunrpc/client.py
```diff
@@ -100,12 +175,26 @@ def close(self) -> None:
...
     def call(
         self,
         proc_desc: ProcedureDescriptor,
         params: Params,
-        params_serializer: XdrSerializer[Params],
-        result_deserializer: XdrDeserializer[Results],
+        params_serializer: Serializer[Params],
+        result_deserializer: Deserializer[Results],
     ) -> Results:
         """Synchronously call an RPC procedure and return the result"""
```
Schamper, line 199:
```suggestion
        """Synchronously call an RPC procedure and return the result."""
```

Schamper (changes requested):

dissect/target/helpers/sunrpc/client.py
Schamper: Maybe an `IntEnum` is better suited, which allows the "best of both worlds"? A named value, but also an integer value for those too lazy to import `FreePrivilegedPort` (e.g. when messing around in a quick script or REPL).

dissect/target/plugins/os/unix/_os.py
```diff
@@ -227,7 +234,12 @@ def _parse_hosts_string(self, paths: list[str] | None = None) -> dict[str, str]:
     def _add_mounts(self) -> None:
         fstab = self.target.fs.path("/etc/fstab")
 
-        for dev_id, volume_name, mount_point, _, options in parse_fstab(fstab, self.target.log):
+        for dev_id, volume_name, mount_point, fs_type, options in parse_fstab(fstab, self.target.log):
+            # Mount nfs, but only when target has been mapped by the `LocalLoader`
+            if fs_type == "nfs" and isinstance(self.target._loader, LocalLoader):
```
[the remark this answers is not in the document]
twiggler: Will discuss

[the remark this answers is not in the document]
twiggler: Discussed with the team. For now, we settled on disabling  nfs mounting  by default. It can be enabled by specifying the `enable-nfs` query parameter. We also show a warning when a nfs share is detected in fstab and nfs is not enabled.

I do wonder how the user can discover `enable-nfs`, `force-directory-fs`, and `fallback-to-directory-fs`. They do not seem to appear  as cli arguments in target-shell or target-query. Judging from `args_to_uri` you need to add them after a `-L` flag or something?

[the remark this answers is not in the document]
Schamper: I had originally wanted the loaders to adapt the `@arg` syntax that plugins also use, but it doesn't look like that made its way into every loader. Only the mqtt loader seems to use it now, but I don't know how "formalized" it is within `target-query`. With that, it could at least be documented in generated docstrings or `-h` output. Right now it would have to be explicitly documented.

dissect/target/plugins/os/unix/_os.py
```diff
@@ -227,7 +234,20 @@ def _parse_hosts_string(self, paths: list[str] | None = None) -> dict[str, str]:
...
+                if "enable-nfs" in self.target.path_query:
+                    self._add_nfs(dev_id, volume_name, mount_point)
+                else:
+                    log.warning(
+                        "NFS mount %s:%s at %s is disabled, to enable pass --enable-nfs to local loader",
```
Schamper, line 244: Maybe also document the URI method?

Schamper (changes requested):

dissect/target/plugins/os/unix/_os.py
```diff
@@ -227,7 +234,20 @@ def _parse_hosts_string(self, paths: list[str] | None = None) -> dict[str, str]:
...
 
-        for dev_id, volume_name, mount_point, _, options in parse_fstab(fstab, self.target.log):
+        for dev_id, volume_name, mount_point, fs_type, options in parse_fstab(fstab, self.target.log):
+            # Mount nfs, but only when target has been mapped by the `LocalLoader`
+            if fs_type == "nfs" and isinstance(self.target._loader, LocalLoader):
+                if "enable-nfs" in self.target.path_query:
+                    self._add_nfs(dev_id, volume_name, mount_point)
+                else:
+                    log.warning(
+                        "NFS mount %s:%s at %s is disabled. To enable, pass --enable-nfs to the local loader. Alternatively, add a query parameter to the target query string: local?force-directory-fs&enable-nfs",  # noqa: E501
```
Schamper, line 244: Don't think you need the `force-directory-fs`?

Schamper (changes requested):

dissect/target/plugins/os/unix/_os.py
twiggler: It's an example, I can remove it.

Schamper: I think an example like that should be in the docs or `-h`, but in the log message it should be clear what exactly is necessary for the logged-about feature to function.

twiggler: I will create a ticket / check if there is already one to streamline the passing of arguments to loaders.

Or perhaps @EinatFox can do that,

Schamper: Can tie into https://github.com/fox-it/dissect.target/issues/920 as well.

Schamper (approved):

PR #1023, opened 2025-02-13
Improve Linux detection

JSCU-CNI:
This PR aims to improve the detection of Linux-like filesystems in dissect.

On a side note, certain functionality in the `unix` space is actually `linux` specific (e.g. user parsing, fstab parsing, hostname parsing). Fixing this disambiguation and mixed shared functionality between `unix` and `linux` OS plugins is something for another time perhaps.

### DIFF
--- a/dissect/target/plugins/os/unix/_os.py
+++ b/dissect/target/plugins/os/unix/_os.py
@@ -48,7 +48,7 @@ class UnixPlugin(OSPlugin):
 
     @classmethod
     def detect(cls, target: Target) -> Filesystem | None:
-        """If the ``/var`` and ``/etc`` folders exist on a filesystem it is treated as a Unix-like filesystem."""
+        # If the ``/var`` and ``/etc`` folders exist on a filesystem it is treated as a Unix-like filesystem
         for fs in target.filesystems:
             if fs.exists("/var") and fs.exists("/etc"):
                 return fs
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/os/unix/_os.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, target: Target):
 
     @classmethod
     def detect(cls, target: Target) -> Filesystem | None:
+        """If the ``/var`` and ``/etc`` folders exist on a filesystem it is treated as a Unix-like filesystem."""
```
Schamper, line 51:
```suggestion
        # If the ``/var`` and ``/etc`` folders exist on a filesystem it is treated as a Unix-like filesystem
```

dissect/target/plugins/os/unix/linux/_os.py
```diff
@@ -25,11 +25,42 @@ def __init__(self, target: Target):
...
+            "/boot/vmlinux",
+            "/opt",  # Backwards compatibility for previous Linux detection.
+        }
+
         for fs in target.filesystems:
-            if (
-                (fs.exists("/var") and fs.exists("/etc") and fs.exists("/opt"))
-                or (fs.exists("/sys/module") or fs.exists("/proc/sys"))
-            ) and not (MacPlugin.detect(target) or WindowsPlugin.detect(target)):
+            # We explicitly exclude filesystems that look more like a MacOS or Windows sysvol.
```
Schamper, line 53:
```suggestion
            # We explicitly exclude filesystems that look more like a macOS or Windows sysvol.
```

Schamper (changes requested):

Schamper (approved):

PR #1025, opened 2025-02-13
Add proper datetime conversion for StartTime and EndTime in the sru plugin

Miauwkeru:

### DIFF
--- a/dissect/target/plugins/os/windows/sru.py
+++ b/dissect/target/plugins/os/windows/sru.py
@@ -413,11 +413,10 @@ class SRUPlugin(Plugin):
             for column, value in column_values:
                 new_value = value
                 if new_value and (transform := TRANSFORMS.get(column)):
-                    _transformed_value = transform(new_value)
-                    if isinstance(_transformed_value, str):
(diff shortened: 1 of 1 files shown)


dissect/target/plugins/os/windows/sru.py
```diff
@@ -401,7 +413,11 @@ def read_records(self, table_name: str, record_type: SRURecord) -> Iterator[SRUR
             for column, value in column_values:
                 new_value = value
                 if new_value and (transform := TRANSFORMS.get(column)):
-                    new_value = self.target.fs.path(transform(new_value))
+                    _transformed_value = transform(new_value)
+                    if isinstance(_transformed_value, str):
+                        new_value = self.target.fs.path(transform(new_value))
+                    else:
+                        new_value = _transformed_value
```
Schamper, line 420:
```suggestion
                    if isinstance((transformed_value := transform(new_value)), str):
                        new_value = self.target.fs.path(transformed_value)
                    else:
                        new_value = transformed_value
```

Schamper (changes requested):

PR #1031, opened 2025-02-14
Add additional profile parsing to Chromium browsers

### LINKED ISSUE #1030: Browser Profiles Are Not Parsed
During an incident, it was discovered that only the `Default` chromium browser profile is parsed. Browser related artefacts from other profiles are skipped. This means that potentially important browser artefacts from different browser profiles could go unnoticed during an investigation.

This issue could be fixed by adding a glob pattern for the other profiles.

qmadev:
Fixes #1030


Schamper (changes requested):
Could you add a unit test? I think you can just re-use the existing test data, but just map it to a different path.

Schamper (approved):

PR #1041, opened 2025-02-19
Add readline ignore sequences to PS1

### LINKED ISSUE #1003: target-shell prompt not rendering ANSI color codes
Sometimes the prompt of `target-shell` is not rendered properly, the ANSI codes are printed as text.

This looks to be happening when using `uv` for the python interpreter mostly. It's possible that this is caused by some libraries that are not included in the `uv` Python distributions (maybe readline?) but i'm not 100% sure.

How to reproduce with docker and uv:

```shell
$ docker run -it --rm ghcr.io/astral-sh/uv:debian-slim uv run -p 3.12 --with dissect.target[full] target-shell /

[1;32m698865f728cc[0m:[1;34m/[0m$
```

Image

yunzheng:
This adds RL_PROMPT_START_IGNORE and RL_PROMPT_END_IGNORE sequences to PS1.

Fixes #1003


yunzheng:
And tab completion still works :)

Schamper:
Can you add a unit test for tab completion 😉

yunzheng:
Added test using pexpect, thanks for the tip :)

I also found an issue when creating the test, the output of `ls` does not honor `NO_COLOR=1`:

image

notice that `localtime` and `network` is colored in `NO_COLOR=1` mode.

yunzheng:
Skipping the test on Windows due to missing pexpect. I'm not sure why the test fails on PyPy. Will look into this another time.

Schamper (changes requested):

tests/tools/test_shell.py
```diff
@@ -23,6 +26,13 @@
 from dissect.target.tools.shell import main as target_shell
 from tests._utils import absolute_path
 
+try:
+    import pexpect
```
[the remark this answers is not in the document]
yunzheng: it was already pulled in via the `ipython` dep, but it's probably better to add it explicitly.

Schamper (changes requested):
Can you resolve the conflict with main and fix the failing Windows tests?

yunzheng:
> Can you resolve the conflict with main and fix the failing Windows tests?

fixed failing Windows tests (skipIf windows), what conflict with main though?

dissect/target/tools/shell.py
```diff
@@ -76,6 +77,56 @@
...
+    BOLD_RED = "\033[1;31m"
+    BOLD_GREEN = "\033[1;32m"
+    BOLD_YELLOW = "\033[1;33m"
+    BOLD_BLUE = "\033[1;34m"
+    BOLD_MAGENTA = "\033[1;35m"
+    BOLD_CYAN = "\033[1;36m"
+    BOLD_WHITE = "\033[1;37m"
+
+    @classmethod
+    def as_dict(cls):
```
Schamper, line 121:
```suggestion
    def as_dict(cls) -> dict[str, str]:
```

dissect/target/tools/shell.py
```diff
@@ -76,6 +77,56 @@
     readline = None
 
 
+def readline_escape(s: str | dict[str, str]):
```
Schamper, line 80:
```suggestion
def readline_escape(s: str | dict[str, str]) -> str | dict[str, str]:
```

Schamper (changes requested):

Schamper (approved):

PR #1043, opened 2025-02-24
Add "alias" to FunctionDescriptor and "required" field to arguments

JSCU-CNI:
As discussed in #1007.

### DIFF
--- a/dissect/target/plugins/os/unix/cronjobs.py
+++ b/dissect/target/plugins/os/unix/cronjobs.py
@@ -1,12 +1,18 @@
 from __future__ import annotations
 
 import re
+from pathlib import Path
 from typing import Iterator
 
-from dissect.target.helpers.record import TargetRecordDescriptor
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.descriptor_extensions import UserRecordDescriptorExtension
+from dissect.target.helpers.record import (
+    TargetRecordDescriptor,
+    create_extended_descriptor,
+)
 from dissect.target.plugin import Plugin, export
 
-CronjobRecord = TargetRecordDescriptor(
+CronjobRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
     "unix/cronjob",
     [
         ("string", "minute"),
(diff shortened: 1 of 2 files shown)


dissect/target/plugin.py
```diff
@@ -528,6 +534,7 @@ def register(plugincls: type[Plugin]) -> None:
                         exported=exported,
                         internal=internal,
                         findable=plugincls.__findable__,
+                        alias=getattr(attr, "__alias__", False),
```
Schamper, line 537:
```suggestion
                        alias=name in getattr(attr, "__aliases__", []),
```
I think we can do this without introducing another attribute on the function itself.

dissect/target/plugin.py
```diff
@@ -552,6 +559,7 @@ def register(plugincls: type[Plugin]) -> None:
                 exported=bool(len(exports)),
                 internal=bool(len(functions)) and not bool(len(exports)),
                 findable=plugincls.__findable__,
+                alias=getattr(attr, "__alias__", False),
```
Schamper, line 562:
```suggestion
                alias=False,
```
I don't think namespaces themselves can be aliased, right?

dissect/target/plugin.py
```diff
@@ -323,6 +325,9 @@ def clone_alias(cls: type, attr: Callable[..., Any], alias: str) -> None:
     clone.__name__ = alias
     clone.__qualname__ = f"{cls.__name__}.{alias}"
 
+    # Mark this clone as an alias
+    clone.__alias__ = True
+
```
Schamper, line 330:
```suggestion
```

Schamper (changes requested):

tests/test_plugin.py
```diff
@@ -811,6 +827,60 @@ def test_plugins_default_plugin(target_default: Target) -> None:
     assert len(list(default_os_plugin_desc)) == 1
 
 
+def test_function_aliases(target_default: Target) -> None:
+    """test if alias functions are tagged as such correctly."""
+
+    # function that is an alias should have an alias property set to True
+    syslog_fd = next(f for f in find_functions("*", target_default)[0] if f.name == "syslog")
```
Schamper, line 834: Why not search for `syslog` directly?

Schamper (changes requested):

Schamper (approved):

PR #1051, opened 2025-02-25
Sanitize linux and windows hostnames

JSCU-CNI:
This PR adds support for cleaning Windows and Linux hostnames of invalid characters.


dissect/target/plugins/os/windows/_os.py
```diff
@@ -103,7 +104,7 @@ def add_mounts(self) -> None:
     def hostname(self) -> str | None:
         key = "HKLM\\SYSTEM\\ControlSet001\\Control\\Computername\\Computername"
         try:
-            return self.target.registry.value(key, "Computername").value
+            return sanitize_string(self.target.registry.value(key, "Computername").value)
```
Horofic, line 107: Could you add a test for Windows as well?

dissect/target/helpers/utils.py
```diff
@@ -162,3 +162,9 @@ def year_rollover_helper(
                 continue
 
             yield relative_ts.replace(tzinfo=tzinfo), line
+
+
+def sanitize_string(value: str) -> str:
```
Horofic, line 167: Maybe something along the lines of `escape_null()` might be more striking? Since it only does that at the moment.

Horofic (dismissed):

Horofic:
General comment. What issue is this trying to solve? Replacing (read changing) hostname contents feels a little off

dissect/target/tools/utils.py
```diff
@@ -294,3 +294,8 @@ def find_and_filter_plugins(
         executed_plugins.add(func_def.name)
 
         yield func_def
+
+
+def escape_str(value: str) -> str:
+    """Escape non-ASCII, unicode characters and bytes to a printable form."""
+    return repr(value).strip("'")
```
Schamper, line 301: Note that if there's a `'` character in `value`, `repr` will use `"` instead.

Schamper (changes requested):

Schamper (approved):

PR #1054, opened 2025-02-26
Fix compatible plugins filter

JSCU-CNI:
This PR fixes the `find_plugins` function with `compatibility=True`. Previously descriptors were yielded if they were not seen before, regardless of their compatibility.

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -407,7 +407,7 @@ def test_find_functions_linux(target_linux: Target) -> None:
 
 
 def test_find_functions_compatible_check(target_linux: Target) -> None:
-    """test if we correctly check for compatibility in `find_functions` and `_filter_compatible`."""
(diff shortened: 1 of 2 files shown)


Schamper (changes requested):
Woops 😄 could you add a unit test, or change an existing one to showcase the behaviour?

tests/test_plugin.py
```diff
@@ -406,6 +406,12 @@ def test_find_functions_linux(target_linux: Target) -> None:
     assert found[0].path == "os.unix.linux.services.services"
 
 
+def test_find_functions_compatible_check(target_linux: Target) -> None:
+    """test if we correctly check for compatibility in `find_functions` and `_filter_compatible`."""
```
Schamper, line 410:
```suggestion
    """test if we correctly check for compatibility in ``find_functions`` and ``_filter_compatible``."""
```

Schamper (changes requested):

Schamper (approved):

PR #1058, opened 2025-03-04
Translate empty Windows evtx value to `None`

JSCU-CNI:
This PR replaces empty evtx values `"-"` to `None` in `format_value`. We also have added some example `Security.evtx` logs in the tests.


dissect/target/plugins/os/windows/log/evtx.py
```diff
@@ -91,16 +94,19 @@ def _parse_chunk(self, _, chunk: bytes) -> Iterator[Record]:
...
-    def _build_record(self, evtx_record) -> Record:
+    def _build_record(self, evtx_record: dict, source: Path) -> Record:
         # predictable order of fields in the list is important, since we'll
         # be constructing a record descriptor from it.
         evtx_record_fields = sorted(evtx_record.items())
 
-        record_values = {"_target": self.target}
+        record_values = {
+            "_target": self.target,
+            "evtx_source": source,
```
Schamper, line 108:
```suggestion
            "source": source,
```
Isn't `source` more widely used in other plugins?

Schamper (changes requested):

dissect/target/plugins/os/windows/log/evtx.py
Schamper: Maybe handle that in the loop below? Like adding a `_` at the end if there's a conflict with an existing field name?

Theoretically `ts`, `evtx_source`, `hostname` and `domain` could be key names too 😄 and we don't actually deal with that either.

dissect/target/plugins/os/windows/log/evtx.py
```diff
@@ -140,13 +146,18 @@ def _build_record(self, evtx_record) -> Record:
                 if key == "EventID":
                     value = int(value)
 
+                if key.lower() in (lower_keys := {k.lower() for k in record_values.keys()}):
```
Schamper, line 149: Are EVTX record fields case insensitive?

dissect/target/plugins/os/windows/log/evtx.py
```diff
@@ -140,13 +146,18 @@ def _build_record(self, evtx_record) -> Record:
                 if key == "EventID":
                     value = int(value)
 
+                if key in (keys := set(record_values.keys())):
```
Schamper, line 149:
```suggestion
                if key in record_values:
```
Don't think you need the set or `.keys()`, can just do it in the dictionary directly.

Need to find an appropriate type hint though.

Schamper (changes requested):

dissect/target/plugins/os/windows/log/evtx.py
```diff
@@ -169,3 +180,20 @@ def format_value(value: Any) -> Any:
...
+
+def unique_key(key: str, dictionary: dict[str, Any], count: int | None = None) -> str:
+    """Return a unique key for a given dict of key value pairs.
+
+    Makes the returned key unique by appending an incrementing integer after the given key name (e.g. ``key_2``).
+    Search is case sensitive so provide lower-cased ``key`` and ``dictionary`` arguments if case-insensitiveness
+    is desired.
+    """
+    count = count or 2
+    new_key = f"{key}_{count}"
```
Schamper, line 193: My only concern is that this behavior is exactly the same as with lists, maybe choose something different? Just adding an incremental amount of underscores? 😂

Schamper (changes requested):

dissect/target/plugins/os/windows/log/evtx.py
Schamper: Or put `duplicate` behind the digit? Maybe a little less confusing. I don't have a super strong opinion on it, just trying to think of something to avoid confusion.

tests/plugins/os/windows/log/test_evtx.py
```diff
@@ -100,3 +111,48 @@ def test_evtx_scraping(target_win):
...
+
+    # Should contain None instead of string "-"
+    assert {getattr(r, "IpAddress", None) for r in records} == {None, "127.0.0.1"}
+
+
+@pytest.mark.parametrize(
+    ("key", "keys", "expected_key"),
+    [
+        ("source", {"source", "_target"}, "source_2"),
+        ("source", {"source", "source_2", "_target"}, "source_3"),
```
Schamper, line 152: I assume these will need updating.

Schamper (changes requested):

dissect/target/plugins/os/windows/log/evtx.py
```diff
@@ -91,16 +94,19 @@ def _parse_chunk(self, _, chunk: bytes) -> Iterator[Record]:
         chnk = evtx.ElfChnk(chunk)
         try:
             for event in chnk.read():
-                yield self._build_record(event)
+                yield self._build_record(event, None)
         except MalformedElfChnkException:
             pass
 
-    def _build_record(self, evtx_record) -> Record:
+    def _build_record(self, evtx_record: dict, source: Path) -> Record:
```
Schamper, line 101:
```suggestion
    def _build_record(self, evtx_record: dict, source: Path | None) -> Record:
```

Schamper (changes requested):

Schamper (approved):

PR #1059, opened 2025-03-05
Fix compatible plugins filter (2)

JSCU-CNI:
This PR fixes the `find_plugins` function with `compatibility=True` for plugin classes with multiple exported functions. Previously only the first descriptor was yielded and any other descriptors of the same plugin class were ignored.

This solution introduces a slightly slower `_filter_compatible` function. We could also keep `seen: set[Plugin]` and iterate over `plugincls.__functions__` once the plugin is determined to be compatible, but that might not work as intended for namespace plugins and regular plugins with multiple function exports.

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -966,6 +966,7 @@ def _filter_compatible(
             plugincls = load(descriptor)
         except Exception:
             if ignore_load_errors:
+                incompatible.add(descriptor.qualname)
                 continue
             raise
 
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -951,7 +951,9 @@ def _filter_compatible(
     descriptors: list[FunctionDescriptor], target: Target, ignore_load_errors: bool = False
 ) -> Iterator[FunctionDescriptor]:
     """Filter a list of function descriptors based on compatibility with a target."""
-    seen = set()
+    compatible = set()
+    incompatible = set()
+
     for descriptor in descriptors:
         try:
             plugincls = load(descriptor)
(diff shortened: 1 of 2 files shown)


dissect/target/plugin.py
```diff
@@ -960,8 +960,8 @@ def _filter_compatible(
                 continue
             raise
 
-        if plugincls not in seen:
-            seen.add(plugincls)
+        if descriptor.path not in seen:
```
Schamper, line 963: You can also keep two sets of classes, one of compatible and one of incompatible. That way you can still do a shortcut for both paths (immediately yield if in compatible, skip if in incompatible).

Schamper (changes requested):
Can you also add a unit test.

dissect/target/plugin.py
```diff
@@ -951,7 +951,9 @@ def _filter_compatible(
     descriptors: list[FunctionDescriptor], target: Target, ignore_load_errors: bool = False
 ) -> Iterator[FunctionDescriptor]:
     """Filter a list of function descriptors based on compatibility with a target."""
-    seen = set()
+    compatible = set()
+    incompatible = set()
+
     for descriptor in descriptors:
         try:
             plugincls = load(descriptor)
```
Schamper, line 959: Maybe it's an idea to make the incompatible/compatible check on the descriptor? Then we can also skip loading it if we know it's already compatible or incompatible (or failed to load previously).

Schamper (changes requested):

dissect/target/plugin.py
```diff
@@ -951,22 +951,31 @@ def _filter_compatible(
...
+            continue
+
+        if descriptor.qualname in incompatible:
+            continue
+
         try:
             plugincls = load(descriptor)
         except Exception:
             if ignore_load_errors:
                 continue
```
Schamper, line 969: You can add it to `incompatible` here too.

Schamper (approved):

PR #1061, opened 2025-03-07
Extend mru plugin to parse OpenSavePidlMRU and LastVisitedPidlMRU keys

M1ra1B0T:
This pull request extends the mru plugin and adds the capabilities to parse OpenSavePidlMRU and LastVisitedPidlMRU mru keys, which are existing since Windows Vista.


dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -353,30 +362,63 @@ def parse_mru_key(target, key, record):
...
             continue
 
         entry_index = mrulist_ex.index(int(value.name))
-        split_idx = value.value.index(b"\x00\x00")
-        # Poor mans null terminated utf-16-le
-        path, bag = value.value[: split_idx + 1], value.value[split_idx + 3 :]
+        if record != OpenSaveMRURecord:
+            split_idx = value.value.index(b"\x00\x00")
+            # Poor mans null terminated utf-16-le
+            path, bag = value.value[: split_idx + 1], value.value[split_idx + 3 :]
```
Schamper, line 379: Doesn't `filename, path, _ = value.value.rsplit(b"\x00\x00")` with here, same as in `lastvisited`?

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -353,30 +362,63 @@ def parse_mru_key(target, key, record):
...
 
     for value in key.values():
         if value.name == "MRUListEx":
             continue
 
         entry_index = mrulist_ex.index(int(value.name))
-        split_idx = value.value.index(b"\x00\x00")
-        # Poor mans null terminated utf-16-le
-        path, bag = value.value[: split_idx + 1], value.value[split_idx + 3 :]
+        if record != OpenSaveMRURecord:
```
Schamper, line 376: Can you implement this behavior in `parse_mru_key` too so we can reuse the same logic in e.g. `lastvisited`?

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -419,3 +461,13 @@ def parse_office_mru_key(target, key, record):
             _user=user,
             _key=key,
         )
+
+
+def get_filepath_from_bag(parsed_bag):
```
Schamper, line 466: Can you add type hints? At least to the new code, but we would really appreciate it if you added it to the entire file <3.

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -353,30 +362,63 @@ def parse_mru_key(target, key, record):
...
+                _target=target,
+                _user=user,
+                _key=key,
+            )
+        else:
+            yield record(
+                regf_mtime=key.ts,
+                index=entry_index,
+                value=path.decode("utf-16-le"),
+                key=key.path,
```
Schamper, line 414: Just a general comment, I think we can get rid of the `key` value in this record, since the registry record extension will add it as `regf_key_path` already.

Schamper (changes requested):

dissect/target/plugins/os/windows/regf/mru.py
M1ra1B0T: @Schamper Technically speaking its not the same, the `path` / `bag` split is not handling multiple consecutive null byte occurrences the same as the code in `lastvisited`. The code here will split at the first occurrence of `\00\00`. If there is another occurrence it will just be a part of the `bag` variable. The code in `lastvisited` however would add everything after a second occurrence of a null bytes sequence to the `_` variable and it wouldn't be passed on to the follow up code (`parse_shell_item_list`). With my test data changing these lines of code does not seem to have had an impact or cause any issues in the shellbags plugin code, as the resulting bags/paths were the same. However, I can't rule out the possibility that this might have side effects - I am just not firm enough in the different valid shellbag structures and how your shellbags plugin would handle these. If there are no valid shellbag structures with consecutive null bytes within, then we could change it to the logic used in `lastvisited`. How do you want to proceed?

dissect/target/plugins/os/windows/regf/mru.py
M1ra1B0T: @Schamper Technically speaking its not the same, the `path` / `bag` split is not handling multiple consecutive null byte occurrences the same as the code in `lastvisited`. The code here will split at the first occurrence of `\00\00`. If there is another occurrence it will just be a part of the `bag` variable. The code in `lastvisited` however would add everything after a second occurrence of a null bytes sequence to the `_` variable and it wouldn't be passed on to the follow up code (`parse_shell_item_list`). With my test data changing these lines of code does not seem to have had an impact or cause any issues in the shellbags plugin code, as the resulting bags/paths were the same. However, I can't rule out the possibility that this might have side effects - I am just not firm enough in the different valid shellbag structures and how your shellbags plugin would handle these. If there are no valid shellbag structures with consecutive null bytes within, then we could change it to the logic used in `lastvisited`. How do you want to proceed?

Schamper: Maybe something like this will work a bit better? This code also doesn't assume that every UTF-16-LE character ends with a null byte (both bytes can be occupied):

```python
def read_wstrings(buf):
    result = []

    idx = 0
    remaining = buf
    while (idx := remaining.find(b"\x00\x00", idx)) != -1:
        if idx & 1:
            idx += 1
            continue

        result.append(remaining[:idx])
        remaining = remaining[idx + 2:]
        idx = 0

    if remaining:
        result.append(remaining)

    return result
```

M1ra1B0T: I have to correct myself. After having another look into my test data I did notice that there are actually examples, where multiple consecutive null bytes occur within a shellbag structure, e.g. when you are having a shell item list with multiple folder/file entries. So I guess we should not do further splits after the first occurrence of `\00\00`, as this will compromise the structure we want to hand over to the `parse_shell_item_list` function.

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -158,57 +161,43 @@ def opensave(self) -> Iterator[OpenSaveMRURecord]:
...
     @export(record=LastVisitedMRURecord)
     def lastvisited(self) -> Iterator[LastVisitedMRURecord]:
         """Return the LastVisitedMRU data.
 
         The ``HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\ComDlg32\\LastVisitedMRU`` registry key
-        contains information about the executable used by an application to open the files that are documented at the
-        OpenSaveMRU registry key. Also each value tracks the directory location for the last file that was accessed by
-        that application.
+        - renamed to ``LastVisitedPidlMRU`` since Windows Vista - contains information about the executable used by
+        an application to open the files that are documented at the OpenSaveMRU registry key. Also, each value
```
Schamper, line 186:
```suggestion
        an application to open the files that are documented at the ``OpenSaveMRU`` registry key. Also, each value
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -321,7 +308,9 @@ def msoffice(self) -> Iterator[MSOfficeMRURecord]:
                         pass
 
 
-def parse_mru_key(target, key, record):
+def parse_mru_key(
+    target: Target, key: RegistryKey, record: UserRegistryRecordDescriptor
```
Schamper, line 312:
```suggestion
    target: Target, key: RegistryKey, record: TargetRecordDescriptor
```
Technically not very correct still but it's the best we can do with the current typing on `flow.record` stuff.

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -321,7 +308,9 @@ def msoffice(self) -> Iterator[MSOfficeMRURecord]:
...
+) -> Iterator[UserRegistryRecordDescriptor]:
```
Schamper, line 313:
```suggestion
) -> Iterator[RunMRURecord | OpenSaveMRURecord | MapNetworkDriveMRURecord | LastVisitedMRURecord]:
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -333,53 +322,98 @@ def parse_mru_key(target, key, record):
...
+            )
 
     for subkey in key.subkeys():
         yield from parse_mru_key(target, subkey, record)
 
 
-def parse_mru_ex_key(target, key, record):
+def parse_mru_ex_key(
+    target: Target, key: RegistryKey, record: UserRegistryRecordDescriptor
+) -> Iterator[UserRegistryRecordDescriptor]:
```
Schamper, line 357: Can you do the same here.

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -333,53 +322,98 @@ def parse_mru_key(target, key, record):
...
-            _target=target,
-            _user=user,
-            _key=key,
-        )
+    for subkey in key.subkeys():
+        yield from parse_mru_ex_key(target, subkey, record)
 
 
-def parse_office_mru(target, key, record):
+def parse_office_mru(target: Target, key: RegistryKey, record: MSOfficeMRURecord) -> Iterator[MSOfficeMRURecord]:
```
Schamper, line 416:
```suggestion
def parse_office_mru(target: Target, key: RegistryKey, record: type[MSOfficeMRURecord]) -> Iterator[MSOfficeMRURecord]:
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -397,7 +431,7 @@ def parse_office_mru(target, key, record):
         pass
 
 
-def parse_office_mru_key(target, key, record):
+def parse_office_mru_key(target: Target, key: RegistryKey, record: MSOfficeMRURecord) -> Iterator[MSOfficeMRURecord]:
```
Schamper, line 434:
```suggestion
def parse_office_mru_key(target: Target, key: RegistryKey, record: type[MSOfficeMRURecord]) -> Iterator[MSOfficeMRURecord]:
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -333,53 +322,98 @@ def parse_mru_key(target, key, record):
...
+            split_idx = value.value.index(b"\x00\x00")
+            # Poor mans null terminated utf-16-le
+            path, bag = value.value[: split_idx + 1], value.value[split_idx + 3 :]
+        else:
+            bag = value.value
+            path = None
         parsed_bag = list(parse_shell_item_list(bag))
         if len(parsed_bag) != 1 or not isinstance(parsed_bag, FILE_ENTRY):
             target.log.debug("Unexpected shell bag entry in MRUListEx entry: %s:%s", key, value)
+        filepath = get_filepath_from_bag(parsed_bag)
```
Schamper, line 381:
```suggestion
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -333,53 +322,98 @@ def parse_mru_key(target, key, record):
...
+            yield record(
+                regf_mtime=key.ts,
+                index=entry_index,
+                filename=path.decode("utf-16-le"),
+                path=filepath,
+                _target=target,
+                _user=user,
+                _key=key,
+            )
+        elif record == OpenSaveMRURecord:
```
Schamper, line 393:
```suggestion
        elif record == OpenSaveMRURecord:
            filepath = "\\".join(el.name.rstrip("\\") for el in parsed_bag)
```
I think this accomplishes the same thing?

dissect/target/plugins/os/windows/regf/mru.py
Schamper: The problem with your current logic is that it doesn't work for UTF-16-LE data that uses the full 16bit width for a character.

```python
In [10]: "🦊\0A\0B".encode("utf-16-le")
Out[10]: b'>\xd8\x8a\xdd\x00\x00A\x00\x00\x00B\x00'
#                       ^^^^^^^^ ^^^^^^^^
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -414,8 +448,17 @@ def parse_office_mru_key(target, key, record):
...
             index=entry_index,
             value=path,
-            key=key.path,
             _target=target,
             _user=user,
             _key=key,
         )
+
+
+def get_filepath_from_bag(parsed_bag: list[SHITEM]) -> str:
```
Schamper, line 457: This function can be removed if that other suggestion achieves the same.

dissect/target/plugins/os/windows/regf/mru.py
Schamper, line 1: Can you add a unit test?

Schamper (changes requested):
Another general improvement question for this plugin, it's been a long time since I've looked at MRU so forgive me if it's a dumb question: it looks like a lot of the parsed out values are file paths, can we use the new (as in, didn't exist at the time of writing this plugin originally) `path` field type? That way we can benefit when running plugins that yield records containing file paths.

Schamper:
> Mh, you could end up with paths that start with or contain parts like `Shared Documents Folder (Users Files)\\<USERS_PROPERTY_VIEW {f42ee2d3-909f-4907-8871-4c22fc0bf756}>\\`, were `<` and `>`would be forbidden characters for windows paths, however does this affect your path handling?

Hm alright, nevermind in that case. Just keep that as it is then 😄

dissect/target/plugins/os/windows/regf/mru.py
Schamper: Rebasing from `main` will take care of this.

dissect/target/plugins/os/windows/regf/mru.py
M1ra1B0T: @Schamper I have currently no idea on how to resolve this.

Schamper: You can use the code in this comment as an example on how to deal with it: https://github.com/fox-it/dissect.target/pull/1061#discussion_r2000888643

M1ra1B0T: @Schamper So maybe we can go through this with an example? This is my example registry value for `lastvisited`:

```
b'{\x003\x007\x00F\x00B\x000\x007\x00C\x009\x00-\x005\x00B\x001\x00B\x00-\x004\x00F\x001\x00D\x00-\x009\x00B\x00C\x008\x00-\x008\x003\x006\x009\x008\x008\x006\x00F\x003\x005\x002\x00C\x00}\x00\x00\x00\x14\x00\x1fP\xe0O\xd0 \xea:i\x10\xa2\xd8\x08\x00+00\x9d\x19\x00/C:\\\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00x\x001\x00\x00\x00\x00\x00\x9dO\xdd\x8c\x11\x00Users\x00d\x00\t\x00\x04\x00\xef\xbe\x8bL\x91\xa8\x92P\xfap.\x00\x00\x00\xea\x05\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00:\x00\x00\x00\x00\x00|\xebv\x00U\x00s\x00e\x00r\x00s\x00\x00\x00@\x00s\x00h\x00e\x00l\x00l\x003\x002\x00.\x00d\x00l\x00l\x00,\x00-\x002\x001\x008\x001\x003\x00\x00\x00\x14\x00Z\x001\x00\x00\x00\x00\x00\x90P\x96\xa2\x10\x00M925C~1.MEI\x00@\x00\t\x00\x04\x00\xef\xbe\x9dO\xdd\x8c\x92Por.\x00\x00\x00\xe4u\x01\x00\x00\x00\x06\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa4\xbeM\x00m\x00.\x00m\x00e\x00i\x00e\x00r\x00\x00\x00\x1a\x00~\x001\x00\x00\x00\x00\x00\x92Pkt\x11\x00Desktop\x00h\x00\t\x00\x04\x00\xef\xbe\x9dO\xdd\x8c\x92Pkt.\x00\x00\x00l|\x01\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00>\x00\x00\x00\x00\x00\x93\xcbI\x00D\x00e\x00s\x00k\x00t\x00o\x00p\x00\x00\x00@\x00s\x00h\x00e\x00l\x00l\x003\x002\x00.\x00d\x00l\x00l\x00,\x00-\x002\x001\x007\x006\x009\x00\x00\x00\x16\x00\x00\x00'
```

It should get us the following file path result: `My Computer\\C:\\Users\\m.meier\\Desktop\\` 

With the logic that is currently implemented it splits into:
```
path = b'{\x003\x007\x00F\x00B\x000\x007\x00C\x009\x00-\x005\x00B\x001\x00B\x00-\x004\x00F\x001\x00D\x00-\x009\x00B\x00C\x008\x00-\x008\x003\x006\x009\x008\x008\x006\x00F\x003\x005\x002\x00C\x00}\x00'

bag = b'\x14\x00\x1fP\xe0O\xd0 \xea:i\x10\xa2\xd8\x08\x00+00\x9d\x19\x00/C:\\\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00x\x001\x00\x00\x00\x00\x00\x9dO\xdd\x8c\x11\x00Users\x00d\x00\t\x00\x04\x00\xef\xbe\x8bL\x91\xa8\x92P\xfap.\x00\x00\x00\xea\x05\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00:\x00\x00\x00\x00\x00|\xebv\x00U\x00s\x00e\x00r\x00s\x00\x00\x00@\x00s\x00h\x00e\x00l\x00l\x003\x002\x00.\x00d\x00l\x00l\x00,\x00-\x002\x001\x008\x001\x003\x00\x00\x00\x14\x00Z\x001\x00\x00\x00\x00\x00\x90P\x96\xa2\x10\x00M925C~1.MEI\x00@\x00\t\x00\x04\x00\xef\xbe\x9dO\xdd\x8c\x92Por.\x00\x00\x00\xe4u\x01\x00\x00\x00\x06\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa4\xbeM\x00m\x00.\x00m\x00e\x00i\x00e\x00r\x00\x00\x00\x1a\x00~\x001\x00\x00\x00\x00\x00\x92Pkt\x11\x00Desktop\x00h\x00\t\x00\x04\x00\xef\xbe\x9dO\xdd\x8c\x92Pkt.\x00\x00\x00l|\x01\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00>\x00\x00\x00\x00\x00\x93\xcbI\x00D\x00e\x00s\x00k\x00t\x00o\x00p\x00\x00\x00@\x00s\x00h\x00e\x00l\x00l\x003\x002\x00.\x00d\x00l\x00l\x00,\x00-\x002\x001\x007\x006\x009\x00\x00\x00\x16\x00\x00\x00'
```

With the `read_wstrings` function however I get a result list with 48 items and a lot of the null bytes are removed, so that the data structure we want to hand over to the function `parse_shell_item_list(bag)` is corrupted. 

 Currently I am missing a way to bring both approaches together. Or maybe I just can't see it right now.

Schamper: @M1ra1B0T if we turn`read_wstrings` into a decorator, we can use it for your purpose as well as keeping it as a fairly universal utility function:

```python
In [5]: def read_wstrings(buf):
   ...:     idx = 0
   ...:     remaining = buf
   ...:     while (idx := remaining.find(b"\x00\x00", idx)) != -1:
   ...:         if idx & 1:
   ...:             idx += 1
   ...:             continue
   ...:
   ...:         yield remaining[:idx]
   ...:         remaining = remaining[idx + 2:]
   ...:         idx = 0
   ...:
   ...:     if remaining:
   ...:         yield remaining
   ...:

In [6]: next(read_wstrings(buf), None)
Out[6]: b'{\x003\x007\x00F\x00B\x000\x007\x00C\x009\x00-\x005\x00B\x001\x00B\x00-\x004\x00F\x001\x00D\x00-\x009\x00B\x00C\x008\x00-\x008\x003\x006\x009\x008\x008\x006\x00F\x003\x005\x002\x00C\x00}\x00'
```

Using `next()` we just take the first wstring from the buffer, and then we can take the bag with `buf[len(path):]`.

The reason I'm being a bit difficult about this part is that I foresee this may become a problem on non-Latin language Windows systems. Best to tackle this while we're working on it.

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -424,8 +451,23 @@ def parse_office_mru_key(target: Target, key: RegistryKey, record: TargetRecordD
...
             index=entry_index,
             value=path,
-            key=key.path,
             _target=target,
             _user=user,
             _key=key,
         )
+
+
+def read_wstrings(buf: bytes) -> Generator[bytes, None, None]:
```
Schamper, line 460:
```suggestion
def read_wstrings(buf: bytes) -> Iterator[bytes]:
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -343,18 +327,31 @@ def parse_mru_key(target: Target, key: RegistryKey, record: TargetRecordDescript
...
-        )
+        if record == LastVisitedMRURecord:
+            entry_index = mrulist.index(value.name) if mrulist else None
+            filename, path, _ = value.value.rsplit(b"\x00\x00")
+
+            yield record(
+                regf_mtime=key.ts,
+                index=entry_index,
+                filename=filename.decode("utf-16-le"),
+                path=path.decode("utf-16-le"),
```
Schamper, line 338:
```suggestion
            values = read_wstrings(value.value)

            yield record(
                regf_mtime=key.ts,
                index=entry_index,
                filename=next(values).decode("utf-16-le"),
                path=next(values).decode("utf-16-le"),
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -18,7 +18,7 @@
 )
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Generator, Iterator
```
Schamper, line 21:
```suggestion
    from collections.abc import Iterator
```

Schamper (changes requested):

PR #1062, opened 2025-03-11
Improve target-query exception handling and logging

JSCU-CNI:
This PR improves `target-query`'s exception handling log messages. The previous log message used the variable `func` which is not correct.


dissect/target/tools/query.py
```diff
@@ -268,13 +268,14 @@ def main() -> None:
...
-                    target.log.exception("Exception while opening child '%s'", args.child)
+                except Exception as e:
+                    target.log.exception("Exception while opening child %r: %s", args.child, e)
+                    target.log.debug("", exc_info=e)
 
             if args.dry_run:
                 print(f"Dry run on: {target}")
 
-            record_entries = []
+            record_entries: list[Generator] = []
```
Schamper, line 278:
```suggestion
            record_entries: list[Iterator[Record]] = []
```
Right?

dissect/target/tools/query.py
```diff
@@ -373,23 +373,27 @@ def main() -> None:
...
                             break
+
                 except Exception as e:
-                    # Ignore errors if multiple functions
-                    if len(funcs) > 1:
-                        target.log.error(f"Exception occurred while processing output of {func}", exc_info=e)
+                    # Ignore errors if multiple functions or multiple targets
+                    if len(record_entries) > 1 or len(args.targets) > 1:
+                        likely_func_name = getattr(func_generator, "__qualname__", "unknown (chained) iterator")
+                        target.log.error("Exception occurred while processing output of %s: %s", likely_func_name, e)
```
Schamper, line 389: Can't you use `func_def.name` here?

dissect/target/tools/query.py
```diff
@@ -373,23 +373,27 @@ def main() -> None:
                 continue
 
             rs = record_output(args.strings, args.json)
-            for entry in record_entries:
+            for func_generator in record_entries:
```
Schamper, line 376: `record_entries` -> `record` makes a lot of sense, but I don't necessarily find this one more readable.

Schamper (changes requested):

dissect/target/tools/query.py
Schamper: Could also change the `record_entries` list to contain a tuple of the generator and `func_def` if that makes it any easier.

Schamper (approved):

PR #1069, opened 2025-03-19
Fix TeamViewer timezone parsing

JSCU-CNI:
Currently the TeamViewer plugin acts as if all timestamps are UTC. They however are not. This PR fixes that.

### DIFF
--- a/dissect/target/plugins/os/windows/log/evtx.py
+++ b/dissect/target/plugins/os/windows/log/evtx.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 import re
 from functools import lru_cache
+from pathlib import Path
 from typing import Any, Iterator
 
 from dissect.eventlog import evtx
@@ -75,8 +76,10 @@ class EvtxPlugin(WindowsEventlogsMixin, plugin.Plugin):
                 self.target.log.exception("Failed to open event log: %s", entry)
                 continue
 
+            self.target.log.info("Processing event log file %s", entry)
+
             for event in evtx.Evtx(entry_data):
-                yield self._build_record(event)
+                yield self._build_record(event, entry)
 
     @plugin.export(record=DynamicDescriptor(["datetime"]))
     def scraped_evtx(self) -> Iterator[DynamicDescriptor]:
@@ -91,16 +94,19 @@ class EvtxPlugin(WindowsEventlogsMixin, plugin.Plugin):
         chnk = evtx.ElfChnk(chunk)
         try:
             for event in chnk.read():
-                yield self._build_record(event)
+                yield self._build_record(event, None)
         except MalformedElfChnkException:
             pass
 
-    def _build_record(self, evtx_record) -> Record:
+    def _build_record(self, evtx_record: dict, source: Path | None) -> Record:
         # predictable order of fields in the list is important, since we'll
         # be constructing a record descriptor from it.
         evtx_record_fields = sorted(evtx_record.items())
 
-        record_values = {"_target": self.target}
+        record_values = {
+            "_target": self.target,
+            "source": source,
+        }
         record_fields = [
             ("datetime", "ts"),
             ("string", "Provider_Name"),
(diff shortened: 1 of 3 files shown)


dissect/target/plugins/apps/remoteaccess/teamviewer.py
```diff
@@ -45,82 +70,140 @@ class TeamViewerPlugin(RemoteAccessPlugin):
...
+                    except UnicodeDecodeError as e:
+                        self.target.log.warning("Unable to parse log line in %s: %s", logfile, e)
+                        self.target.log.debug("", exc_info=e)
                         continue
 
-                    # End of file, quit while loop
                     if not line:
                         break
 
                     line = line.strip()
```
Schamper, line 121:
```suggestion
```
Can just let it exhaust the iterator.

dissect/target/plugins/apps/remoteaccess/teamviewer.py
```diff
@@ -45,82 +70,140 @@ class TeamViewerPlugin(RemoteAccessPlugin):
...
-
-                    # Skip empty lines
-                    if not line:
+                    if not line or line.startswith("# "):
```
Schamper, line 122:
```suggestion
                    if not (line := line.strip()) or line.startswith("# "):
```

dissect/target/plugins/apps/remoteaccess/teamviewer.py
```diff
@@ -45,82 +70,140 @@ class TeamViewerPlugin(RemoteAccessPlugin):
...
+                    log = match.groupdict()
 
-                    # Correct for use of : as millisecond separator
-                    if ts_time.count(":") > 2:
-                        ts_time = ":".join(ts_time.split(":")[:3])
-                    # Correct for missing year in date
-                    if ts_day.count("/") == 1:
+                    # Older TeamViewer versions first mention the start time and then leave out the year,
+                    # so we have to correct for the missing year in date.
+                    if log["date"].count("/") == 1:
```
Schamper, line 142: Can you use variables for date and time instead of the dictionary every time?

dissect/target/plugins/apps/remoteaccess/teamviewer.py
```diff
@@ -45,82 +70,140 @@ class TeamViewerPlugin(RemoteAccessPlugin):
...
+
+        Start: 2021/11/11 12:34:56
+        Start: 2024/12/31 01:02:03.123 (UTC+2:00)
+    """
+
+    if match := RE_START.search(line):
+        dt = match.groupdict()
+
+        # Drop milliseconds
+        if "." in dt["time"]:
```
Schamper, line 196: Same here.

Schamper (changes requested):

dissect/target/plugins/apps/remoteaccess/teamviewer.py
Schamper: Can you add a unit test for why that manual `fh.readline()` is necessary?

Schamper (changes requested):
See comment regarding the `fh.readline()`.

dissect/target/plugins/apps/remoteaccess/teamviewer.py
Poeloe: The `while True` can definitely be replaces with just a `for line in fh:`. Please go ahead and 😄

Poeloe: Well, the `while True` is used to skip lines that cause a decoding error. In that way, we only skip a line, instead of the full remainder of the file.

We could add `errors="replace"` in the `with` statement, to replace the decoding errors in the lines and we just continue parsing that line with the replacements. Then we don't need the `while`-statement

Poeloe (approved):
Approved (the change to `errors="replace"` 😉)

Schamper (approved):

Schamper (approved):

PR #1076, opened 2025-03-24
Improve GenericTarSubloader for Windows-like tarballs

JSCU-CNI:
This PR improves the mapping of Windows targets in the GenericTarSubloader.

We now make sure to set the VirtualFilesystem to case-insensitive if the tar looks like a Windows target.

### DIFF
--- a/dissect/target/loaders/tar.py
+++ b/dissect/target/loaders/tar.py
@@ -15,11 +15,6 @@ from dissect.target.helpers import fsutil, loaderutil
 from dissect.target.helpers.lazy import import_lazy
 from dissect.target.loader import Loader, SubLoader
 
-ContainerImageTarSubLoader: TarSubLoader = import_lazy(
-    "dissect.target.loaders.containerimage"
-).ContainerImageTarSubLoader
-_GenericTarSubLoader: TarSubLoader = import_lazy("dissect.target.loaders.tar").GenericTarSubLoader
-
 log = logging.getLogger(__name__)
 
 TAR_EXT_COMP = (
@@ -38,7 +33,7 @@ TAR_MAGIC_COMP = (
     # gzip
     b"\x1f\x8b",
     # bzip2
-    b"\x42\x5A\x68",
+    b"\x42\x5a\x68",
     # xz
     b"\xfd\x37\x7a\x58\x5a\x00",
 )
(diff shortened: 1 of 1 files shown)


Schamper (changes requested):

dissect/target/loaders/tar.py
```diff
@@ -130,15 +139,22 @@ def detect(tarfile: tf.TarFile) -> bool:
 
     def map(self, target: target.Target) -> None:
         volumes = {}
+        windows_found = False
 
         for member in self.tar.getmembers():
             if member.name == ".":
                 continue
 
+            if member.name.startswith(WINDOWS_MEMBERS):
```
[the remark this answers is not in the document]
Schamper: I don't expect `WiNdOwS` to show up very often, but I can imagine in some cases you have one of `Windows`, `WINDOWS` or `windows`, depending on how and who created the archive. After all, you're checking whether you should mark the virtual filesystem as case-insensitive, but then you only do a case-sensitive check for Windows-like filesystems.

If it doesn't break anything or create unexpected side effects (I can't think of any at the moment), I don't think it can hurt.

Schamper (approved):

Schamper (changes requested):
Gotta fix the docs error 😄

Schamper:
> ```
> /home/runner/work/dissect.target/dissect.target/tests/_docs/api/dissect/target/loaders/tar/index.rst:148:duplicate object description of dissect.target.loaders.tar.GenericTarSubLoader, other instance in api/dissect/target/loaders/tar/index, use :no-index: for one of them
> ```
> 
> How would we add a `:no-index:` here? It seems like the lazy-importing causes a duplicate reference in the generated autoapi docs.

I vaguely remember a certain kind of annotation you can add to skip something for autoapi, but I can't find it right now. Maybe just prefixing it with `_` will already work?

dissect/target/loaders/tar.py
```diff
@@ -18,7 +18,7 @@
 ContainerImageTarSubLoader: TarSubLoader = import_lazy(
```
Schamper, line 18:
```suggestion
_ContainerImageTarSubLoader: TarSubLoader = import_lazy(
```

dissect/target/loaders/tar.py
```diff
@@ -46,13 +46,20 @@
...
+    "winnt",
+    "/winnt",
+)
+
 
 class TarLoader(Loader):
     """Load tar files."""
 
     __subloaders__ = [
         ContainerImageTarSubLoader,
```
Schamper, line 61:
```suggestion
        _ContainerImageTarSubLoader,
```

Schamper (changes requested):

Schamper:
Fixed it by moving the code around.

Schamper (approved):

Issue #789, opened 2024-08-05
Plugins should be able to expose their parsers independently of a target


MaxGroot:
## Problem description

(Chose as generic of a title as I could think of)

Dissect.target has plugins for all kinds of artifacts. It would be nice if you could query plugins even if there isn't a real 'target'. For example, calling the evtx plugin on a .evtx file, parsing a wtmp file, or recovering a file that has been quarantined by Windows Defender.

Currently, the plugin system is not really designed for this. Many plugins have a list of paths to check against, but these paths are not exposed / set via a uniform API. This prevents us from extending / providing the needed paths on the fly.  A band-aid solution has been the `LogLoader`, which remaps paths from the host into a virtual filesystem at a location that makes the `check_compatible` of the desired plugin pass.  This currently only supports a few forensic artifacts. The problem is that improving on this would either require:

1. Having a very long list of hardcoded paths in the `LogLoader`;
2. Requiring the user to manually specify the `LogLoader` with a path to map to, asking of them to know exactly which path dissect expects for a forensic artifact (proposed in https://github.com/fox-it/dissect.target/pull/787);
3. Using a dropfolder of sorts (proposed and closed without merge in https://github.com/fox-it/dissect.target/pull/399);

However, all three would run into flexibility constraints when plugins also require a certain os check to be passed, or a file existing in a user's home on the virtual target;

This problem has been discussed more thoroughly in https://github.com/fox-it/dissect.target/pull/399:

> Underwater, plugins could use something like get_paths() (that in turn calls the implemented methods of that plugin) but also allows getting "extra" paths from the command line, loader logic, or whatever other API. I can imagine a system like this also making the log loader obsolete, since you would simply be able to execute e.g. the evtx plugin directly on some paths you pass in from the CLI.

- Ideally, a long-term solution would make the `LogLoader` obsolete. As stated in #787:
- > We are planning to enhance our plugin functionality so you can query plugins with just target-query. 

- Ideally, the taken route is generic enough that it can also work with registry hives (https://github.com/fox-it/dissect.target/issues/411) 

## Opinionated suggestion

[...]

PR #1080, opened 2025-03-26
Add Windows RDP bitmap cache plugin

JSCU-CNI:
This PR adds functionality to recover bitmaps from the Windows RDP Bitmap Cache. We've based the plugin on [bmc-tools](https://github.com/ANSSI-FR/bmc-tools). Unlike bmc-tools, this plugin does not support BMC files with a bits-per-pixel other than 32. Such files should be quite rare nowadays.

We did run into an existing problem: we cannot expose the parser easily so it can be run against separate `.bin` or `.bmc` files that do not reside in an already existing Target (as mentioned in issues such as https://github.com/fox-it/dissect.target/issues/789).

See the grid export below for example output of the `rdpcache.recover` plugin function.
image


Schamper:
https://github.com/fox-it/dissect.target/pull/1082 might allow this to work on individual files.

Schamper (changes requested):
Very nice contribution!

dissect/target/plugins/os/windows/rdpcache.py
```diff
@@ -0,0 +1,478 @@
...
+        dest="no_individual_tiles",
+        action="store_true",
+        help="do not save each recovered tile as a separate bitmap",
+    )
+    @arg(
+        "--remnants",
+        "-r",
+        choices=["only", "include", "exclude"],
+        default="exclude",
+        type="str",
```
Schamper, line 407:
```suggestion
        type=str,
```
This actually causes a failure when trying to execute the plugin. Something to unit test in the future 😄.

Schamper (changes requested):

Schamper (approved):

PR #1081, opened 2025-03-26
Add support for leap days in year_rollover_helper

JSCU-CNI:
This PR adds support for leap days in `year_rollover_helper` and generally makes the function more robust when encountering invalid dates.


Schamper (approved):
That must've been fun to debug 😄.

PR #1082, opened 2025-03-28
WIP

twiggler:


Schamper (changes requested):
I've proposed some changes in https://github.com/fox-it/dissect.target/pull/1082/commits/054c9c391158aee6d406994173c7fe096ee16bee

Let me know what you think.

As I was looking at this and fiddling with it a bit myself, I was thinking about how this could work for all types of plugins. Some idea sketches:
- For user bound paths, we could just add a `get_user_paths(self) -> Iterator[tuple[UserDetails, Path]]` and do the same as here
- Some plugins have complicated file requirements, e.g.:

```python
class Foo(Plugin):
    @export
    def bar(self): ...  # Uses a.txt

    @export
    def baz(self): ...  # Uses b.txt
```
It's not really possible to cover these with the current proposed solution. However, #1122 may provide an answer here? Perhaps you could bind something like:
```python
@export
def bar(self): ...
@bar.paths
def bar(self): return [self.target.fs.path("a.txt")]
```
That approach for plugins that are structured like this could also work for a possible `detect(fh)` feature as described in #789

- This method still doesn't quite cover "getting all paths that are relevant to a plugin", since it's only really meant for "paths the plugin will parse for artefacts", and not configuration paths. So we can't quite 1:1 use this method to collect files with `acquire`. I think we should focus on this approach first, but perhaps in a next stage we can think of how to tackle that, perhaps a `_get_aux_paths()` to be implemented by plugins, and a `Plugin.get_all_paths` to be consumed by `acquire`?

dissect/target/target.py
```diff
@@ -639,7 +632,7 @@ def add_plugin(
         if not isinstance(p, plugin.Plugin):
             raise PluginError(f"Not a subclass of Plugin: {p}")
 
-        if check_compatible:
+        if check_compatible and not self.is_direct:
```
twiggler, line 635: I also tried this solution, but ran into an issue that some plugins tried to load which should not load (something with fortis IIRC).

The tests seem to pass, but I still wonder if this is safe?

twiggler:
> One big change is to treat paths as-is as passed by the user. Globs should be expanded by the shell in my opinion, and there should be a clearer distinction between directories and files. Most plugins will work directly on files, however something like a browser plugin may restrict itself to profile directories instead.
> 
> Let me know what you think.

As discussed in private communication, I implemented  the glob expansion to support shells which do not auto-expand, most notably the windows shell, although powershell might do expansion.

However, as pointed out we don´t do manual expansion of command line arguments anywhere else. In the long term, manual expansion, when supported, is  perhaps best performed when parsing command line arguments, instead of  the "business logic" portion.

twiggler:
> Let me know what you think.
> 
> As I was looking at this and fiddling with it a bit myself, I was thinking about how this could work for all types of plugins. 
> Some idea sketches:

Great that you bring this up. I am having  doubts if this is the right path. Basically, every solution feels  a bit wrong to me, although I also think they are acceptable for the short- to mid term.

We had this discussion before, but at first sight there is tension between making it easy to write plugins, and separating concerns so that individual components are straightforward to compose.

Or perhaps this is a false dichotomy. Perhaps we can offer a simple facade to help people on their way, which abstract a collection of more fine-grained api's, while more seasoned developers can  program directly against those more complex api's.

Ultimately, it is much easier to compose indiviual components, then to break up an aggregate after the fact.

Now there is the constraint that we are programming in Python, and certain abstractions which might get compiled away in other languages, might result in performance problems. I don´t know.

Further, there is the more practical concern if navigating to a position where  things are easier composable is worth the cost investment. I don´t know yet.    

I will  try to get more concrete in a later post

dissect/target/plugins/apps/webserver/iis.py
```diff
@@ -76,269 +80,84 @@ def __init__(self, target: Target):
...
+                    self.target.log.info("Processing IIS log file %s in %s format", log_file, format)
+                    yield from parsers[format](self.target, log_file)
 
-        auto = self.get_files("auto")
-        for autoFile in auto:
-            self.target.log.info("Parsing IIS log file %s in unknown format", autoFile)
-            yield from self.parse_autodetect_format_log(autoFile)
+        # We don't implement _get_paths() in the IIS plugin because there's little use for it for the way the plugin
+        # is currently implemented. So handle direct files here.
+        if self.target.is_direct:
```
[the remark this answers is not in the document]
twiggler: I did a quick inventory of how this would pan out for other plugins which might benefit from single file support:

### nginx
Multiple categories of files discovered by parsing of config file and scanning of directories
Getting files per category  would require implementation of autodetection of error and access log
(the code is kinda there), and being able to pass a file category as an argument to get_files  

### caddy
works

### apache:
Multiple types of files discovered by parsing of config file and scanning of directories
Getting files per category (such as in iis) would require implementation autodetection of error and access log
(the code is kinda there),  and being able to pass a file category as an argument to get_files  

### citrix:
similar to apache

### mcaffee:
works

sophos:
Multiple categories of log_files , `get_ paths` requires category argument

symantec:
Multiple categories of log files, `get_ paths` requires category argument

trendmicro:
Multiple categories of log files, `get_ paths` requires category argument

teamviewer:
works

audit:
works

lastlog:
works

messages:
works

mssql:
works

dissect/target/target.py
twiggler: The root cause was that DefaultOS didn´t have a date time & locale plugin. This caused the plugin system to try date time & locale plugins of other OS'es, like FortiOS (this is the behavior only when the DefaultOS is selected). During the Ruff refactor, a default date time & locale plugin has been added.

In principle, turning off compatibility checks when loading a plugin is safe; we should a default implementation for date  time & local plugins.

In conclusion, I prefer this over  the previous implementation, because when single file / direct mode is engaged, you typically don´t have any meaningful checks.

dissect/target/plugins/apps/webserver/iis.py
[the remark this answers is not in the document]
twiggler: We agree that while the current solution is  adequate for the short-to mid term to satisfy user demands for single file capability, it is not flexible enough to accommodate other plugins which process multiple types of files. 

Therefore, we have decided to make architectural changes, such as moving the parsers out of the plugin, before supporting single files in the general case. The explicit switch in the plugin on "direct-mode" is acceptable for now.

dissect/target/target.py
```diff
@@ -70,7 +71,7 @@ class Target:
         path: The path of a target.
     """
 
-    def __init__(self, path: str | Path | None = None):
+    def __init__(self, path: str | Path | None = None, minimal: bool = False):
```
Schamper, line 74:
```suggestion
    def __init__(self, path: str | Path | None = None):
```
This is no longer necessary.

tests/test_plugin.py
```diff
@@ -1219,8 +1219,8 @@ def test_exported_plugin_format(descriptor: FunctionDescriptor) -> None:
     Resources:
         - https://docs.dissect.tools/en/latest/contributing/style-guide.html
     """
-    # Ignore DefaultPlugin and NamespacePlugin instances
-    if descriptor.cls.__base__ is NamespacePlugin or descriptor.cls is DefaultPlugin:
+    # Ignore DefaulOStPlugin and NamespacePlugin instances
```
Schamper, line 1222:
```suggestion
    # Ignore DefaultOSPlugin and NamespacePlugin instances
```
I made a typo.

tests/test_target.py
```diff
@@ -322,7 +322,7 @@ def test_target_win_fallback_dirfs(mocked_win_volumes_fs: tuple[Mock, Mock, Mock
     # both volumes are present
     assert len(target.volumes) == 2
 
-    # FSs from both volumes, bad and good, are mounted + 2 mounts by DefaultPlugin
+    # FSs from both volumes, bad and good, are mounted + 2 mounts by DefaulOStPlugin
```
Schamper, line 325:
```suggestion
    # FSs from both volumes, bad and good, are mounted + 2 mounts by DefaultOSPlugin
```

Schamper (changes requested):

Schamper (approved):

PR #1084, opened 2025-04-01
Add Windows MSN plugin

JSCU-CNI:
This PR adds support to dissect for the Windows Live Messenger application. Currently works on `WLM 2009` and `MSN 7.5` (on Windows 7 and XP), other versions might work too but have not been tested. Artifacts from `Messenger Plus! Live` are currently not supported.

### DIFF
--- a/dissect/target/plugins/apps/chat/chat.py
+++ b/dissect/target/plugins/apps/chat/chat.py
@@ -1,0 +1,51 @@
+from typing import Union
+
+from dissect.target.helpers.descriptor_extensions import UserRecordDescriptorExtension
+from dissect.target.helpers.record import create_extended_descriptor
+from dissect.target.plugin import NamespacePlugin
+
+COMMON_FIELDS = [
+    ("datetime", "ts"),
+    ("string", "client"),
+    ("string", "account"),
+    ("string", "sender"),
+    ("string", "recipient"),
+]
+
+GENERIC_USER_FIELDS = [
+    ("datetime", "ts_mtime"),
+    ("string", "client"),
+    ("string", "account"),
+]
+
+GENERIC_ATTACHMENT_FIELDS = [
+    *COMMON_FIELDS,
+    ("path", "attachment"),
+    ("string", "description"),
+]
+
+GENERIC_MESSAGE_FIELDS = [
+    *COMMON_FIELDS,
+    ("string", "message"),
+]
+
+ChatUserRecord = create_extended_descriptor([UserRecordDescriptorExtension])(
+    "chat/user",
+    GENERIC_USER_FIELDS,
+)
+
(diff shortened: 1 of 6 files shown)


dissect/target/plugins/apps/chat/chat.py
Schamper, line 1: Don't forget the `__init__.py`.

Also, `messaging` vs `chat`? I assume we will want to use this namespace for applications such as WhatsApp, Signal, IRC, Slack, Teams, etc as well. I think `chat` is nice and short but I'm not sure what the most commonly used nomenclature would be for these apps.

dissect/target/plugins/apps/chat/msn.py
```diff
@@ -0,0 +1,116 @@
...
+                    if profile.is_dir():
+                        yield user_details, profile
+
+    def check_compatible(self) -> None:
+        if not self.installs:
+            raise UnsupportedPluginError("No Microsoft MSN installs found on target")
+
+    @export(record=[ChatMessageRecord, ChatAttachmentRecord])
+    def history(self) -> Iterator[ChatMessageRecord | ChatAttachmentRecord]:
+        """Yield MSN chat history messages
```
Schamper, line 57:
```suggestion
        """Yield MSN chat history messages.
```

dissect/target/plugins/apps/chat/msn.py
```diff
@@ -0,0 +1,116 @@
...
+                                    yield ChatAttachmentRecord(
+                                        **common,
+                                        recipient=None,  # unknown with Invitations
+                                        attachment=file.text,
+                                        description=entry.find(".//Text").text,
+                                    )
+
+
+def convert_email(string: str) -> int:
+    """Convert MSN email address to 10 digit Passport ID"""
```
Schamper, line 111:
```suggestion
    """Convert MSN email address to 10 digit Passport ID."""
```
Can you also add a docstring what this is used for? From the test file I gather it's for the history path.

dissect/target/plugins/os/windows/_os.py
```diff
@@ -311,6 +311,12 @@ def users(self) -> Iterator[WindowsUserRecord]:
                     home = profile_image_path.value
                     name = home.split("\\")[-1]
 
+                # Windows XP uses %variables% in home paths
```
Schamper, line 314: Can `t.resolve` fix this?

Schamper (changes requested):
:buzzer:

dissect/target/plugins/apps/chat/chat.py
Schamper: Let's leave it at `chat` until some zealot comes by and has a strong opinion on it.

Schamper (approved):

PR #1085, opened 2025-04-01
Improve detection for iOS targets

JSCU-CNI:
This PR improves the detection for iOS targets in dissect. The following OS functions should now work on iOS:
1. `users`
2. `architecture`
3. `hostname`
4. `version`
5. `activity`
6. `install_date`
7. `timezone`
8. `language`

We have also added an `applications` plugin.


dissect/target/plugins/os/unix/_os.py
```diff
@@ -38,6 +38,10 @@
 
 
 class UnixPlugin(OSPlugin):
+    """"""
```
Schamper, line 41: Empty docstring?

dissect/target/plugins/os/unix/bsd/darwin/_os.py
```diff
@@ -0,0 +1,55 @@
...
+ARCH_MAP = {
+    b"\x0c\x00\x00\x01": "arm64",  # big endian, x64
+    b"\x01\x00\x00\x0c": "arm64",  # little endian, x64
+    b"\x0c\x00\x00\x00": "arm32",  # big endian, x32
+    b"\x00\x00\x00\x0c": "arm32",  # little endian, x32
+}
+
+
+class DarwinPlugin(BsdPlugin):
+    """"""
```
Schamper, line 17: Empty docstring?

dissect/target/plugins/os/unix/bsd/darwin/_os.py
```diff
@@ -0,0 +1,55 @@
+from pathlib import Path
```
Schamper, line 1:
```suggestion
from __future__ import annotations

from pathlib import Path
```

dissect/target/plugins/os/unix/bsd/darwin/ios/_os.py
```diff
@@ -0,0 +1,81 @@
+import plistlib
```
Schamper, line 1:
```suggestion
from __future__ import annotations

import plistlib
```

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
Schamper, line 1: Empty test file?

tests/plugins/os/unix/bsd/darwin/ios/test_applications.py
Schamper, line 1: Empty test file?

dissect/target/plugins/os/unix/bsd/darwin/_os.py
```diff
@@ -0,0 +1,55 @@
...
+# https://en.wikipedia.org/wiki/Mach-O
+ARCH_MAP = {
+    b"\x0c\x00\x00\x01": "arm64",  # big endian, x64
+    b"\x01\x00\x00\x0c": "arm64",  # little endian, x64
+    b"\x0c\x00\x00\x00": "arm32",  # big endian, x32
+    b"\x00\x00\x00\x0c": "arm32",  # little endian, x32
+}
+
+
+class DarwinPlugin(BsdPlugin):
```
Schamper, line 16: Does the OS detection from a `Target.open` still work if this has no `detect`?

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/ios/_os.py
```diff
@@ -1,46 +1,124 @@
...
+class Config:
+    SYSTEM: dict
+    GLOBAL: dict
+    VERSION: dict
+
+    def __post_init__(self):
+        for field in self.__dataclass_fields__.keys():
+            path = getattr(self, field)
+            if path.is_file():
+                setattr(self, field, plistlib.load(path.open("rb")))
```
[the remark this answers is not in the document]
Schamper: Depends a little bit if you want to merge everything into a single dictionary or not. If yes, then it doesn't really matter. If not (e.g. go the `config["SYSTEM"]` route or whatever) you could use something like this:

```python
class Config(TypedDict):
    SYSTEM: dict
    GLOBAL: dict
    VERSION: dict
```

Then you get autocompletion for the first level of the dictionary. Everything after that is unknown, of course.

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
```diff
@@ -0,0 +1,46 @@
...
+    target_ios.add_plugin(IOSPlugin)
+    target_ios.add_plugin(LocalePlugin)
+    target_ios.add_plugin(IOSApplicationsPlugin)
+
+    assert target_ios.os == OperatingSystem.IOS
+    assert target_ios.version == "iPhone OS 17.3 (21D50)"
+    assert target_ios.architecture == "arm64-ios"
+    assert target_ios.hostname == "This-Iss-iPhone"
+    assert target_ios.timezone == "America/New_York"
+    # assert target_ios.language == ["en_US"]  # unable to test this in pytest for some reason
```
Schamper, line 31: What's wrong?

dissect/target/plugins/os/unix/bsd/darwin/ios/_os.py
```diff
@@ -0,0 +1,98 @@
...
+from dissect.target.filesystem import Filesystem, VirtualFilesystem
+from dissect.target.helpers.record import COMMON_UNIX_FIELDS, TargetRecordDescriptor
+from dissect.target.plugin import OperatingSystem, export
+from dissect.target.plugins.os.unix.bsd.darwin._os import (
+    DarwinPlugin,
+    detect_macho_arch,
+)
+from dissect.target.target import Target
+
+IOSUserRecord = TargetRecordDescriptor(
```
Schamper, line 16: Why are the user records for iOS and macOS defined in their OS plugin files, but the application records are defined in `record.py`?

I think record definitions should primarily be in the file they are used in (so not `record.py`). I think originally the user records were defined in that file because they were used in a bunch of places and very high level OS stuff. I'm not sure why the application records are in there now, can't those be moved out?

The user records can maybe still go in there, it's convenient for example for `plugins/general/users.py` (which needs a type update for the new user records too).

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
```diff
@@ -0,0 +1,46 @@
...
+from dissect.target.plugins.os.unix.bsd.darwin.ios._os import IOSPlugin
+from dissect.target.plugins.os.unix.bsd.darwin.ios.applications import (
+    IOSApplicationsPlugin,
+)
+from dissect.target.plugins.os.unix.bsd.darwin.ios.locale import LocalePlugin
+from dissect.target.target import Target
+from tests._utils import absolute_path
+
+
+def test_unix_bsd_darwin_ios_os(target_ios: Target, fs_ios: VirtualFilesystem) -> None:
```
Schamper, line 12:
```suggestion
def test_ios_os(target_ios: Target, fs_ios: VirtualFilesystem) -> None:
```

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
```diff
@@ -0,0 +1,46 @@
...
+    """Test if we correctly detect an iOS target.
```
Schamper, line 13: It doesn't test detection since you manually add the `IOSPlugin`, only tests some plugin functionality now.

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/darwin/ios/_os.py
```diff
@@ -0,0 +1,95 @@
...
+    GLOBAL: dict[str, Any]
+    VERSION: dict[str, Any]
+
+    def __post_init__(self):
+        for field in self.__dataclass_fields__.keys():
+            path = getattr(self, field)
+            if path.is_file():
+                setattr(self, field, plistlib.load(path.open("rb")))
+            else:
+                setattr(self, field, {})
```
Schamper, line 95: This didn't change the problem that `Config` will seem to have an `__init__` that looks like this:

```python
class Config
    def __init__(self, SYSTEM: dict[str, Any], GLOBAL: dict[str, Any], VERSION: dict[str, Any]): ...
```

And then you initialize it with 3 `TargetPath` objects instead. You fix it at runtime in `__post_init__`, but I want to prevent we merge something like this now and have to completely change it later when we will apply stronger type hinting to this project.

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
```diff
@@ -0,0 +1,53 @@
...
+from dissect.target.plugins.os.unix.bsd.darwin.ios.locale import LocalePlugin
+from dissect.target.target import Target
+from tests._utils import absolute_path
+
+
+def test_ios_detect(target_bare: Target, fs_ios: VirtualFilesystem) -> None:
+    """Test if we correctly detect an iOS target."""
+    target_bare.filesystems.add(fs_ios)
+    target_bare.apply()
+    assert IOSPlugin.detect(target_bare)
```
Schamper, line 16: I think it's better to assert that `target_bare._os_plugin` is correct.

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
Schamper: I'll take a look as well later.

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/darwin/ios/_os.py
Schamper: That'll also be a huge pain with a type checker, since it will always complain when you try to access an attribute on `dict` that doesn't exist on `Path`, you'd need to assert it was a `dict` first.

Maybe better to use a `classmethod` initializer like `Config.load(path1, path2, path2)`.

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
```diff
@@ -0,0 +1,54 @@
...
+    Data based on example iOS image from Digital Corpora.
+
+    Resources:
+        - https://corp.digitalcorpora.org/corpora/mobile/iOS17/
+    """
+
+    fs_ios.map_dir("/", absolute_path("_data/plugins/os/unix/bsd/darwin/ios/_os"))
+    target_ios.add_plugin(IOSPlugin)
+    target_ios.add_plugin(LocalePlugin)
+    target_ios.add_plugin(IOSApplicationsPlugin)
```
Schamper, line 32:
```suggestion
    target_ios.add_plugin(IOSApplicationsPlugin)
    target_ios.apply()
```
This properly initializes the OS portion of the target and allows the `.language` to work.

tests/plugins/os/unix/bsd/darwin/ios/test__os.py
Schamper: The new suggestion fixes it. Do note that I get `en-US` as output, should that be normalized?

Schamper (changes requested):

Schamper (changes requested):
Can you fix the tests? Looks like a missing `from __future__ import annotations` somewhere.

Schamper (approved):

PR #1089, opened 2025-04-02
Add `pytest-xdist` to `tox` and run multi-threaded

JSCU-CNI:
This PR significantly improves the total time the tests take to complete on most systems. Limited to four threads to keep CI/CD workers happy.

Before on `x86_64`:
```shell
$ tox -e py313
...
2028 passed, 1 skipped in 120.77s (0:02:00)
```

After on `x86_64`:
```
$ tox -e py313
...
2028 passed, 1 skipped, 1 warning in 53.27s
```


tox.ini
```diff
@@ -16,10 +16,11 @@ extras = dev
 deps =
     pytest
     pytest-cov
+    pytest-xdist
     coverage
     docutils
 commands =
-    pytest --basetemp="{envtmpdir}" {posargs:--color=yes --cov=dissect --cov-report=term-missing -v tests}
+    pytest --basetemp="{envtmpdir}" {posargs:--color=yes --cov=dissect --cov-report=term-missing -n 4 -v tests}
```
Schamper, line 23: Would `-n auto` also keep CI/CD happy? GitHub appears to have 4 cores: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Schamper: Looks like you can set `PYTEST_XDIST_AUTO_NUM_WORKERS`: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

Schamper (approved):
Doesn't seem all that much faster on GitHub but it's a nice local improvement 😂 I'd actually been running this locally for a bit already.

PR #1090, opened 2025-04-03
Target reload

twiggler:
Add "reload" command to target-shell, which helps when running against a live target.

### DIFF
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -450,7 +450,7 @@ class Target:
     def reload(self) -> Target:
         """Reload the current target.
 
-        Using the loaded with which the target was originally loaded,
+        Using the loader with which the target was originally loaded,
         reload the path of the current target.
         This is useful when the target is live.
 
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -447,6 +447,25 @@ class Target:
         for child_plugin in self._child_plugins.values():
             yield from child_plugin.list_children()
 
+    def reload(self) -> Target:
+        """Reload the current target.
+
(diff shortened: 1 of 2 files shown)


dissect/target/tools/shell.py
```diff
@@ -719,6 +719,12 @@ def do_info(self, line: str) -> bool:
         print_target_info(self.target)
         return False
 
+    def do_reload(self, line: str) -> bool:
+        """reload the target"""
+        self.target = self.target.reload()
+        self.chdir(str(self.cwd))  # self.cwd has reference into the old target :/
```
Schamper, line 725: The typing on `self.cwd` looks to be wrong, I know we've talked about this many times but would you mind fixing it up a tiny bit? (Mostly just changing it from `str` to `Path`).

Schamper (changes requested):

dissect/target/tools/shell.py
Schamper: Yes, sorry.

dissect/target/target.py
```diff
@@ -447,6 +447,25 @@ def list_children(self) -> Iterator[ChildTargetRecord]:
         for child_plugin in self._child_plugins.values():
             yield from child_plugin.list_children()
 
+    def reload(self) -> Target:
+        """Reload the current target.
+
+        Using the loaded with which the target was originally loaded,
```
Schamper, line 453:
```suggestion
        Using the loader with which the target was originally loaded,
```

Schamper (approved):

PR #1094, opened 2025-04-07
Mount NFS shares on ESXi

twiggler:
Also:

- do not set KEEP_ALIVE on MacOS  because it is unsupported.


dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -42,6 +46,8 @@
     ],
 )
 
+log = logging.getLogger(__name__)
```
Schamper, line 49: I don't think I see this used?

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -190,6 +199,53 @@ def configstore(self) -> dict[str, Any]:
...
 
+    def _mount_nfs_shares(self) -> None:
+        # Mount NFS shares
+        if not self._configstore:
+            self.target.log.warning("No configstore found")
+            return
+
+        nfs_shares: dict[str, Any] = self._configstore.get("esx", {}).get("storage", {}).get("nfs_v3_datastores", {})
+        if not nfs_shares:
+            self.target.log.info("No NFS shares found in datastore")
```
Schamper, line 210: Shouldn't we check if NFS mounting is desired before logging a bunch of things?

Schamper (changes requested):

dissect/target/helpers/esxi/hash.py
```diff
@@ -0,0 +1,57 @@
+def mix64(a: int, b: int, c: int):
+    """
+    Mixes three 64-bit values reversibly.
+    """
```
Schamper, line 4:
```suggestion
    """Mixes three 64-bit values reversibly."""
```

dissect/target/helpers/esxi/hash.py
```diff
@@ -0,0 +1,57 @@
+def mix64(a: int, b: int, c: int):
```
Schamper, line 1:
```suggestion
def mix64(a: int, b: int, c: int) -> int:
```

dissect/target/helpers/esxi/hash.py
```diff
@@ -0,0 +1,57 @@
...
+    a = (a - b - c) ^ ((c & 0xFFFFFFFFFFFFFFFF) >> 12)
+    b = (b - c - a) ^ (a << 18)
+    c = (c - a - b) ^ ((b & 0xFFFFFFFFFFFFFFFF) >> 22)
+    # Normalize to 64 bits
+    return a & 0xFFFFFFFFFFFFFFFF, b & 0xFFFFFFFFFFFFFFFF, c & 0xFFFFFFFFFFFFFFFF
+
+
+def hash(key: bytes, level: int) -> int:
+    """
+    Hashes a variable-length key into a 64-bit value.
```
Schamper, line 24:
```suggestion
    """Hashes a variable-length key into a 64-bit value.
```

dissect/target/plugins/os/unix/_os.py
```diff
@@ -294,10 +286,33 @@ def _add_mounts(self) -> None:
...
     def _add_nfs(self, address: str, exported_dir: str, mount_point: str) -> None:
+        if not self._check_nfs_enabled(address, exported_dir, mount_point):
+            return
+
         # Try all users to see if we can access the share
         def auth_setter(nfs_client: NfsClient, filehandle: FileHandle, _: list[int]) -> None:
-            for user in self.users():
+            users = list(self.users())
+            if not users:
+                self.target.log.debug("No users found, trying root")
```
Schamper, line 312:
```suggestion
                self.target.log.debug("No users found, trying root for mounting NFS share %s:%s at %s", address, exported_dir, mount_point)
```

dissect/target/plugins/os/unix/_os.py
```diff
@@ -314,6 +329,7 @@ def auth_setter(nfs_client: NfsClient, filehandle: FileHandle, _: list[int]) ->
                         nfs_client.close()
                         raise e
 
+            self.target.log.debug("No user has access to NFS share")
             raise FilesystemError("No user has access to NFS share")
```
Schamper, line 333:
```suggestion
            self.target.log.debug("No user has access to NFS share %s:%s at %s", address, exported_dir, mount_point)
            raise FilesystemError("No user has access to NFS share %s:%s at %s", address, exported_dir, mount_point)
```

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -190,6 +197,59 @@ def configstore(self) -> dict[str, Any]:
     def os(self) -> str:
         return OperatingSystem.ESXI.value
 
+    def _mount_nfs_shares(self) -> None:
+        # Mount NFS shares
+        if not self._configstore:
+            self.target.log.warning("No configstore found")
```
Schamper, line 203:
```suggestion
            self.target.log.warning("Unable to mount NFS shares, no configstore found")
```

dissect/target/helpers/esxi/hash.py
Schamper, line 1: Maybe this belongs more in `dissect.hypervisor`? Like the envelope implementations too.

Schamper (changes requested):

dissect/target/plugins/os/unix/esxi/_os.py
twiggler: From discussions with the analysts I gathered they want to be informed when there are nfs stores, even with nfs disabled. (Although strictly speaking, this message states there are *no*  nfs stores :)). 

Maybe debug log level is more appropriate in this case?

Schamper: Nitpicky, but shouldn't that then be a log in `acquire` and not `dissect.target`?

dissect/target/helpers/esxi/hash.py
Schamper: There probably are other use-cases elsewhere, but we don't have those yet so it can also be seen as a premature optimization. Probably fine to keep it here for the time being, or just roll it into `esxi/_os.py` anyway despite it already being quite big.

twiggler: Will roll into `esxi/_os` then :+1:

dissect/target/plugins/os/unix/esxi/_os.py
twiggler: Technically this code could also be invoked form target-shell or target-query in combination with the  local loader. 

I do think the warnings should be easier to see when running acquire, will spar a bit with Michaël tomorrow

twiggler: No problemo

Schamper (changes requested):

dissect/target/plugins/os/unix/_os.py
```diff
@@ -294,10 +286,31 @@ def _add_mounts(self) -> None:
                     self.target.log.debug("Mounting %s (%s) at %s", fs, fs.volume, mount_point)
                     self.target.fs.mount(mount_point, fs)
 
+    @property
+    def is_nfs_enabled(self) -> bool:
+        return isinstance(self.target._loader, LocalLoader) and "enable-nfs" in self.target.path_query
+
+    def _log_nfs_mount_disabled(self, address: str, exported_dir: str, mount_point: str) -> None:
+        log.warning(
+            "NFS mount %s:%s at %s is disabled. To enable, pass --enable-nfs to the local loader. Alternatively, add a query parameter to the target query string: local?enable-nfs",  # noqa: E501
```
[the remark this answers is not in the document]
twiggler: Perhaps it is a bit implicit that nfs is  disabled when not using the local loader.

We could rephrase as:
```
 To enable, use the local loader and pass --enable-nfs. Alternatively....
 ```
 
 If it rarely happens that the user has access to the live disk when first analyzing the dead disk,
 we could also choose to only show the warning when using the local loader.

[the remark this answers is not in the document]
Comment (Schamper) [reply dissect/target/plugins/os/unix/_os.py:295]: I'd rather there be two different log messages:
- One for the case where we're on a local loader but no `--enable-nfs`
  - The current message is good for this
- One for the case where we're not on a local loader
  - This message should say something along the lines of "NFS mount %s:%s at %s is unavailable on a non-local target"

File: dissect/target/plugins/os/unix/_os.py [hunk 10]
```diff
@@ -314,7 +337,8 @@ def auth_setter(nfs_client: NfsClient, filehandle: FileHandle, _: list[int]) ->
                         nfs_client.close()
                         raise e
 
-            raise FilesystemError("No user has access to NFS share")
+            self.target.log.debug("No user has access to NFS share %s:%s at %s", address, exported_dir, mount_point)
+            raise FilesystemError("No user has access to NFS share %s:%s at %s", address, exported_dir, mount_point)
```
Schamper, line 341:
```suggestion
            raise FilesystemError(f"No user has access to NFS share {address}:{exported_dir} at {mount_point}")
```

Schamper (changes requested):

Schamper (changes requested):

Issue #1087, opened 2025-04-01
target-query arguments are not passed to namespace plugin invocations


JSCU-CNI:
The `generate_argparse_for_plugin_class` function does not parse `__args__` of it's exported functions. It probably should, just like `generate_argparse_for_unbound_method`. Perhaps it should be possible to register plugin-wide arguments in a `NamespacePlugin`, or the arguments should be derived at runtime when parsing the plugin class.

To reproduce, add an `@arg()` decorator to for example the `FirefoxPlugin.history` function and see the difference between `target-query -f firefox.history --argument` and `target-query -f firefox --argument`. I would expect both invocations to work.

Schamper:
Note for implementation details: you'll need to separate the arguments into those that the namespace function actually accepts. E.g. not every function in the namespace will accept a random `--argument` or `argument=blah` kwarg.

PR #1101, opened 2025-04-14
Allow namespaces to accept arguments

Schamper:
Fixes `-f mft --compact`


Schamper:
> Does this PR address #1087 as well?

Only in the sense that the arguments will reach the namespace plugin in the first place, not forwarding them to their sub-functions just yet.

twiggler (approved):

PR #1102, opened 2025-04-14
Fix multiple namespace plugins in same file

### LINKED ISSUE #974: Multiple plugins inside same file not registered
Given the following setup:

```
dissect/target/plugins/apps/nsexample/
├── foobar.py
├── __init__.py
└── nsexample.py
```

\\*\***nsexample/nsexample.py** **\**
```python
from dissect.target.plugin import NamespacePlugin


class ExampleNamespacePlugin(NamespacePlugin):
    """example namespace plugin"""
    __namespace__ = "nsexample"
```

\\*\***nsexample/foobar.py** **\**
```python
from typing import Iterator
from dissect.target.plugin import export
from dissect.target.plugins.apps.nsexample.nsexample import ExampleNamespacePlugin


class FooPlugin(ExampleNamespacePlugin):
    """Foo plugin."""

    __namespace__ = "foo"

    @export(output="yield")
    def hello(self) -> Iterator[str]:
        """hello from foo.example"""
        yield "hello from foo.example"


class BarPlugin(ExampleNamespacePlugin):
    """Bar plugin."""

    __namespace__ = "bar"

    @export(output="yield")
    def hello(self) -> Iterator[str]:
        """hello from bar.example"""
        yield "hello from bar.example"
```

We would expect both `foo.hello` and `bar.hello` to be registered as plugins. However it seems only `bar.hello` is registered. the namespace plugin `nsexample.hello` does seem to pick up `foo.hello`.

```
$ target-query -l | grep hello
bar.hello - hello from bar.example (output: lines)
nsexample.hello - Return hello for: foo,bar (output: records)
```

Schamper:
Fixes #974.

The plugins were registered correctly (you could execute them fine), but weren't listed correctly.

Will add unit test later.


Schamper:
I can't reproduce what you say about running the namespace only executing the last one, works fine for me:

```python
from dissect.target.plugin import NamespacePlugin, export


class NS(NamespacePlugin):
    __namespace__ = "ns"

    def check_compatible(self):
        return None


class Foo(NS):
    __namespace__ = "foo"

    @export(output="yield")
    def baz(self):
        yield from ["foo"]


class Bar(NS):
    __namespace__ = "bar"

    @export(output="yield")
    def baz(self):
        yield from ["bar"]
```

```bash
~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f foo.baz -q
foo

~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f bar.baz -q
bar

~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f foo -q    
'foo'

~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f bar -q    
'bar'

~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f ns.baz -q
foo
bar

~/Shares/src/dissect/dissect.target ❯ uv run target-query ~/Parallels/VPN.pvm -f ns -q    
'foo'
'bar'
```

The `repr` behavior looks weird, but that's probably some other issue 😄.

I can reproduce the `*.baz` behavior so I will fix that.

PR #1103, opened 2025-04-14
Add record output and regex needle support to qfind

JSCU-CNI:
This PR is a draft for some features we would like to add to the qfind plugin. Before we continue we'd like to check if our chosen implementation is desired or if we should continue in another way. Since qfind can be invoked using both `target-query -f qfind` and `target-qfind` modifying the plugin can be a bit tricky.

@Schamper Could you give us your thoughts on the added comment lines?


dissect/target/helpers/scrape.py
```diff
@@ -60,7 +60,14 @@ def find_needles(
             offset = fh.tell()
 
         read_size = min(block_size, end - offset) if end is not None else block_size
-        if not (next_block := fh.read(read_size)):
+        next_block = None
+
+        try:
+            next_block = fh.read(read_size)
```
Schamper, line 66: A normally behaving `fh.read()` shouldn't be able to raise `EOFError` but should return `b""` instead, can you explain or otherwise add a unit test for when this could happen?

dissect/target/plugins/scrape/qfind.py
```diff
@@ -25,27 +36,32 @@ def check_compatible(self) -> None:
...
         ignore_case: bool = False,
         allow_non_ascii: bool = False,
         unique: bool = False,
         window: int = 256,
-    ) -> None:
+    ) -> Iterator[QFindHitRecord] | None:
         """Find a needle in a haystack.
 
+        Needles starting with a ``#`` are ignored in needle files.
+        Hex encode such needles to include them in the needle stack.
```
Schamper, line 63: Took me a minute to understand that you meant to hex encode needles starting with `#`, can you rephrase?

dissect/target/plugins/scrape/qfind.py
```diff
@@ -109,9 +133,21 @@ def qfind(
                 tmp[re.compile(re.escape(encoded_needle), re.IGNORECASE)] = _
             needle_lookup = tmp
 
+        if regex:
+            tmp = {}
+            for needle, _ in needle_lookup.items():
+                tmp[re.compile(needle, re.IGNORECASE if ignore_case else re.NOFLAG)] = _
```
Schamper, line 139: Does this work properly if the `ignore_case` branch just above this has already compiled the needle into a regex pattern? I get an error when trying to compile an already compiled pattern.

dissect/target/plugins/scrape/qfind.py
```diff
@@ -109,9 +133,21 @@ def qfind(
...
+            needle_lookup = tmp
+
+        # Output records if invoked through target-query only.
+        # TODO: Find a more suitable method to derive invocation method.
+        record = False
+        if "target-query" in sys.argv[0]:
+            record = True
```
Schamper, line 146: I had thought about this use-case while writing this plugin but wanted to get this working normally first and then worry about it later 😄.

My initial gut feeling is to make the plugin be record-yielding first and foremost, and move all the "graphical" stuff to the `target-qfind` `main`. You could deal with the progress bar by just having a `progress=True` keyword argument that is not exposed with an argument.

Another option is to make most of this method a `_qfind` method in the plugin, and have a small wrapping `qfind` record `@export`, and an `@internal` `qfind_ext`. Where you would use `qfind_ext` in `target-qfind`.

I think I like the first option a bit better, but it's good to have an open discussion about it to exchange ideas.

dissect/target/plugins/scrape/qfind.py
```diff
@@ -125,8 +161,25 @@ def qfind(
                 print(header)
                 palette = [(offset - before_offset, utils.COLOR_NORMAL), (needle_len, utils.COLOR_BG_RED)]
                 utils.hexdump(buf, palette, offset=before_offset)
+
+            elif record:
+                if unique:
+                    if buf in seen:
+                        continue
+                    seen.add(buf)  # TODO: store digest of buffer instead, to reduce memory usage
```
Schamper, line 169: Good idea.

dissect/target/plugins/scrape/qfind.py
```diff
@@ -125,8 +161,25 @@ def qfind(
...
+
+                yield QFindHitRecord(
+                    offset=offset,
+                    needle=original_needle,
+                    codec=codec,
+                    content=buf.hex(),
```
Schamper, line 175: There exists a `bytes` type in `flow.record` which may be better suited.

dissect/target/plugins/scrape/qfind.py
```diff
@@ -125,8 +161,25 @@ def qfind(
...
+                    _target=self.target,
+                )
+
             else:
                 codec = "utf-8" if codec == "hex" else codec
+
+                # TODO: needle_len is incorrect with regex matches and causes invalid color lengths
```
Schamper, line 182: You should probably use the match span in case of a regex needle then.

Schamper (changes requested):

dissect/target/helpers/scrape.py
Schamper: Sounds more like that is something that should be fixed 😄. Is there publicly available data or instructions on how to create reproducable data?

Schamper: Can you undo this change in the meantime?

dissect/target/plugins/scrape/qfind.py
```diff
@@ -102,68 +136,75 @@ def qfind(
...
+                offset=offset,
+                needle=original_needle,
+                codec=codec,
+                match=match,
+                content=buf.strip(b"\x00") if strip_null_bytes else buf,
+                _target=self.target if self.target._os else None,
+            )
+
+
+def progress_handler(target: Target) -> Callable[[Container | Volume, int, int], None]:
```
Schamper, line 188: I'd move this to `tools/qfind.py` and make the `progress` kwarg a `Callable[[Container | Volume, int, int], None]`.

dissect/target/tools/qfind.py
```diff
@@ -40,19 +53,69 @@ def main() -> int:
         log.error("No targets provided")
         return 1
 
+    if args.record or args.json:
+        RECORD_OUTPUT = True
+        rs = record_output(args.strings, args.json)
```
Schamper, line 58: Rather initialize `rs = None` and do that `if` later based on that.

dissect/target/helpers/scrape.py
```diff
@@ -74,22 +74,22 @@ def find_needles(
         last_needle_pos = -1
         last_needle_end = 0
         while True:
-            needle_positions = [
-                (pos, needle)
-                for needle in needles
+            needle_positions = []
```
Schamper, line 77: On another note (but a bigger ask), maybe it's interesting to write some benchmarks for this piece like we did here: https://github.com/fox-it/dissect.util/pull/64

Schamper (changes requested):

dissect/target/plugins/scrape/qfind.py
```diff
@@ -102,68 +136,75 @@ def qfind(
...
+
+            match = match.group() if isinstance(needle, re.Pattern) and match else original_needle
+
+            yield QFindMatchRecord(
+                offset=offset,
+                needle=original_needle,
+                codec=codec,
+                match=match,
+                content=buf.strip(b"\x00") if strip_null_bytes else buf,
+                _target=self.target if self.target._os else None,
```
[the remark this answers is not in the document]
Schamper: Sounds like we need to fix that then 😉.

https://github.com/fox-it/dissect.target/blob/bf3860a417ac9c8de2ba61c7470134ae574e7190/dissect/target/helpers/descriptor_extensions.py#L82

dissect/target/helpers/scrape.py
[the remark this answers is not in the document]
Schamper: In some micro benchmarking the list comprehension seems slower now. I thought it was faster when I tested it initially, so let's just drop this comment.

dissect/target/tools/qfind.py
Schamper: Why would you need a record stream for each target? Your current code also only has a single record stream.

Schamper (changes requested):

dissect/target/plugins/scrape/qfind.py
```diff
@@ -102,68 +129,46 @@ def qfind(
...
-
-    return update
+                if digest in seen:
+                    continue
+                seen.add(digest)
+
+            match = match.group() if match else original_needle
+
+            yield QFindMatchRecord(
+                disk=disk,
```
Schamper, line 167:
```suggestion
                disk=repr(disk),
```
To make it a bit more explicit what's happening.

dissect/target/plugins/scrape/qfind.py
```diff
@@ -102,68 +129,46 @@ def qfind(
...
+        if ignore_case and not regex:
             tmp = {}
             for encoded_needle, _ in needle_lookup.items():
                 tmp[re.compile(re.escape(encoded_needle), re.IGNORECASE)] = _
             needle_lookup = tmp
 
+        elif regex:
+            tmp = {}
+            for needle, _ in needle_lookup.items():
+                tmp[re.compile(needle, re.IGNORECASE if ignore_case else re_NOFLAG)] = _
```
Schamper, line 144:
```suggestion
            for encoded_needle, _ in needle_lookup.items():
                tmp[re.compile(encoded_needle, re.IGNORECASE if ignore_case else re_NOFLAG)] = _
```

dissect/target/plugins/scrape/qfind.py
```diff
@@ -1,21 +1,32 @@
...
+re_NOFLAG = 0  # re.NOFLAG is Python 3.11 and newer only
+
+QFindMatchRecord = TargetRecordDescriptor(
+    "qfind/match",
+    [
+        ("string", "disk"),
+        ("varint", "offset"),
+        ("string", "needle"),
+        ("string", "codec"),
+        ("string", "match"),
```
Schamper, line 26:
```suggestion
        ("bytes", "match"),
```
I think this should be bytes too?

Schamper (changes requested):

dissect/target/plugins/scrape/qfind.py
```diff
@@ -102,68 +132,46 @@ def qfind(
...
             tmp = {}
             for encoded_needle, _ in needle_lookup.items():
                 tmp[re.compile(re.escape(encoded_needle), re.IGNORECASE)] = _
             needle_lookup = tmp
 
+        elif regex:
+            tmp = {}
+            for encoded_needle, _ in needle_lookup.items():
+                tmp[re.compile(encoded_needle, re.IGNORECASE if ignore_case else re_NOFLAG)] = _
+            needle_lookup = tmp
```
Schamper, line 148:
```suggestion
        if ignore_case or regex:
            tmp = {}
            for encoded_needle, _ in needle_lookup.items():
                encoded_needle = encoded_needle if regex else re.escape(encoded_needle)
                tmp[re.compile(encoded_needle, re.IGNORECASE if ignore_case else re_NOFLAG)] = _
            needle_lookup = tmp
```
I think we can roll these two loops into one.

Schamper (changes requested):

dissect/target/tools/qfind.py
```diff
@@ -40,27 +64,270 @@ def main() -> int:
...
+                    palette = (
+                        [(hit.offset - before_offset, utils.COLOR_NORMAL), (needle_len, utils.COLOR_BG_RED)]
+                        if not NO_COLOR
+                        else None
+                    )
+                    utils.hexdump(hit.content, palette, offset=before_offset)
+
+                else:
+                    codec = "utf-8" if hit.codec == "hex" else hit.codec
+                    before_part = recover_string(
```
Schamper, line 110: Would this "recovering" of the string not be nice to have in the plugin too?

Schamper (changes requested):
Apologies for the delay, thought I had already finished this. Had a small thought while looking at it again though.

Schamper (changes requested):
If you address the unit test failure this is good to go.

Schamper:
The match buffers are different because the first hit is at the beginning of the "disk":
```python
ipdb> p buf
b'ABCD\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```
And the second hit has a "window" before it:
```python
ipdb> p buf
b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00ABCD\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

Probably the test data requires some changing to add some null bytes to the beginning, so that both buffers have the same "window" around it?

PR #1106, opened 2025-04-16
Fix bug in kape loader when using with target-mount

william-billaud:
When target mount is used with a relative path ressource (eg ./kape), mount point will be corrupted as fuse process is not aware of the local path.
Also fix tanium loader as tanium loader is similar to kape.

eg;

```
python3 -m dissect.target.tools.mount ../../../WORK/TEST/evidence/kape/target mnt
❯ ls mnt/fs/sysvol
ls: cannot access 'mnt/fs/sysvol': Input/output error
❯ tree -a mnt
mnt
├── filesystems
│   └── fs_1
└── fs
    └── $fs$
        └── fs0
```

Expect result is 

```
python3 -m dissect.target.tools.mount /WORK/TEST/evidence/kape/target mnt
❯ ls mnt/fs/sysvol
'$Extend'  '$LogFile'  '$MFT'   Boot   ProgramData  'Program Files'   Users   Windows
❯  tree -L 2 -a mnt
mnt
├── filesystems
│   ├── fs_0
│   └── fs_1
└── fs
    ├── $fs$
    ├── c:
    └── sysvol
```


Schamper:
Would using `self.absolute_path` instead of `self.path` work for this? A change for that was merged in this morning:

https://github.com/fox-it/dissect.target/blob/main/dissect/target/loader.py#L66-L74

william-billaud:
From my test (replacing line with `path = self.absolute_path if self.absolute_path.is_dir() else next(open_vhdx(self.absolute_path)).path()`).

But absolute_path can be null if resolve is False, so wouldn't it be a problem to rely on it inside a Loader?

william-billaud:
Any reason to not create a new variable and not just setting the path to the absolute path by default (and making this pr useless)?
This problem with relative path references and fuse is maybe present in other loaders and it would prevent this kind of bug.

I will make the change in the pr and let you see if you prefer to close it and just make the change mentioned above.

Schamper:
It's not very well documented, but we make extensive use of symbolic links to pull off all kinds of tricks with the loaders. So every use of a not-resolved and a resolved path is usually very deliberate.

In this case however, I think it makes sense to use the absolute path for these loaders.

william-billaud:
OK, thanks for your reply. I've changed to use absolute_path.

Schamper (changes requested):
I can't comment on it, but can you also make the `DirLoader` use `self.absolute_path`?

william-billaud:
@Schamper done

william-billaud:
@DissectBot agree [company="OrangeCyberdefense"]

Schamper (approved):

PR #1117, opened 2025-04-22
Add workaround for ANSI codes target-shell tests

### LINKED ISSUE #1111: Shell prompt tab autocomplete test does not work in uv environments
It seems like the test `tests/tools/test_shell.py::test_shell_prompt_tab_autocomplete` does not work in uv environments.

https://github.com/fox-it/dissect.target/blob/25aad3824be87e302974ae04d2b235ee6dc86b61/tests/tools/test_shell.py#L409

JSCU-CNI:
Should fix #1111. Only tested locally on uv cpython 3.12 and 3.13 so far, curious to see if the regular CI test environments still pass.

### DIFF
--- a/tests/tools/test_shell.py
+++ b/tests/tools/test_shell.py
@@ -48,14 +48,6 @@ INPUT_STREAM = f"""
     last line
 """
 
-if HAS_PEXPECT and platform.system() != "Windows":
-    ANSI_ESCAPE = re.compile(rb"\x07|\x08|\x0d|\x7f|\x1b[@-_][0-?]*[ -/]*[@-~]")
-    original_new_data = pexpect.expect.Expecter.new_data
-
-    def ansi_new_data(cls: pexpect.expect.Expecter, data: bytes) -> int | None:
-        """Matches on carriage returns (``\x0d`` / ``\r``) so change your ``child.expect()`` calls accordingly."""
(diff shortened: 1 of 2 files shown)


tests/tools/test_shell.py
```diff
@@ -47,6 +49,41 @@
     last line
 """
 
+if HAS_PEXPECT:
+
+    class AnsiExpecter(pexpect.expect.Expecter):
```
Schamper, line 54: Maybe stupid question, but can we use `unittest.mock.patch` (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) to temporarily patch `pexpect` with this behavior?

tests/tools/test_shell.py
```diff
@@ -411,7 +448,7 @@ def test_shell_prompt_tab_autocomplete() -> None:
     target_path = absolute_path("_data/tools/info/image.tar")
```
Schamper, line 448: Can't comment on it, but can you remove the skip for PyPy with this fix?

Schamper (changes requested):

Schamper (dismissed):

Schamper:
I can't reproduce either, let me at least fix the pypy3.9 tests because of the faulty version check, and then we can disable this test for PyPy3.10.

Schamper (approved):

PR #1129, opened 2025-04-30
Add some benchmarks

### LINKED ISSUE #1126: Add benchmark tests for scrape plugin
As discussed in https://github.com/fox-it/dissect.target/pull/1103/files#r2063022448

JSCU-CNI:
Fixes #1126. Inspired (read: copy/paste'd) setup from https://github.com/fox-it/dissect.util/pull/64.


Schamper:
Can we make the journal (and maybe the walkfs) benchmark more targeted? For example, only parse a single entry.

Schamper:
> > Can we make the journal (and maybe the walkfs) benchmark more targeted? For example, only parse a single entry.
> 
> What do you propose? Call `next()` once on a plugin function generator?

It's tricky for sure, since just benchmarking `next()` is not very fruitful (it will change the state). At least for the journal parser, `_parse_entry_object` seems to be deterministic.

It's a bit annoying that CodSpeed does not yet support the `benchmark.pedantic` mode from `pytest-benchmark` (see https://github.com/CodSpeedHQ/pytest-codspeed/issues/78), otherwise we could put plugin setup in there and test a single iteration of it easily.

Perhaps for the time being initializing the plugin and calling a single next is just the best we can do until CodSpeed supports that. So basically a `lambda: next(MyPlugin(target).func())` as benchmark.

Schamper (approved):

PR #1136, opened 2025-05-06
Support systemd drop files

twiggler:
Fixes:
-  `_select_parser` could be invoked with a `Path`, while it expects a `FileSystemEntry`. Additionally, there were multiple code paths in `parse` and `parse_config` to support `TargetPath`, `Path`, and `File` objects. Removed the dependence on `FileSystemEntry` by converting entries to `TargetPath` in the config file system.
- Systemd network parser did not collapse singleton values
- When multiple sections with the same name are encountered, the resulting key is a list of those sections:
Example:
```
key value1
   key2 value2a
key value1
   key2 value2b
   key3 value3b
```
  
 now parses as  `{"key value1": [{"key2": "value2a"}, {"key2": "value2b", "key3": "value3b"}]}`. Previously, this parsed as `{"key value1": {"key2": ["value2a", "value2b"], "key3": "value3b"}}`, where the parallel lists could become misaligned.

Supports:
- Drop files when invoking parse with a systemd file
- Empty value now clears corresponding key when merging systemd files.

### DIFF
--- a/dissect/target/filesystems/config.py
+++ b/dissect/target/filesystems/config.py
@@ -9,7 +9,6 @@ from dissect.target.exceptions import ConfigurationParsingError, FileNotFoundErr
 from dissect.target.filesystem import FilesystemEntry, VirtualFilesystem
 from dissect.target.helpers import fsutil
 from dissect.target.helpers.configutil import ConfigurationParser, parse
-from dissect.target.helpers.fsutil import TargetPath
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -123,7 +122,7 @@ class ConfigurationFilesystem(VirtualFilesystem):
         entry = file_entry
         config_parser = None
         try:
-            target_path = TargetPath(entry.fs, entry.path)
+            target_path = entry.fs.path(entry.path)
             config_parser = parse(target_path, *args, **kwargs)
         except ConfigurationParsingError as e:
             # If a parsing error gets created, it should return the `entry`
--- a/dissect/target/filesystems/config.py
+++ b/dissect/target/filesystems/config.py
@@ -9,6 +9,7 @@ from dissect.target.exceptions import ConfigurationParsingError, FileNotFoundErr
 from dissect.target.filesystem import FilesystemEntry, VirtualFilesystem
 from dissect.target.helpers import fsutil
 from dissect.target.helpers.configutil import ConfigurationParser, parse
+from dissect.target.helpers.fsutil import TargetPath
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -122,7 +123,8 @@ class ConfigurationFilesystem(VirtualFilesystem):
         entry = file_entry
         config_parser = None
         try:
-            config_parser = parse(entry, *args, **kwargs)
+            target_path = TargetPath(entry.fs, entry.path)
+            config_parser = parse(target_path, *args, **kwargs)
         except ConfigurationParsingError as e:
             # If a parsing error gets created, it should return the `entry`
             log.debug("Error when parsing %s with message '%s'", entry.path, e)
--- a/dissect/target/helpers/configutil.py
+++ b/dissect/target/helpers/configutil.py
@@ -9,7 +9,6 @@ from collections import deque
 from collections.abc import ItemsView, Iterable, Iterator, KeysView
 from configparser import ConfigParser, MissingSectionHeaderError
 from dataclasses import dataclass
-from fnmatch import fnmatch
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -22,14 +21,12 @@ from typing import (
 from defusedxml import ElementTree
 
 from dissect.target.exceptions import ConfigurationParsingError, FileNotFoundError
-from dissect.target.helpers.fsutil import TargetPath
+from dissect.target.helpers.utils import to_list
 
 if TYPE_CHECKING:
     from types import TracebackType
 
     from typing_extensions import Self
-
-    from dissect.target.filesystem import FilesystemEntry
 
 try:
     from ruamel.yaml import YAML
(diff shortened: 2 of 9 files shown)


dissect/target/filesystems/config.py
```diff
@@ -122,7 +123,8 @@ def _convert_entry(self, file_entry: FilesystemEntry, *args, **kwargs) -> Config
         entry = file_entry
         config_parser = None
         try:
-            config_parser = parse(entry, *args, **kwargs)
+            target_path = TargetPath(entry.fs, entry.path)
```
twiggler, line 126: reviewer: Perhaps there is a nicer way?

dissect/target/helpers/configutil.py
```diff
@@ -180,6 +180,43 @@ def read_file(self, fh: TextIO | io.BytesIO) -> None:
         if not isinstance(self.parsed_data, dict):
             self.parsed_data = self._collapse_dict(self.parsed_data, False)
 
+    def merge(self, other: ConfigurationParser) -> ConfigurationParser:
```
twiggler, line 183: Reviewer: could push this down to SystemDParser

dissect/target/filesystems/config.py
Schamper: Perhaps `entry.fs.path(entry.path)`?

I still think mixing the filesystem layer with the config layer is very confusing to begin with, but this is probably at least closer to the intended API.

dissect/target/helpers/configutil.py
Schamper: I think it's fine, could maybe be useful for other parsers in general.

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
 ) -> ConfigurationParser:
     parser_type = _select_parser(entry, hint)
 
     parser = parser_type.create_parser(options)
-    with entry.open("rb") if isinstance(entry, Path) else entry.open() as fh:
+    with entry.open("rb") as fh:
         open_file = io.TextIOWrapper(fh, encoding="utf-8") if not isinstance(parser, Bin) else io.BytesIO(fh.read())
         parser.read_file(open_file)
 
+    if isinstance(parser, SystemD) and isinstance(entry, Path):
```
Schamper, line 972: `entry` is now always a `Path` right?

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+        return _parse_drop_files(entry, options, parser)
+
     return parser
 
 
-def _select_parser(entry: FilesystemEntry, hint: str | None = None) -> ParserConfig:
+def _parse_drop_files(
+    pathOrEntry: Path, options: ParserOptions, main_parser: ConfigurationParser
```
Schamper, line 979:
```suggestion
    path: Path, options: ParserOptions, main_parser: ConfigurationParser
```
And change accordingly.

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+) -> ConfigurationParser:
+    if not isinstance(pathOrEntry, Path):
+        return main_parser
```
Schamper, line 982: Is this check still necessary?

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+
+    drop_folder = pathOrEntry.with_name(pathOrEntry.name + ".d")
+    if not drop_folder.exists():
```
Schamper, line 985:
```suggestion
    if not (drop_folder := path.with_suffix(".d")).exists():
```

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+        return main_parser
+    drop_files = sorted(drop_folder.glob("*.conf"))
+
+    for drop_file in drop_files:
```
Schamper, line 989:
```suggestion

    for drop_file in sorted(drop_folder.glob("*.conf")):
```

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+        if not drop_file.is_file():
+            continue
+
+        drop_file_parser = ParserConfig(SystemD).create_parser(options)
+        with drop_file.open("rb") as fh:
+            open_drop_file = io.TextIOWrapper(fh, encoding="utf-8")
```
Schamper, line 995: Why not open the `drop_file` immediately in text mode?

dissect/target/helpers/configutil.py
```diff
@@ -903,42 +949,63 @@ def parse(path: FilesystemEntry | TargetPath | Path, hint: str | None = None, *a
...
+    return main_parser
+
+
+def _select_parser(path: Path, hint: str | None = None) -> ParserConfig:
     if hint and (parser_type := CONFIG_MAP.get(hint)):
         return parser_type
 
     for match, value in MATCH_MAP.items():
-        if fnmatch(entry.path, f"{match}"):
+        if path.match(f"{match}"):
```
Schamper, line 1007:
```suggestion
        if path.match(match):
```
Useless f-string?

Schamper (changes requested):

dissect/target/filesystems/config.py
twiggler: `entry.fs.path(entry.path)` is indeed a bit nicer.

I still think mixing the filesystem layer with the config layer is very confusing to begin with, but this is probably at least closer to the intended API.

Yeah at least with this change the parsers are decoupled from the config layer.

I did not want to make more substantial changes without fully understanding why it was written this way, and the primary author
was on holiday though. I will discuss with him and possibly create a ticket depending on the outcome.

dissect/target/helpers/configutil.py
twiggler: No, thanks

Schamper (approved):

PR #1096, opened 2025-04-09
Add Podman and OCI container support

JSCU-CNI:
This PR:

1. Adds the `PodmanPlugin` for `images` and `containers`.
2. Adds a Podman `OverlayFilesystem` implementation
3. Extends the `ContainerImageTarSubLoader` with OCI image support
4. Adds child support for Podman containers in `PodmanChildTargetPlugin`
5. Adds a `ContainerPlugin` namespace plugin for `DockerPlugin` and `PodmanPlugin`
6. Improves `DockerPlugin` for snap installs and unifies `ports` output for container records

All modifications are tested using new or modified tests.

PR #1138, opened 2025-05-08
Fix default Docker daemon data directories

Zawadidone:
`get_data_path` can return `None` in case the directory is not configured.

https://docs.docker.com/engine/daemon/#daemon-data-directory

### DIFF
--- a/dissect/target/plugins/apps/container/docker.py
+++ b/dissect/target/plugins/apps/container/docker.py
@@ -344,25 +344,37 @@ def find_installs(target: Target) -> Iterator[Path]:
         "sysvol/ProgramData/docker/config/daemon.json",
     ]
 
+    default_data_paths = [
+        # Linux
+        "/var/lib/docker",
+        # Windows
+        "sysvol/ProgramData/docker",
+    ]
+
     user_config_paths = [
         # Docker Desktop (macOS/Windows/Linux)
         ".docker/daemon.json",
     ]
 
-    if (default_root := target.fs.path("/var/lib/docker")).exists():
-        yield default_root
+    for path in default_data_paths:
+        if (path := target.fs.path(path)).exists():
+            yield path
 
     for path in default_config_paths:
-        if (config_file := target.fs.path(path)).exists() and (
-            data_root_path := target.fs.path(get_data_path(config_file))
-        ).exists():
+        if (
+            (config_file := target.fs.path(path)).exists()
+            and (data_root_path := get_data_path(config_file))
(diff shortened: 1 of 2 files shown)


Zawadidone:
I stumbled on this while working on https://github.com/fox-it/dissect.target/pull/1139.

The PR does not solve the issue because the default Docker daemon data directory  (`/var/lib/docker` - https://github.com/fox-it/dissect.target/pull/1138/files#diff-592a62677741e44ccba1c0c8d78ad11118bb5cb1bb09f209d6b227ddd4975603R339-R344) that exists is not returned by the function `find_installs`: https://github.com/fox-it/dissect.target/pull/1096/files#diff-592a62677741e44ccba1c0c8d78ad11118bb5cb1bb09f209d6b227ddd4975603R330-R354. It is not required to configure the data directory in the daemon configuration therefore the two default values for Linux and Windows need to be checked.

Schamper:
> I stumbled on this while working on #1139.
> 
> The PR does not solve the issue because the default Docker daemon data directory (`/var/lib/docker` - https://github.com/fox-it/dissect.target/pull/1138/files#diff-592a62677741e44ccba1c0c8d78ad11118bb5cb1bb09f209d6b227ddd4975603R339-R344) that exists is not returned by the function `find_installs`: https://github.com/fox-it/dissect.target/pull/1096/files#diff-592a62677741e44ccba1c0c8d78ad11118bb5cb1bb09f209d6b227ddd4975603R330-R354. It is not required to configure the data directory in the daemon configuration therefore the two default values for Linux and Windows need to be checked.

Doesn't line 330-337 in https://github.com/fox-it/dissect.target/pull/1096 from JSCU do exactly that? Except that this PR has a value for Windows too (and comments, nice).

I personally fail to see how https://github.com/fox-it/dissect.target/pull/1096 fails to address this, can you be more explicit in why it doesn't? It looks like almost the exact same code as this PR, except with some error logging.

I would like to see this PR be a combination of this and that portion of #1096. So keep the Windows directory and comments near the paths, have the body of `find_installs` from #1096, and the unit test from this PR.

Schamper (changes requested):

Zawadidone:
Okay a misunderstanding from my side, I used to wrong version of the code because my uv environment was bricked.

Schamper:
@JSCU-CNI unless I'm overlooking something I think this is pretty much the same as that particular piece of code in your PR. I propose to merge this one first (since it's a lot smaller) and then you can simply rebase your PR. I'll get to that soon, you know how it is 😉.

Schamper (approved):

PR #1141, opened 2025-05-13
Add checks for a windows UTC timezone inside windows datetime

Miauwkeru:
It can occur that the systemtime structure only contains 0's. So we set
a default of 1 for month and day to make it function correctly

### DIFF
--- a/dissect/target/plugins/os/windows/datetime.py
+++ b/dissect/target/plugins/os/windows/datetime.py
@@ -158,6 +158,7 @@ class WindowsTimezone(tzinfo):
         assert dt.tzinfo is self
 
         tzi = self.dynamic_dst.get(dt.year, self.tzi)
+        # If the time zone does not support daylight saving time the wMonth member is zero
         if tzi.daylight_date.wMonth == 0 or tzi.standard_date.wMonth == 0:
             return False
 


Schamper (changes requested):
Would you mind specifically adding a unit test for `parse_systemtime_transition`?

dissect/target/plugins/os/windows/datetime.py
```diff
@@ -76,6 +76,9 @@ def parse_systemtime_transition(systemtime: c_tz._SYSTEMTIME, year: int) -> date
     Reference:
         - https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-time_zone_information
     """
+    if systemtime.wDay > 5:
```
Schamper, line 79: It can't be lower than `1` either, why only specifically check for this one? Either check for everything or not at all I think?

dissect/target/plugins/os/windows/datetime.py
```diff
@@ -155,6 +158,9 @@ def is_dst(self, dt: datetime) -> bool:
         assert dt.tzinfo is self
 
         tzi = self.dynamic_dst.get(dt.year, self.tzi)
```
Schamper, line 160:
```suggestion
        tzi = self.dynamic_dst.get(dt.year, self.tzi)
        # If the time zone does not support daylight saving time the wMonth member is zero
```

Schamper (changes requested):

Schamper (approved):

PR #1147, opened 2025-05-18
Allow glob in path argument of target-fs

### LINKED ISSUE #1002: Support wildcards in target-fs
It would be great if we could use wildcards in target-fs. This does work in target-shell, here we can do for example `save var/log/syslog*`

If we try this in with target-fs (`target-fs <file> cp var/log/syslog*` it results in file not found

Matthijsy:
This PR adds support for globs in the `path` argument of `target-fs`, this way we can for example extract all EVTX files from a collect by running `target-fs <target> cp "**/*.evtx"` or all syslog files by running `target-fs <target> cp /var/log/syslog*`

Fixes #1002

### DIFF
--- a/dissect/target/tools/fs.py
+++ b/dissect/target/tools/fs.py
@@ -156,13 +156,16 @@ def main() -> int:
         log.debug("", exc_info=e)
         return 1
 
-    path = target.fs.path(args.path)
+    glob_path = str(args.path).lstrip("/")
 
-    if not path.exists():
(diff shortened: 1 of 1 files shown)


Schamper (changes requested):
Would you mind adding a few unit tests?

Schamper (changes requested):
Can you fix the linter?

Schamper (approved):

PR #1153, opened 2025-05-26
Lower target mount cache

twiggler:
Although there were multiple factors contributing to the high memory usage of target.mount, the cache size was a bit too generous. 

Significantly lowered the cache size

### DIFF
--- a/dissect/target/plugins/os/windows/tasks/job.py
+++ b/dissect/target/plugins/os/windows/tasks/job.py
@@ -204,8 +204,7 @@ class AtTask:
 
         # check which prio bit is set
         for key in self.at_data.task_prio.fields:
-            value = getattr(self.at_data.task_prio, key)
-            if value:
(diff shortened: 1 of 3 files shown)


dissect/target/plugins/os/windows/tasks/job.py
```diff
@@ -203,7 +203,8 @@ def __init__(self, job_file: TargetPath, tzinfo: datetime.tzinfo = datetime.time
         self.data = self.at_data.user_data
 
         # check which prio bit is set
-        for key, value in self.at_data.task_prio._values.items():
+        for key in self.at_data.task_prio.fields:
+            value = getattr(self.at_data.task_prio, key)
             if value:
```
Schamper, line 208:
```suggestion
            if getattr(self.at_data.task_prio, key):
```

Schamper (changes requested):

PR #1167, opened 2025-06-10
Add __os__ to OSPlugin

Miauwkeru:
This creates a shorthand for when we want to theck the MRO of t._os_plugin.
This was required for acquire, otherwise we'd need to initialize every one
of those os_plugins with the target and that would be very expensive.

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1259,6 +1259,8 @@ class OSPlugin(Plugin):
         Returns:
             A slug of the OS name, e.g. 'windows' or 'linux'.
         """
+        if self.__os__:
+            return self.__os__.value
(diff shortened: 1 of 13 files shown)


Schamper (changes requested):
Why is this change necessary? You can walk the MRO just fine without it:

```python
In [21]: t._os_plugin
Out[21]: dissect.target.plugins.os.unix.linux.fortios._os.FortiOSPlugin

In [22]: instance, _ = t._functions["os"]

In [23]: for klass in t._os_plugin.mro():
    ...:     if klass is OSPlugin:
    ...:         break
    ...:     print(klass.os.__get__(instance))
    ...:
fortios
linux
unix
```

PR #1174, opened 2025-06-16
Small shell improvements

### LINKED ISSUE #1003: target-shell prompt not rendering ANSI color codes

JSCU-CNI:
This PR fixes #1003 for `target-diff shell`, adds glob support to `hexdump` and adds the `mounts` command.


dissect/target/tools/info.py
```diff
@@ -155,6 +156,9 @@ def get_volumes_info(target: Target) -> list[dict[str, str | int]]:
     return [{"name": v.name, "size": v.size, "fs": v.fs.__class__.__name__} for v in target.volumes]
 
 
+def get_mounts_info(target: Target) -> list:
+    return [{"fs": fs.__class__.__name__, "path": path} for path, fs in target.fs.mounts.items()]
```
Schamper, line 160: Is the class name preferred over `__type__`?

dissect/target/tools/diff.py
```diff
@@ -402,23 +403,19 @@ def prompt(self) -> str:
...
-        if os.getenv("NO_COLOR"):
-            suffix = f"{prompt_base}:{self.cwd}$ "
-        else:
-            suffix = f"\x1b[1;32m{prompt_base}\x1b[0m:\x1b[1;34m{self.cwd}\x1b[0m$ "
+        base = f"{src_name}/{dst_name}" if src_name != dst_name else src_name
+        suffix = "{base}:{self.cwd}$" if os.getenv("NO_COLOR") else "{BOLD_GREEN}{base}{RESET}:{BOLD_BLUE}{cwd}{RESET}$"
 
         if len(self.targets) <= 2:
-            return f"(diff) {suffix}"
+            return f"(diff) {suffix} ".format(base=base, cwd=self.cwd, **ANSI_COLORS)  # type: ignore
```
Schamper, line 410: Can't you format the suffix earlier so you avoid doing it twice in the return?

Schamper (changes requested):

Schamper (changes requested):
Can you add some unit tests?

Schamper (approved):

Schamper:
The tests seem stuck, maybe a change causes a lock somewhere?

Schamper (changes requested):

PR #711, opened 2024-05-21
Improve DPAPI plugin

JSCU-CNI:
This PR improves several DPAPI related features:

* move LSA logic from DPAPI to separate plugin
* move SAM and CREDHIST plugins to `dissect.target.plugins.os.windows.credential`
* add Windows XP support to DPAPI plugin (RC4 and DES3)
* fix bug in `_SHA256` identifier (`0x800C` instead of `0x8004`)
* add providers for DPAPI masterkey passphrases

The latter "dpapi provider" feature is experimental and we are keen to discuss a better `InternalNamespacePlugin` implementation.

PR #1175, opened 2025-06-16
Add programmatic support for nested namespaces

### LINKED ISSUE #758: Nested namespaces do not work
Namespaces such as `foo.bar.baz` do not seem to work with `target-query -f foo.bar.baz`. They do seem to be discovered by `find_plugin_functions` correctly but are not mapped correctly in the python `target` object.

Implementation notes:

t.foo will return an instance of FooPlugin

Implement a getattr in the Plugin class

getattr checks if self is a namespace

if not -> just super() it (which raises an error)

if it is, do a lookup for `{self.namespace}.{requested attr`}

This should return another instance of the plugin class (BarPlugin with namespace `foo.bar`)

Because you return another instance of a class, you get the "recursion" for free

- This method of doing lookups instead of  "building" something during registration time has the benefit of working on "cold starts" of the plugin system too
- Note: also check the namespace handling in Target.get_function. Maybe consider implementing the logic completely in here and the getattr in Plugin just forwards it to this function

Miauwkeru:
closes #758

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -438,7 +438,7 @@ class Plugin:
         if not self.__namespace__:
             raise AttributeError(name)
 
-        # Attempt to lookup the atribute using the namespace
+        # Attempt to lookup the attribute using the namespace
         try:
             _, func = self.target.get_function(f"{self.__class__.__namespace__}.{name}")
         except PluginNotFoundError:


dissect/target/plugin.py
```diff
@@ -877,7 +902,7 @@ def find_functions(
         else:
             # If we don't have an exact function match, do a slower treematch
             if matches := list(_filter_tree_match(pattern, os_filter, show_hidden=show_hidden)):
-                found.extend(matches)
+                found.extend(match for match in matches if match not in found)
             else:
```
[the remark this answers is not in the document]
Schamper: Dictionary keys maintain insertion order and can have a faster membership check.

dissect/target/plugin.py
```diff
@@ -1285,12 +1310,23 @@ def list_children(self) -> Iterator[ChildTargetRecord]:
         raise NotImplementedError
 
 
+def full_namespace(namespace_plugin: type[NamespacePlugin]) -> None:
+    namespace = namespace_plugin.__namespace__
+    for klass in namespace_plugin.mro()[1:]:
```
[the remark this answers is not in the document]
Schamper:
```python
class Foo(Plugin):
    __namespace__ = "foo"

class FooBar(Plugin):
    __namespace__ = "foo.bar"
```

twiggler (changes requested):

dissect/target/plugin.py
```diff
@@ -432,21 +432,25 @@ def __init_subclass__(cls, **kwargs):
...
     def __getattr__(self, name: str, /) -> Any:
         if not self.__namespace__:
             raise AttributeError(name)
 
-        funcs = self.__functions__
-        if isinstance(self, NamespacePlugin):
-            funcs = self.__subplugins__
+        # If the namespace doesn't expose the function ``name``, we might have encountered a implicit nested namespace.
+        # These are the plugins that build upon a namespace, but not through inheritance
+        # So we attempt to resolve it by looking for the namespace with the name
```
twiggler, line 445:
```suggestion
        # So we attempt to resolve it by looking up the namespace by name.
```

dissect/target/plugin.py
```diff
@@ -432,21 +432,25 @@ def __init_subclass__(cls, **kwargs):
...
+
     def __getattr__(self, name: str, /) -> Any:
         if not self.__namespace__:
             raise AttributeError(name)
 
-        funcs = self.__functions__
-        if isinstance(self, NamespacePlugin):
-            funcs = self.__subplugins__
+        # If the namespace doesn't expose the function ``name``, we might have encountered a implicit nested namespace.
+        # These are the plugins that build upon a namespace, but not through inheritance
```
twiggler, line 444:
```suggestion
        # These are the plugins that build upon a namespace, but not through inheritance.
```

twiggler (changes requested):

twiggler (dismissed):

tests/helpers/test_docs.py
```diff
@@ -42,7 +42,7 @@ def test_docs_get_func_description() -> None:
     func = IISLogsPlugin.logs
     func_desc = docs.get_func_description(func, with_docstrings=False)
 
-    assert func_desc == "iis.logs - Return contents of IIS (v7 and above) log files. (output: records)"
+    assert func_desc == "(webserver.)iis.logs - Return contents of IIS (v7 and above) log files. (output: records)"
```
Schamper, line 45: I still don't understand what this `(webserver.)` part is and I don't see it documented anywhere? What is it and why is it necessary? It's super confusing because I have no idea how to invoke `target-query -f <what goes here?>` now. Do I do `target-query -f '(webserver.)iis.logs'`?

Schamper (changes requested):

tests/helpers/test_docs.py
Schamper: #758 originates from this: https://github.com/fox-it/dissect.target/pull/711#issuecomment-2328561977

It's specifically for `__namespace__ = "foo.bar"` cases, the `NamespacePlugin` does not play a role in it nor does subclassing. I thought that was clear with https://github.com/fox-it/dissect.target/pull/1175#discussion_r2152602390

The only purpose of `NamespacePlugin` is grouping of "same type" plugins. It plays no role in the "plugin/namespace tree". There's no need for the `webserver.iis` syntax and it only complicates the plugin calling conventions with no real benefit.

tests/test_plugin.py
```diff
@@ -377,7 +377,7 @@ def f6(self) -> str:
         ("test.[!y]*", 1),  # Found with tree search, all in test not starting with y (so x13 is ok)
         ("test.???.??", 1),  # Found with tree search, using question marks
         ("x13", 0),  # Not Found: Part of namespace but no match
-        ("Warp.*", 0),  # Not Found: Namespace != Module so 0
+        ("Warp.*", 1),  # Found: Using the namespace
```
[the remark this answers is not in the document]
Schamper: > you can use both `-f foo.bar` and `-f foo.bar.buz` and `-f foo.bar.baz` without any issues. This is without the `Plugin.__getattr__`.

But adding the `__getattr__` makes it work within the Python API too, which is intended behavior and what #758 is for.

> It felt like an oversight to not have, so I added it for consistency on how the plugin got called. Because if I was a new user, and was able to run a plugin with `-f foo.bar.baz,foo.bar.buz` I'd also expect `-f foo.bar.*` to function.

You are introducing an inconsistency, since the wildcard match is explicitly only for module paths. Now it's somehow also for namespaces?

Regardless, this is scope creep and beyond the requested task of #758. If you think this is a good idea, you should create a ticket for it and it should be refined with the stakeholders. It does not belong in this PR.

dissect/target/plugin.py
```diff
@@ -1285,12 +1311,29 @@ def list_children(self) -> Iterator[ChildTargetRecord]:
...
+
+    When ``plugin_cls`` has multiple plugins as its base, the namespace will reflect that
+    """
+    if plugin_cls.__namespace__ is None:
+        return None
+
+    mro = plugin_cls.mro()
+    mro.reverse()
+    # Join all the namespaces together
+    return ".".join(klass.__namespace__ for klass in mro if issubclass(klass, Plugin) and klass.__namespace__)
```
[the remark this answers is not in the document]
twiggler: > For what it's worth, the plugin system has no goal to ever rely on class inheritance for building namespace/plugin trees.

Yeah fully support this.

>  only the specialized NamespacePlugin (which currently has an extremely confusing name) uses it, and only to group plugins within the same "purpose", e.g. browsers.

I understand now that the `NamespacePlugin` is not meant to be used recursively. So given that the implied tree has only two levels, perhaps I should rephrase to: 

why do we use the class hierarchy to group and aggregate plugins (functions).

tests/test_plugin.py
```diff
@@ -597,6 +597,44 @@ def baz(self) -> Iterator[str]:
...
+    assert isinstance(target_bare.foo, Foo)
+    assert isinstance(target_bare.foo.bar, FooBar)
+    assert isinstance(target_bare.foo.baz, FooBaz)
+    assert hasattr(target_bare.foo.bar, "bazz")
+
+    with pytest.raises(AttributeError):
+        target_bare.foo.bazz()
+
+    with pytest.raises(AttributeError):
+        target_bare.foo.bar.foo()
```
Schamper, line 635: Maybe add a test calling some functions that should work.

Schamper (changes requested):

tests/test_plugin.py
Schamper: Can you also update the PR title to be more in line with the change.

dissect/target/plugin.py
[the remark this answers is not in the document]
twiggler: The discussion will continue in https://github.com/fox-it/dissect.target/issues/1232.

I will write some ideas there.

Schamper (approved):

Schamper (approved):

PR #1187, opened 2025-06-19
Stop consuming arguments for each plugin

Miauwkeru:
the rest values get consumed when a plugin gets executed.
So if there are multiple plugins that use the same flag,
the flag only gets passed to the first function


Schamper:
Can you also change the PR title to something less factual and more what it achieves? I.e. "Stop consuming arguments for each plugin" or "Allow conflicting argument names in plugins" or something along those lines.

A unit test to ensure this behavior stays in place would also be nice.

twiggler (approved):

PR #1189, opened 2025-06-19
Fix inconsistent duplicate field mappings in various plugins (part 2)

JSCU-CNI:
This PR is part of an ongoing effort to fix inconsistencies in record field types across dissect.

Also includes:

- Add missing tests for the `ActivitiesCachePlugin` Windows plugin.
- Add translation for download states in `ChromiumMixin`
- Add support for multiple DNS entries in `WireguardPlugin`
- Fix inconsistencies in Linux Sockets plugin
- Add translation for priority integers in `JournalPlugin`
- Add translation for Amcache PcaGeneralDb types
- Add improved Office trusteddocs artefact parsing
- Add tests for `ActivitiesCachePlugin`
- Move amcache tests to correct folder

PR #990, opened 2025-01-14
Fix inconsistent duplicate field mappings in various plugins

JSCU-CNI:
This PR is part of an ongoing effort to fix inconsistencies in record field types across dissect.

PR #1192, opened 2025-06-23
Add record field type tests

JSCU-CNI:
This PR adds tests to prevent duplicate field names with incompatible field types as identified and fixed in #1189, #990 and #982.


Schamper (changes requested):
Can you add an example in the docstring for what this actually checks/calls out? It's a bit hard to determine from just the test.

Schamper (changes requested):
Maybe just generalize the `[]` case, since every `flow.record` type can be made a list. Tests are failing because of a missing `uint16[]`

Schamper (approved):

PR #1194, opened 2025-06-23
Check for optional dependencies

JSCU-CNI:
This PR removes debug stacktraces from `target-* -vv` stderr output  from Carbon Black, SMB and MQTT loaders and filesystems. This change will cause several mocks to fail in the tests, unsure how we should fix those.

### DIFF
--- a/dissect/target/filesystems/cb.py
+++ b/dissect/target/filesystems/cb.py
@@ -18,7 +18,6 @@ try:
     from cbc_sdk.live_response_api import LiveResponseError, LiveResponseSession
 
     HAS_CARBON_BLACK = True
-
 except ImportError:
     HAS_CARBON_BLACK = False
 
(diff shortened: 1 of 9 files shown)


dissect/target/filesystems/cb.py
```diff
@@ -28,6 +35,9 @@ class CbFilesystem(Filesystem):
     __type__ = "cb"
 
     def __init__(self, session: LiveResponseSession, prefix: str, *args, **kwargs):
+        if not HAS_CARBON_BLACK:
+            raise ImportError("Please install 'carbon-black-cloud-sdk-python' to use CbFilesystem.")
```
Schamper, line 39: Can you make all the error messages consistent? Like the SMB one.

Schamper (changes requested):
Small style comment, will have a thought about the mocks after tests have run.

Schamper:
I'd like to merge #1265 and #1263 in first before continuing with this one. Gives me a little bit more of a stable basis to continue working on.

PR #1204, opened 2025-06-25
Add support for Hyper-V backups to VBK loader

Zawadidone:
Add support for full backup files of full Windows VM images.


Schamper:
Still waiting for the test file.

Schamper:
Do you happen to have a full VBK file with a VHDX image I could poke around in? Can optionally email it to me if it's too large to upload to GitHub.

Schamper:
> I have several yes, but it's all customer data so I cannot share it. Let me know if you really want it, we might be able to work something out.

That sounds like too much hassle for test data. Was hoping you'd have a test setup with Hyper-V 😄. Maybe I'll look into creating it myself or if you have the time it'd be nice to have a proper full VBK test file.

Schamper:
> Oh, you want to use a full VBK file for the tests? It will be pretty huge. I'm guessing at least 15GB. Will that work with GitHub LFS?

Why that big? For the other VBK test files we just have an empty VM with an empty small disk and let Veeam create a backup of that. Don't need to put a full install on the disk.

PR #1207, opened 2025-06-26
Search additional paths for home directories

twiggler:
- Ensure `%windir%/SysWOW64/config/systemprofile` is acquired and discoverable by plugins.
- Move `misc_home_dirs` into target.
- Added a new function `all_home_dirs` to discover user directories without requiring user information

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1263,7 +1263,7 @@ class OSPlugin(Plugin):
         """Yields miscellaneous user paths and user keys.
 
         Example:
-            ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
+            ("c:/Windows/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
         """
 
         # Consider moving this to the UsersPlugin class, and create separate UsersPlugin subclasses for each OS
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1160,6 +1160,8 @@ class OSPlugin(Plugin):
     methods with the same ``@classmethod`` or ``@export(...)`` annotation.
     """
 
+    _misc_home_dirs: tuple[tuple[str, str | None]] = ()
+
     def __init_subclass__(cls, **kwargs):
         # Note that cls is the subclass
         super().__init_subclass__(**kwargs)
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1259,6 +1259,15 @@ class OSPlugin(Plugin):
         raise NotImplementedError
 
     @internal
+    def misc_home_dirs(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        """Return a generator home directories and a user key.
+
+        Example output: ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
+        """
+
+        raise NotImplementedError
+
+    @internal
     def os_tree(self) -> list[str]:
         """Returns the :func:`os` value of this and all the OS plugin parents."""
         result: list[str] = []
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1160,8 +1160,6 @@ class OSPlugin(Plugin):
     methods with the same ``@classmethod`` or ``@export(...)`` annotation.
     """
 
-    _misc_home_dirs: tuple[tuple[str, str | None]] = ()
-
     def __init_subclass__(cls, **kwargs):
         # Note that cls is the subclass
         super().__init_subclass__(**kwargs)
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1259,12 +1259,15 @@ class OSPlugin(Plugin):
         raise NotImplementedError
 
     @internal
-    def misc_home_dirs(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
-        """Return a generator home directories and a user key.
+    def misc_user_paths(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        """Yields miscellaneous user paths and user keys.
 
-        Example output: ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
+        Example:
+            ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
         """
 
+        # Consider moving this to the UserPlugin class, and create separate UserPlugin subclasses for each OS
+        # when the mechanism becomes more complex.
         raise NotImplementedError
 
     @internal
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -1266,8 +1266,8 @@ class OSPlugin(Plugin):
             ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
         """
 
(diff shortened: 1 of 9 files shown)


dissect/target/plugins/general/users.py
```diff
@@ -82,8 +82,49 @@ def all(self) -> Iterator[UserDetails]:
...
+        for misc_home_dir, user_criterion in self.target.misc_home_dirs():
+            # We skip the entry if no user can be found.
+            # We cannot use set membership check here because of https://github.com/fox-it/dissect.target/issues/1203
+            if any(seen_path.samefile(misc_home_dir) for seen_path in seen) or not user_criterion:
+                continue
+
+            if (user := self.find(**{user_criterion[0]: user_criterion[1]})) is None:
+                continue
+
+            yield UserDetails(user=UnixUserRecord(name=misc_home_dir), home_path=misc_home_dir)
```
Schamper, line 108: Maybe introduce a barebones `UserRecord`?

dissect/target/plugins/os/default/_os.py
```diff
@@ -50,6 +50,14 @@ def version(self) -> str | None:
     def users(self) -> Iterator[Record]:
         yield from ()
 
+    @internal
+    def misc_home_dirs(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
```
Schamper, line 54: I'm not entirely sure if I like this being on the OS plugin. You'd need to add it to the base `OSPlugin` class too, if we are going that way though.

Maybe rename it to `misc_user_paths`, it feels a bit more generic (home directory is more a Unixy term).

dissect/target/plugins/os/unix/_os.py
```diff
@@ -159,6 +159,14 @@ def users(self, sessions: bool = False) -> Iterator[UnixUserRecord]:
                     _target=self.target,
                 )
 
+    @internal
+    def misc_home_dirs(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        if (root_path := self.target.resolve("root")).exists():
+            yield (root_path, ("uid", 0))
+
+        if (home_path := self.target.resolve("home")).exists():
```
Schamper, line 167: Why the resolves here, and not just `t.fs.path`?

dissect/target/plugins/os/unix/bsd/darwin/_os.py
```diff
@@ -32,6 +34,13 @@ def detect(cls, target: Target) -> Filesystem | None:
                 return fs
         return None
 
+    @internal
+    def misc_home_dirs(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        yield from super().misc_home_dirs()
+
+        if (user_path := self.target.resolve("Users")).exists():
```
Schamper, line 41: Same question.

dissect/target/plugins/general/users.py
twiggler: Not sure what you mean, but setting user this way was a bug, we have a user record from `self.find`.

Amended unit test

Schamper: I meant to not abuse `UnixUserRecord` even though we may not be running on a Unix system, but it's irrelevant now.

dissect/target/plugins/os/default/_os.py
twiggler: I thought about this over the weekend and I think having separate user plugins per OS is a bit too much, for now.

`misc_user_paths` doesn't feel  out of place to me in OSplugin (at least for now). 

Will add a comment to consider refactoring if more machinery is needed

dissect/target/plugins/general/users.py
```diff
@@ -82,8 +82,49 @@ def all(self) -> Iterator[UserDetails]:
...
+                continue
+
+            if (user_details := self.find(**{user_criterion[0]: user_criterion[1]})) is None:
+                continue
+
+            yield UserDetails(user=user_details.user, home_path=misc_home_dir)
+            seen.add(misc_home_dir)
+
+    @cached_property
+    def all_home_dirs(self) -> Iterator[TargetPath]:
```
Schamper, line 112:
```suggestion
    def all_home_paths(self) -> Iterator[TargetPath]:
```
?

dissect/target/plugins/os/unix/_os.py
```diff
@@ -159,6 +159,14 @@ def users(self, sessions: bool = False) -> Iterator[UnixUserRecord]:
                     _target=self.target,
                 )
 
+    @internal
+    def misc_user_paths(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        if (root_path := self.target.fs.path("root")).exists():
+            yield (root_path, ("uid", 0))
+
+        if (home_path := self.target.fs.path("home")).exists():
```
Schamper, line 167:
```suggestion
        if (root_path := self.target.fs.path("/root")).exists():
            yield (root_path, ("uid", 0))

        if (home_path := self.target.fs.path("/home")).exists():
```
It doesn't make any functional difference but visually it helps discern quicker what the intent is.

dissect/target/plugins/os/unix/bsd/darwin/_os.py
```diff
@@ -32,6 +34,13 @@ def detect(cls, target: Target) -> Filesystem | None:
                 return fs
         return None
 
+    @internal
+    def misc_user_paths(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        yield from super().misc_user_paths()
+
+        if (user_path := self.target.fs.path("Users")).exists():
```
Schamper, line 41:
```suggestion
        if (user_path := self.target.fs.path("/Users")).exists():
```

dissect/target/plugins/os/windows/_os.py
```diff
@@ -314,6 +314,25 @@ def users(self) -> Iterator[WindowsUserRecord]:
...
+                ("%windir%/SysWOW64/config/systemprofile", ("sid", "S-1-5-18")),
+            ]
+            if (resolved_path := self.target.resolve(path)).exists()
+        )
+
+        # Homedirs for users when there is no user profile information available.
+        user_dirs = ("sysvol/Users", "sysvol/Documents and Settings")
+        for user_dir in user_dirs:
+            if (resolved := self.target.resolve(user_dir)).exists():
+                yield from ((entry, None) for entry in resolved.iterdir() if entry.is_dir())
```
Schamper, line 334: Why not make this a list comprehension too like the one above? Or at least put the `user_dirs` tuple inside the loop?

dissect/target/plugin.py
```diff
@@ -1258,6 +1258,18 @@ def os(self) -> str:
         """
         raise NotImplementedError
 
+    @internal
+    def misc_user_paths(self) -> Iterator[tuple[str, tuple[str, str] | None]]:
+        """Yields miscellaneous user paths and user keys.
+
+        Example:
+            ("%windir%/ServiceProfiles/LocalService", ("sid", "S-1-5-19"))
```
Schamper, line 1266: This example doesn't reflect reality, where we yield resolved paths instead of paths containing environment variables.

Schamper (changes requested):

dissect/target/plugins/os/windows/_os.py
twiggler: > Why not make this a list comprehension too like the one above? 

With multiple nestings / sources I find the python list comprehension to be less clear than an explicit loop

> Or at least put the user_dirs tuple inside the loop?

Sure

PR #1210, opened 2025-06-30
Add creation date to the `sam` plugin

qmadev:
This PR adds the creation date for local user accounts in the `sam` plugin. Regripper already had this functionality but it would be nice if Dissect has is as well.

https://github.com/keydet89/RegRipper3.0/blob/master/plugins/samparse.pl


Schamper (changes requested):
Can you add a unit test.

Schamper (approved):

PR #1211, opened 2025-06-30
Change empty volume logging

### LINKED ISSUE #1164: Change undetectable filesystem warning to info log
It's quite an annoying log message because it's quite common that there's at least one partition with no filesystem on it. Maybe read the first 1MB or something and if it's all 0 bytes change it to info, otherwise keep it as warning.

Schamper:
Closes #1164.


dissect/target/target.py
```diff
@@ -902,8 +902,13 @@ def apply(self) -> None:
                 if getattr(vol, "fs", None) is not None:
                     self.target.filesystems.add(vol.fs)
             except FilesystemError as e:  # noqa: PERF203
-                self.target.log.warning("Can't identify filesystem: %s", vol)
-                self.target.log.debug("", exc_info=e)
+                vol.seek(0)
```
twiggler, line 905: If the `seek` or `read` fail, is it ok if we crash here?

Or perhaps add a nested try-catch?

twiggler: If you are sure

 that previously successful seeks and reads guarantee that future seeks and reads succeed,

And there is no value in recovery by logging and skipping the volume,

You can merge this in

PR #1238, opened 2025-07-17
Fix registry data collection and date parsing

falklindner-dev:
* Fix registry data collection and date parsing
* Added registry keys for 32bit applications in WOW6432Node Filtered registry keys from hives in RegBack folder to omit duplication Added more robust date parsing, some software vendors do not adhere to standard
* Refactored code to inject date parsing into class
* Created new helper for standard date parsing
* Added another Uninstall location in HKEY Users 
* Updated the tests for applications.py


dissect/target/helpers/parsers.py
```diff
@@ -0,0 +1,37 @@
...
+
+import datetime
+from typing import Callable
+
+# Define a type hint for our parser function for better readability and type checking.
+# It's a function that takes a string and returns a datetime object or None.
+DateTimeParser = Callable[[str], datetime.datetime | None]
+
+
+def default_datetime_parser(date_string: str) -> datetime.datetime | None:
```
Schamper, line 11: I think for the time being, this function might be better placed in https://github.com/fox-it/dissect.target/blob/main/dissect/target/helpers/utils.py

dissect/target/plugins/os/windows/regf/applications.py
```diff
@@ -20,13 +20,41 @@
...
+
 
 class WindowsApplicationsPlugin(Plugin):
     """Windows Applications plugin."""
 
-    def __init__(self, target: Target):
+    def __init__(
+        self,
+        target: Target,
+        datetime_parser: DateTimeParser = default_datetime_parser,
```
Schamper, line 36: I think with the current coding style in the Dissect projects, it's better to just use `default_datetime_parser` as a regular helper function, rather than in a constructor pattern like this. Maybe rename the helper function to `common_datetime_parser` so it has a more descriptive name.

dissect/target/plugins/os/windows/regf/applications.py
```diff
@@ -20,13 +20,41 @@
...
+
+        key_list = list(self.target.registry.keys(KEYS))
+        """
+        Removing all keys from Backup registry hives located in sysvol/windows/system32/config/RegBack
+        Parsing keys from RegBack leads to duplicate applications. A VirtualHive might not have a filepath attribute.
+        Checking if the attribute exists first. If not, it is assumed it's not from RegBack.
+        """
+        self.keys = [
+            key for key in key_list if not hasattr(key.hive, "filepath") or "RegBack" not in key.hive.filepath.parts
+        ]
```
Schamper, line 57: This is of course not limited to the application plugin, so I think we should think of a more robust and uniform solution for this. Can you create a new ticket for this problem in general (duplicate entries due to `RegBack` hives)?

Schamper (changes requested):

dissect/target/plugins/os/windows/regf/applications.py
falklindner-dev: Done in [Issue](https://github.com/fox-it/dissect.target/issues/1321)

falklindner-dev:
The updated version now fixes #1321 and implements comments 1 and 2

dissect/target/helpers/parsers.py
falklindner: Done in latest commit

Issue #1179, opened 2025-06-18
Plugin `_filter_compatible` check is slow on UNIX targets


JSCU-CNI:
Currently commands such as `target-query -l "*" example-target.img` can take quite long on UNIX targets. Checks on Windows targets seem to perform significantly faster. Perhaps some performance profiling is in order.

Schamper:
> It seems some NamespacePlugins would be skipped entirely with this approach: NamespacePlugins without further inheritance, such as the `mft` plugin, which is of course not acceptable.

This is a bug, see https://github.com/fox-it/dissect.target/issues/1180#issuecomment-2985205923.

PR #1249, opened 2025-07-22
Improve _filter_compatible performance by filtering out NamespacePlugins

### LINKED ISSUE #1180: Exported functions from NamespacePlugin not findable
Putting this issue here after a quick internal discussion. It seems that currently for some plugins that are created from a `NamespacePlugin`, some of the functions don't have a `findable` value of `True`, while this should be the intended behavior when looking at the docs.

```python
>>> from dissect.target.plugin import find_functions
>>> hidden = [f.name for f in find_functions("\*", show_hidden=True)[0] if f not in find_functions("\*")[0] and f.findable is False]
>>> hidden
['example', 'example_none', 'example_record', 'example_user_registry_record', 'example_yield', '_dpapi_keyprovider.keys', 'locate.locate', 'packagemanager.logs', 'mft.body', 'mft.records', 'mft.timeline', 'browser.cookies', 'browser.downloads', 'browser.extensions', 'browser.history', 'browser.passwords', 'chat.history', 'editor.extensions', 'editor.history', 'editor.tabs', 'remoteaccess.filetransfer', 'remoteaccess.logs', 'ssh.authorized_keys', 'ssh.known_hosts', 'ssh.private_keys', 'ssh.public_keys', 'ssh.config', 'ssh.sessions', 'webserver.logs', 'webserver.access', 'webserver.error']
```

This isn't a big issue with the `target-query` tool, as this by default passes the `show_hidden=True` to find functions. So commands like
```python
target-query -l
target-query -f "\*" <TARGET>
```
still return all the plugins including the ones with `findable=False`. However, this behavior feels unintentional. Plugins like `mft.records` are very important plugins during investigations, so to me it feels like they should definitely be `findable`.

As discussed, this is probably caused by the implementations of the `NamespacePlugin` class in the different plugins. However, since there are multiple plugins that have this problem, the problem is most likely a misunderstanding of how plugins of type `NamespacePlugin`should be implemented. A linting check could be added to check for the wrong usage of the `NamespacePlugin`.

JSCU-CNI:
Should fix #1180 and slightly improve #1179.


Schamper:
I want to approach PRs that tough the plugin system one at a time, so let's first resolve https://github.com/fox-it/dissect.target/pull/1135 before continuing this one.

PR #1252, opened 2025-07-24
Cross lookup windows user name from sam 

### LINKED ISSUE #1195: Windows: user name does not necessarily equal leaf of profile path
In the windows os plugin, we assume that the leaf of the profile path as read from the registry is equal to the user name. This is not necessary the case. (see https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows-client/renaming-user-account-not-change-profile-path)

Reproduce:
1. Login as local administrator
2. Rename profile directory of target user account.
3. Modify `ProfileImagePath` key in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\{SSID of user}`
4. Login as the user

Windows seems to work normally.  

Note that the information from the `sam` plugin is canonical and has the true user name.

twiggler:
Closes #1195

### DIFF
--- a/dissect/target/plugins/os/windows/_os.py
+++ b/dissect/target/plugins/os/windows/_os.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import operator
 import struct
+from functools import cached_property
 from typing import TYPE_CHECKING, Any
 
 from dissect.target.exceptions import RegistryError, RegistryValueNotFoundError
@@ -284,6 +285,7 @@ class WindowsPlugin(OSPlugin):
         except RegistryError:
             pass
 
+    @cached_property
     def _sam_by_sid(self) -> dict[str, SamRecord]:
         if not (machine_sid := next(self.target.machine_sid(), None)):
             return {}
@@ -292,8 +294,7 @@ class WindowsPlugin(OSPlugin):
         try:
             for sam_record in self.target.sam():
                 # Compose SID from domain_sid and RID
-                sid = f"{machine_sid.sid}-{sam_record.rid}"
-                sam_users[sid] = sam_record
+                sam_users[f"{machine_sid.sid}-{sam_record.rid}"] = sam_record
         except Exception as e:
             self.target.log.warning("Could not read SAM records")
             self.target.log.debug("", exc_info=e)
@@ -305,8 +306,6 @@ class WindowsPlugin(OSPlugin):
         # Be aware that this function can never do anything which needs user
         # registry hives. Initializing those hives will need this function,
         # which will then cause a recursion.
-
-        sam_users = self._sam_by_sid()
 
         key = "HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\ProfileList"
         sids = set()
@@ -327,7 +326,7 @@ class WindowsPlugin(OSPlugin):
                 else:
                     home = profile_image_path.value
                     # Use SAM username if available
-                    name = sam_users[sid].username if sid in sam_users else home.split("\\")[-1]
+                    name = self._sam_by_sid[sid].username if sid in self._sam_by_sid else home.split("\\")[-1]
 
                 yield WindowsUserRecord(
                     sid=subkey.name,
--- a/dissect/target/plugins/os/windows/_os.py
+++ b/dissect/target/plugins/os/windows/_os.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any
 from dissect.target.exceptions import RegistryError, RegistryValueNotFoundError
 from dissect.target.helpers.record import WindowsUserRecord
 from dissect.target.plugin import OperatingSystem, OSPlugin, export, internal
+from dissect.target.plugins.os.windows.credential.sam import SamRecord
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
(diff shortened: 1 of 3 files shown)


dissect/target/plugins/os/windows/_os.py
```diff
@@ -283,29 +284,50 @@ def architecture(self) -> str | None:
         except RegistryError:
             pass
 
+    def _sam_by_sid(self) -> dict[str, SamRecord]:
```
Schamper, line 287: Can we cache this?

dissect/target/plugins/os/windows/_os.py
```diff
@@ -283,29 +284,50 @@ def architecture(self) -> str | None:
...
+        if not (machine_sid := self.target.machine_sid()):
+            return {}
+
+        sam_users: dict[str, SamRecord] = {}
+        try:
+            for sam_record in self.target.sam():
+                # Compose SID from domain_sid and RID
+                sid = f"{machine_sid}-{sam_record.rid}"
+                sam_users[sid] = sam_record
```
Schamper, line 296:
```suggestion
                sam_users[f"{machine_sid}-{sam_record.rid}"] = sam_record
```

dissect/target/plugins/os/windows/generic.py
```diff
@@ -560,26 +560,34 @@ def codepage(self) -> str | None:
...
-    def sid(self) -> Iterator[ComputerSidRecord]:
-        """Return the machine- and optional domain SID of the system."""
-
+    def _machine_sid(self) -> tuple[int, datetime] | None:
+        """Return the machine SID of the system with the timestamp."""
         try:
             key = self.target.registry.key("HKLM\\SAM\\SAM\\Domains\\Account")
 
             # The machine SID is stored in the last 12 bytes of the V value as little-endian
             # The machine SID differs from a 'normal' binary SID as only holds 3 values and lacks a prefix / Revision
```
Schamper, line 569:
```suggestion
            # The machine SID differs from a 'normal' binary SID as it only holds 3 values and lacks a prefix / Revision
```

dissect/target/plugins/os/windows/generic.py
```diff
@@ -560,26 +560,34 @@ def codepage(self) -> str | None:
...
             # NOTE: Consider moving this to dissect.util.sid if we encounter this more often
             sid = struct.unpack_from("<III", key.value("V").value, -12)
-
-            yield ComputerSidRecord(
-                ts=key.timestamp,
-                sidtype="Machine",
-                sid=f"S-1-5-21-{sid[0]}-{sid[1]}-{sid[2]}",
-                _target=self.target,
-            )
+            return (f"S-1-5-21-{sid[0]}-{sid[1]}-{sid[2]}", key.timestamp)
```
Schamper, line 572: I'm not really a fan of this approach, maybe just return the `ComputerSidRecord` directly in `machine_sid`, and maybe add a `domain_sid` export too, and just chain the two in `sid`.

The previous code that uses the `machine_sid` should be changed to handle a record instead.

tests/plugins/os/windows/test__os.py
```diff
@@ -270,6 +271,30 @@ def test_windows_user(target_win_users: Target) -> None:
     assert users[1].home == windows_path("C:\\Users\\John")
 
 
+def test_windows_user_from_sam(target_win_users: Target) -> None:
```
Schamper, line 274: Can you add a `"""Tests whether..."""``` story here of why we test this?

Schamper (changes requested):

dissect/target/plugins/os/windows/generic.py
twiggler: I rewrote this as requested.

I find the pattern of parsing a final output (a flow record) as an intermediate step to be a bit awkward. It brings to mind the controller/service analogy from web development. Right now, our plugins seem to be acting as both a "controller" (producing final, formatted records) and a "service" (containing reusable logic).

Having this method return a more structured intermediate result (like a DTO) instead of a final record helps to untangle these two roles. While it's not a full refactor, it's a step toward a cleaner separation of concerns.

A concrete downside of only outputting records is that we lose structural information—for instance, the fact that domain_sid and machine_sid are singletons, which a DTO could preserve.

dissect/target/plugins/os/windows/generic.py
```diff
@@ -561,25 +562,32 @@ def codepage(self) -> str | None:
...
-
             yield ComputerSidRecord(
                 ts=key.timestamp,
                 sidtype="Machine",
                 sid=f"S-1-5-21-{sid[0]}-{sid[1]}-{sid[2]}",
                 _target=self.target,
             )
         except (RegistryError, struct.error):
-            pass
+            self.target.log.exception("Cannot read machine SID from registry.")
```
Schamper, line 582:
```suggestion
            self.target.log.exception("Cannot read machine SID from registry")
```

dissect/target/plugins/os/windows/generic.py
```diff
@@ -592,4 +600,11 @@ def sid(self) -> Iterator[ComputerSidRecord]:
                 _target=self.target,
             )
         except (RegistryError, struct.error):
-            pass
+            self.target.log.exception("Cannot read domain SIDs from registry.")
```
Schamper, line 603:
```suggestion
            self.target.log.exception("Cannot read domain SID from registry")
```

tests/plugins/os/windows/test__os.py
```diff
@@ -270,6 +272,42 @@ def test_windows_user(target_win_users: Target) -> None:
...
+    """Verify the home folder name is independent of the SAM username.
+
+    The final component of the home folder path (e.g., 'C:\\Users\\Username')
+    is not always identical to the account's username after a rename. This
+    test ensures our code correctly handles this distinction.
+
+    See for reference:
+    https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows-client/renaming-user-account-not-change-profile-path
```
Schamper, line 283:
```suggestion
    References:
        - https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows-client/renaming-user-account-not-change-profile-path
```

Issue #283, opened 2023-06-20
Windows Search Index parser


Zawadidone:
> Windows Search Indexer [[Microsoft Documentation](https://learn.microsoft.com/en-us/windows/win32/search/-search-3x-wds-overview)] is a service which enables faster searching of files, emails, and other content on Windows systems. The service builds an index that the system refers to whenever a search is run.

> Microsoft changed the structure of the Search index in Windows 11, dropping the former ESE database structure and implementing SQLite in its place. Stroz Friedberg’s research into both the old and new structures has revealed how the information in the Windows 10 Search index is mapped in the Windows 11 Search index.

## References

https://www.aon.com/cyber-solutions/aon_cyber_labs/windows-search-index-the-forensic-artifact-youve-been-searching-for/
https://github.com/strozfriedberg/sidr

DevJoost:
I'm going to work on this.

PR #1254, opened 2025-07-28
Add Windows Search Index plugin

JSCU-CNI:
Implements #283. Depends on two bugfixes in https://github.com/fox-it/dissect.sql/pull/37 and https://github.com/fox-it/dissect.esedb/pull/48.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,36 @@
+{
+    "configurations": [
+        {
+            "type": "debugpy",
+            "request": "launch",
+            "name": "target-query",
+            "module": "dissect.target.tools.query",
+            "args": [
+                "${input:target}",
+                "-f",
+                "${input:functions}"
+            ]
+        },
+        {
+            "type": "debugpy",
+            "request": "launch",
+            "name": "target-shell",
+            "module": "dissect.target.tools.shell",
+            "args": [
+                "${input:target}"
+            ]
+        }
+    ],
+    "inputs": [
+        {
+            "type": "promptString",
+            "id": "target",
+            "description": "Path to target file (e.g. /path/to/image.vmdk)"
+        },
+        {
+            "type": "promptString",
+            "id": "functions",
+            "description": "Comma separated target-query function names (e.g. hostname,users)"
+        },
+    ]
+}


dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+        ``GatherLogs/SystemIndex/SystemIndex.*.(Crwl|gthr)`` files or ``Windows-gather.db`` and ``Windows-usn.db`` files.
+
+        The difference between the fields ``System_Date*`` and ``System_Document_Date*`` should be researched further.
+        It is unclear what the field ``InvertedOnlyMD5`` is a checksum of (record or file content?). It might be possible
+        to correlate the field ``System_FileOwner`` with a ``UserRecordDescriptor``. The field ``System_FileAttributes`` should be
+        investigated further.
+
+        No test data available for indexed Outlook emails, this plugin might not be able to handle indexed email messages.
+
+        Resources:
```
Schamper, line 123:
```shell
❯ rg References: dissect | wc -l          
     130

❯ rg Resources: dissect | wc -l 
      77
```
We're not great on either but we should definitely pick a side.

I think `References:` is the one that's supported by Sphinx: https://github.com/sphinx-doc/sphinx/blob/01af04241b1305e7bdb8687c45381c5638275d2f/sphinx/ext/napoleon/docstring.py#L396

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+            if db_path.suffix == ".edb":
+                yield from self.parse_esedb(db_path, user_details)
+
+            elif db_path.suffix == ".db":
+                yield from self.parse_sqlite(db_path, user_details)
+
+            else:
+                self.target.log.warning("Unknown Windows Search Index database file %r", db_path)
+
+    def parse_esedb(self, db_path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
```
Schamper, line 142:
```suggestion
    def parse_esedb(self, path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
```

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+            # Translates e.g. ``System_DateModified`` to ``15F-System_DateModified``
+            COLUMNS = {col.split("-", maxsplit=1)[-1]: col for col in table.column_names}
+
+            for record in table.records():
+                values = {
+                    clean_col: record.get(full_col) for clean_col, full_col in COLUMNS.items() if record.get(full_col)
+                }
+                yield from self.build_record(values, user_details, db_path)
+
+    def parse_sqlite(self, db_path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
```
Schamper, line 158:
```suggestion
    def parse_sqlite(self, path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
```

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+                self.target.log.warning("Unknown Windows Search Index database file %r", db_path)
+
+    def parse_esedb(self, db_path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
+        """Parse the EseDB ``SystemIndex_PropertyStore`` table."""
+
+        with db_path.open("rb") as fh:
+            db = EseDB(fh)
+            table = db.table("SystemIndex_PropertyStore")
+
+            # Translates e.g. ``System_DateModified`` to ``15F-System_DateModified``
```
Schamper, line 149: Can you comment on why?

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+            COLUMNS = {col.split("-", maxsplit=1)[-1]: col for col in table.column_names}
```
Schamper, line 150:
```suggestion
            columns = {col.split("-", maxsplit=1)[-1]: col for col in table.column_names}
```
I don't think the capitalization is necessary.

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+
+            for record in table.records():
+                values = {
+                    clean_col: record.get(full_col) for clean_col, full_col in COLUMNS.items() if record.get(full_col)
```
Schamper, line 154: This executes `get` twice. I wonder if the tradeoff from changing this to a for-loop and eliminating that second execution is faster than the bump you get from it being a comprehension.

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,252 @@
...
+            db = SQLite3(fh)
+
+            # Contains ``WorkId``, ``ColumnId`` and ``Value``
+            table = db.table("SystemIndex_1_PropertyStore")
+
+            # ``ColumnId`` is translated using the ``SystemIndex_1_PropertyStore_Metadata`` table
+            COLUMNS = {
+                row.get("Id"): row.get("UniqueKey", "").split("-", maxsplit=1)[-1]
+                for row in db.table("SystemIndex_1_PropertyStore_Metadata").rows()
+            }
```
Schamper, line 171: Same here.

dissect/target/plugins/os/windows/search.py
Schamper: Also, doesn't this database have like thousands of columns? Wouldn't it be a lot faster if you create a little shim class that dynamically translates the column name lookup?

Schamper (changes requested):

dissect/target/plugins/general/users.py
```diff
@@ -32,14 +31,11 @@ class UsersPlugin(InternalPlugin):
...
 
-    def __init__(self, target: Target):
-        super().__init__(target)
-        self.find = lru_cache(32)(self.find)
-
     def check_compatible(self) -> None:
         if not hasattr(self.target, "users"):
             raise UnsupportedPluginError("Unsupported Plugin")
 
+    @lru_cache(maxsize=32)  # noqa: B019
```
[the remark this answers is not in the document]
Schamper: Did you try using `patch.object`? https://docs.python.org/3/library/unittest.mock.html#patch-object

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,259 @@
...
+                self.target.log.warning("Unknown Windows Search Index database file %r", db_path)
+
+    def parse_esedb(self, path: Path, user_details: UserDetails | None) -> Iterator[SearchIndexRecords]:
+        """Parse the EseDB ``SystemIndex_PropertyStore`` table."""
+
+        with path.open("rb") as fh:
+            db = EseDB(fh)
+            table = db.table("SystemIndex_PropertyStore")
+
+            # Translates e.g. ``System_DateModified`` to ``15F-System_DateModified`` as these column name prefixes might
```
Schamper, line 149: Doesn't it do the opposite? Translate `15F-System_DateModified` to `System_DateModified`?

dissect/target/plugins/os/windows/search.py
Schamper: It's probably slightly better, but I'd be curious if something like this wouldn't still be faster (pseudocode):

```python
class Shim:
    def __init__(self, table, record):
        self.table = table
        self.record = record
        self.columns = {col.split("-", maxsplit=1)[-1]: col for col in table.column_names}

    def get(self, name):
        return self.record.get(self.columns.get(name, name))
```

So translating the `.get` on the fly, instead of iterating thousands of columns only to end up reading 4 of them.

Schamper (changes requested):

Schamper (changes requested):

dissect/target/plugins/os/windows/search.py
```diff
@@ -0,0 +1,270 @@
...
+            ) not in seen:
+                seen.add(digest)
+                yield path.resolve(), None
+
+        for user_details in self.target.user_details.all_with_home():
+            for user_path in self.USER_PATHS:
+                for path in user_details.home_path.glob(user_path):
+                    if (digest := hashutil.common(path.open())) not in seen:
+                        seen.add(digest)
+                        yield path.resolve(), user_details
```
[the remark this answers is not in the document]
Schamper: There's also `TargetPath.samefile` as used in `evt.py`, don't know if that helps?

Schamper (changes requested):

PR #1258, opened 2025-07-30
Fix firefox tests for dissect.sql bugfix

JSCU-CNI:
Companion PR for https://github.com/fox-it/dissect.sql/pull/39.


Schamper:
Thanks for the excellent explanation!

Issue #1202, opened 2025-06-25
Target `name`, `path` and `path_query` are incorrect for URI targets


Schamper:
```bash
❯ uv run target-shell 'ssh://root:password@10.0.0.69/?os=default' -p
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 17.0.0 (clang-1700.0.13.3)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.31.0 -- An enhanced Interactive Python. Type '?' for help.


Loaded targets in 'targets' variable. First target is in 't'.

In [1]: t.path
Out[1]: PosixPath('password@10.0.0.69')

In [2]: t.path_query
Out[2]: {}

In [3]: t.name
Out[3]: 'password@10.0.0.69'
```

PR #1265, opened 2025-08-01
Fix various loader and unit test inconsistencies

### LINKED ISSUE #1199: URI based loaders are broken
Currently it looks like all URI-based loaders are broken, for example the SMB loader:

```bash
❯ uv run target-shell 'smb://administrator:Dissect123!@10.10.128.3'
2025-06-24T08:35:33.480732Z [error    ] administrator:Dissect123!@10.10.128.3: Failed to initiate loader: Missing URI connection details [dissect.target.target]
2025-06-24T08:35:33.481576Z [error    ] Error opening shell [dissect.target.tools.shell]
  File "dissect.target/dissect/target/tools/shell.py", line 1595, in main
    open_shell(args.targets, args.python, args.registry, args.commands)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "dissect.target/dissect/target/tools/shell.py", line 1472, in open_shell
    targets = list(Target.open_all(targets))
  File "dissect.target/dissect/target/target.py", line 374, in open_all
    raise TargetError(f"Failed to find any loader for targets: {paths}")
dissect.target.exceptions.TargetError: Failed to find any loader for targets: ['smb://administrator:Dissect123!@10.10.128.3']
```

Schamper:
Fixes #1199, #1202, #1269, and other various issues with existing loaders.

- Streamline all loader unit tests
- Add additional tests to ensure that loaders are actually picked by the loader system, rather that just a `.detect()` passing
- Add additional tests to ensure that loaders are actually picked by `Target.open`/`Target.open_all`, and that we end up with a sane `target.path`
- Make the resulting `target.path` more consistent and predictable
- Some other unit test cleanup

### DIFF
--- a/dissect/target/helpers/loaderutil.py
+++ b/dissect/target/helpers/loaderutil.py
@@ -80,7 +80,7 @@ def _try_open(fs: Filesystem, path: str) -> BinaryIO | None:
     return None
 
 
-def parse_path_uri(path: str | Path) -> tuple[Path | None, urllib.parse.ParseResult | None]:
+def parse_path_uri(path: str | Path | None) -> tuple[Path | None, urllib.parse.ParseResult | None]:
     """Parse a path URI into its path component and a parsed path, if applicable.
 
     If the path is a string, it will be parsed as a URI. If it is a Path object, it will be returned as is.
--- a/dissect/target/target.py
+++ b/dissect/target/target.py
@@ -76,7 +76,6 @@ class Target:
     """
 
     def __init__(self, path: str | Path | None = None):
-        self.path = None
         self.path_scheme = None
         self.path_query = {}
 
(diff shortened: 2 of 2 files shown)


twiggler (changes requested):
I think I found something: 

When I run with `-f network.interfaces -L local local?force-directory-fs`,
it doesn´t seem to process the `force-directory-fs`.

Could be that I am using the wrong syntax? Also tried with `force-directory-fs=1`

Schamper:
> I think I found something:
> 
> When I run with `-f network.interfaces -L local local?force-directory-fs`, it doesn´t seem to process the `force-directory-fs`.
> 
> Could be that I am using the wrong syntax? Also tried with `force-directory-fs=1`

This doesn't work on `main` either, and I think falls more into https://github.com/fox-it/dissect.target/issues/920 than this PR to fix.

PR #1270, opened 2025-08-04
Fix incorrect timestamps of `ZipInfo` objects

### LINKED ISSUE #1267: Velociraptor collection (ZIP) with incorrect timestamps leads to `ValueError`
The `ZipFilesystemEntry` in https://github.com/fox-it/dissect.target/blob/main/dissect/target/filesystems/zip.py has the following `date_time`, which is incorrect.

This leads to the exception `ValueError` when the function `datetime` is called by the function `lstat`: https://docs.python.org/3/library/datetime.html.

```python
In [10]: self.entry.date_time
Out[10]: (1980, 0, 0, 0, 0, 0)
In [11]: datetime(*self.entry.date_time, tzinfo=timezone.utc).timestamp()
[...]
File ~/.local/share/uv/python/pypy-3.9.19-macos-aarch64-none/lib/pypy3.9/datetime.py:430, in _check_date_fields(year, month, day)
    428     raise ValueError('year must be in %d..%d' % (MINYEAR, MAXYEAR), year)
    429 if not 1 <= month <= 12:
--> 430     raise ValueError('month must be in 1..12', month)
    431 dim = _days_in_month(year, month)
    432 if not 1 <= day <= dim:

ValueError: ('month must be in 1..12', 0)
```

## References
* https://github.com/kaitai-io/kaitai_struct_formats/issues/562
* https://isc.sans.edu/diary/30296

respondersGY:
Closes #1267

### DIFF
--- a/dissect/target/filesystems/zip.py
+++ b/dissect/target/filesystems/zip.py
@@ -85,11 +85,6 @@ class ZipFilesystemEntry(VirtualDirectory):
         super().__init__(fs, path)
         self.entry = entry
 
-        if self.entry.date_time[0] < 1980 or self.entry.date_time[:3] == (1980, 0, 0):
-            self.entry.date_time = (1980, 1, 1, 0, 0, 0)
-        elif self.entry.date_time[0] > 2107:
-            self.entry.date_time = (2107, 12, 31, 23, 59, 59)
-
     def open(self) -> BinaryIO:
         if self.is_dir():
             raise IsADirectoryError(self.path)
@@ -165,6 +160,11 @@ class ZipFilesystemEntry(VirtualDirectory):
             mode = stat.S_IFDIR | mode
         elif not self.entry.is_dir() and not stat.S_ISREG(mode):
             mode = stat.S_IFREG | mode
+
+        if self.entry.date_time[0] < 1980 or self.entry.date_time[:3] == (1980, 0, 0):
+            self.entry.date_time = (1980, 1, 1, 0, 0, 0)
+        elif self.entry.date_time[0] > 2107:
+            self.entry.date_time = (2107, 12, 31, 23, 59, 59)
 
         return fsutil.stat_result(
             [
--- a/dissect/target/filesystems/zip.py
+++ b/dissect/target/filesystems/zip.py
@@ -161,10 +161,11 @@ class ZipFilesystemEntry(VirtualDirectory):
         elif not self.entry.is_dir() and not stat.S_ISREG(mode):
             mode = stat.S_IFREG | mode
 
-        if self.entry.date_time[0] < 1980 or self.entry.date_time[:3] == (1980, 0, 0):
-            self.entry.date_time = (1980, 1, 1, 0, 0, 0)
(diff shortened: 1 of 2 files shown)


tests/filesystems/test_zip.py
```diff
@@ -49,8 +49,8 @@ def _create_zip(prefix: str = "", zip_dir: bool = True) -> io.BytesIO:
             cur.append(p)
             _mkdir(zf, "/".join(cur))
 
-    zf.writestr(zipfile.ZipInfo(f"{prefix}file_1"), "file 1 contents")
-    zf.writestr(zipfile.ZipInfo(f"{prefix}file_2"), "file 2 contents")
+    zf.writestr(zipfile.ZipInfo(f"{prefix}file_1", (1980, 0, 0, 0, 0, 0)), "file 1 contents")
+    zf.writestr(zipfile.ZipInfo(f"{prefix}file_2", (2107, 0, 0, 0, 0, 0)), "file 2 contents")
```
Schamper, line 53: Note that this doesn't trigger the `> 2107` case.

dissect/target/filesystems/zip.py
```diff
@@ -85,6 +85,11 @@ def __init__(self, fs: ZipFilesystem, path: str, entry: zipfile.ZipInfo):
         super().__init__(fs, path)
         self.entry = entry
 
+        if self.entry.date_time[0] < 1980 or self.entry.date_time[:3] == (1980, 0, 0):
+            self.entry.date_time = (1980, 1, 1, 0, 0, 0)
+        elif self.entry.date_time[0] > 2107:
+            self.entry.date_time = (2107, 12, 31, 23, 59, 59)
```
Schamper, line 91: Can you deal with this in `lstat` instead of here? Would rather keep the original entry untouched.

Schamper (changes requested):

dissect/target/filesystems/zip.py
```diff
@@ -161,6 +161,11 @@ def lstat(self) -> fsutil.stat_result:
         elif not self.entry.is_dir() and not stat.S_ISREG(mode):
             mode = stat.S_IFREG | mode
 
+        if self.entry.date_time[0] < 1980 or self.entry.date_time[:3] == (1980, 0, 0):
+            self.entry.date_time = (1980, 1, 1, 0, 0, 0)
+        elif self.entry.date_time[0] > 2107:
+            self.entry.date_time = (2107, 12, 31, 23, 59, 59)
```
Schamper, line 167:
```suggestion
        date_time = self.entry.date_time
        if date_time[0] < 1980 or date_time[:3] == (1980, 0, 0):
            date_time = (1980, 1, 1, 0, 0, 0)
        elif date_time[0] > 2107:
            date_time = (2107, 12, 31, 23, 59, 59)
```

And then change the line below (I can't put a suggestion there) to this:
```python
datetime(*date_time, tzinfo=timezone.utc).timestamp(),
```

Schamper (changes requested):

Schamper (approved):

PR #1275, opened 2025-08-08
Add the description field to the Windows plugin services

reynas:
Having the description field of the Windows service present in the output could be another element to pivot on or find anomalies


reynas:
@DissectBot agree

dissect/target/plugins/os/windows/services.py
```diff
@@ -20,6 +20,7 @@
         ("datetime", "ts"),
         ("string", "name"),
         ("wstring", "displayname"),
+        ("wstring", "description"),
```
Schamper, line 23:
```suggestion
        ("string", "displayname"),
        ("string", "description"),
```
`wstring` is actually just an alias for `string` these days.

Schamper (changes requested):

Schamper (approved):

Issue #1269, opened 2025-08-04
`ZipLoader` loads Velociraptor collection which should be loaded by `VelociraptorLoader`


respondersGY:
This fixes the issue but it is probably not the best way to fix the issue.

```diff 
diff --git i/dissect/target/loader.py w/dissect/target/loader.py
index 124c0c5..5088492 100644
— i/dissect/target/loader.py
+++ w/dissect/target/loader.py
@@ -222,6 +222,7 @@ register("remote", "RemoteLoader")
 register("mqtt", "MqttLoader")
 register("asdf", "AsdfLoader")
 register("tar", "TarLoader")
+register("velociraptor", "VelociraptorLoader")
 register("zip", "ZipLoader")
 register("vmx", "VmxLoader")
 register("vmwarevm", "VmwarevmLoader")
@@ -248,7 +249,6 @@ register("log", "LogLoader")
 register("overlay2", "Overlay2Loader")
 register("overlay", "OverlayLoader")
 register("phobos", "PhobosLoader")
-register("velociraptor", "VelociraptorLoader")
 register("uac", "UacLoader")
 register("smb", "SmbLoader")
 register("cb", "CbLoader")
```

PR #1283, opened 2025-08-12
Move order of Velociraptor and ZIP loaders

Matthijsy:
Due to the new ZIP loader Velociraptor ZIPs are no longer handled by the Velociraptor loader. By moving the ZIP loader below the Velociraptor loader we make sure that the Velociraptor loader can detect before the `GenericZipSubloader` is detected.


Schamper:
Closes #1269 and is also addressed by #1265.

Schamper (changes requested):
Can you make the "register" portion of `loader.py` be the same from #1265? Then that'll be a clean rebase.

Schamper (approved):

PR #1287, opened 2025-08-16
Unify webserver log path retrieval

qmadev:
Related to [#264](https://github.com/fox-it/acquire/pull/264).

The idea here is to use the ~~`_log_paths()`~~ `_get_paths()` function to retrieve the resolved log paths for the various webservers that we support. Future webservers should implement this function to automatically collect the resolved log files with Acquire. They also need to be added in the Acquire collection module.


qmadev:
Fixed the linter, checks should be good now @twiggler

PR #1298, opened 2025-08-27
Add `dissect.sql` to dependencies

qmadev:
closes https://github.com/fox-it/acquire/issues/216


Schamper (changes requested):
You can remove it from `[full]` now.

Schamper (approved):

PR #1301, opened 2025-08-27
Fix VirtualFilesystem directory member overwrites

### LINKED ISSUE #1445: Fix TarLoader and TarFilesystem directory member overwrites
Follow up on review for PR #1301

JSCU-CNI:
This PR fixes an issue in VirtualFilesystem and consequently in the tar and zip implementations (`GenericTarSubLoader`, `TarFilesystem`, `ZipFilesystem`) of dissect. When an archive file has an out-of-order member definition of a directory, the previous path is erroneously overwritten, discarding any children of that member in the process.

Steps to reproduce:
```
$ tar -tf tests/_data/loaders/tar/test-dir.tar 
folder/file
folder/

$ target-shell tests/_data/loaders/tar/test-dir.tar
test-dir.tar:/$ stat folder/file
/folder/file: No such file or directory
```

The core issue lies in the `VirtualDirectory.add()` method.

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -1018,7 +1018,7 @@ class VirtualDirectory(FilesystemEntry):
         raise TypeError(f"lattr is not allowed on VirtualDirectory: {self.path}")
 
     def add(self, name: str, entry: FilesystemEntry) -> None:
-        """Add an entry to this :class:`VirtualDirectory`."""
+        """Add an entry to this :class:`VirtualDirectory`. Overrides already existing entries with ``name``."""
         if not self.fs.case_sensitive:
             name = name.lower()
 
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -1017,10 +1017,24 @@ class VirtualDirectory(FilesystemEntry):
     def lattr(self) -> Any:
         raise TypeError(f"lattr is not allowed on VirtualDirectory: {self.path}")
 
-    def add(self, name: str, entry: FilesystemEntry) -> None:
-        """Add an entry to this :class:`VirtualDirectory`. Overrides already existing entries with ``name``."""
+    def add(self, name: str, entry: FilesystemEntry, clobber: bool = False) -> None:
+        """Add an entry to this :class:`VirtualDirectory`.
+
+        Only overrides already existing entries with ``name`` if argument ``clobber`` is set to ``True``.
+
+        Args:
+            name: Name of the entry in this directory
+            entry: FilesystemEntry to add to this directory
+            clobber: Allow overwrites (default: False)
+
+        Raises:
+            KeyError if ``name`` already exists in this directory and ``clobber`` is ``False``
+        """
         if not self.fs.case_sensitive:
             name = name.lower()
+
+        if not clobber and name in self.entries:
+            raise KeyError(f"Entry {name!r} already exists in {self!r}")
 
         self.entries[name] = entry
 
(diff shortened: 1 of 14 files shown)


dissect/target/filesystem.py
```diff
@@ -870,7 +870,7 @@ def lattr(self) -> Any:
         raise TypeError(f"lattr is not allowed on VirtualDirectory: {self.path}")
 
     def add(self, name: str, entry: FilesystemEntry) -> None:
-        """Add an entry to this :class:`VirtualDirectory`."""
+        """Add an entry to this :class:`VirtualDirectory`. Overrides already existing entries with ``name``."""
```
Schamper, line 873: Maybe we can add a `clobber: bool` argument to `add`? And it could either raise a `KeyError` or something on conflict, or at least log it.

If possible, I'd like to solve this in `VirtualFilesystem` as much as possible.

Schamper (changes requested):

JSCU-CNI:
We understand review resources are limited. What can we do to move this PR along? The core issue this PR addresses is a difficult to spot issue with archives, which is why we'd like this solution to be implemented sooner rather than later before (more) users miss artifacts.

JSCU-CNI:
This issue also occurs on zip archives created by forensic tools such as Cellebrite. The Android and iOS images of the Cellebrite CTF 2024 and 2025 demonstrate the same described behavior.

dissect/target/filesystems/tar.py
```diff
@@ -68,7 +68,11 @@ def __init__(
 
             entry_cls = TarFilesystemDirectoryEntry if member.isdir() else TarFilesystemEntry
             file_entry = entry_cls(self, rel_name, member)
-            self._fs.map_file_entry(rel_name, file_entry)
+
+            try:
+                self._fs.map_file_entry(rel_name, file_entry)
+            except KeyError as e:
+                log.debug("Skipping directory member %r in tar as %r is already mapped: %s", member, file_entry.path, e)
```
Schamper, line 75: Maybe add a clobber kwarg here too?

dissect/target/filesystems/zip.py
```diff
@@ -65,7 +65,11 @@ def __init__(
                 continue
 
             rel_name = self._resolve_path(mname)
-            self._fs.map_file_entry(rel_name, ZipFilesystemEntry(self, rel_name, member))
+            file_entry = ZipFilesystemEntry(self, rel_name, member)
+            try:
+                self._fs.map_file_entry(rel_name, file_entry)
+            except KeyError as e:
+                log.debug("Skipping directory member %r in zip as %r is already mapped: %s", member, file_entry.path, e)
```
Schamper: Maybe add a clobber kwarg here too?

Schamper (changes requested):
I only see a test for the tar filesystem, but I'd like to also see a test for VirtualFilesystem directly itself. That way, this behavior will be asserted through direct tests too, instead of only through an implementation detail of the tar filesystem.

dissect/target/filesystems/tar.py
JSCU-CNI: What do you mean? We explicitly want to prevent clobber behavior here to prevent the described issue.

Schamper: Yeah nvm, brain not braining.

Schamper (changes requested):
Only thinking of this now, but can you also add a separate test for the zip filesystem?

Just so that any individual changes to either virtual, zip or tar filesystem actually verify against this behavior, instead of silently starting to fail again.

Schamper (approved):

PR #1315, opened 2025-09-04
Fix incorrect timestamps of ZipInfo objects

william-billaud:
Some issues are still present after #1270. This PR complete to make ZipLoader more robust

Example 1
```
datetime(1980,0,1,0,0,0)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-533ec990f051> in <module>
----> 1 datetime(1980,0,1,0,0,0)

ValueError: month must be in 1..12
```
Example 2
```
datetime(1980,1,0,0,0,0)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-dd5756bd5dae> in <module>
----> 1 datetime(1980,1,0,0,0,0)

ValueError: day is out of range for month
```

### DIFF
--- a/dissect/target/filesystems/zip.py
+++ b/dissect/target/filesystems/zip.py
@@ -161,11 +161,10 @@ class ZipFilesystemEntry(VirtualDirectory):
         elif not self.entry.is_dir() and not stat.S_ISREG(mode):
             mode = stat.S_IFREG | mode
 
-        date_time = self.entry.date_time
-        if date_time[0] < 1980 or date_time[:2] == (1980, 0) or date_time[:3] == (1980, 1, 0):
-            date_time = (1980, 1, 1, 0, 0, 0)
-        elif date_time[0] > 2107:
-            date_time = (2107, 12, 31, 23, 59, 59)
+        try:
+            timestamp = datetime(*self.entry.date_time, tzinfo=timezone.utc).timestamp()
+        except ValueError:
+            timestamp = 315532800  # datetime(*(1980, 1, 1, 0, 0, 0), tzinfo=timezone.utc).timestamp()
 
         return fsutil.stat_result(
             [
@@ -177,7 +176,7 @@ class ZipFilesystemEntry(VirtualDirectory):
                 0,
                 self.entry.file_size,
                 0,
-                datetime(*date_time, tzinfo=timezone.utc).timestamp(),
+                timestamp,
                 0,
             ]
         )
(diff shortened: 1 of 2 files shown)


Schamper (changes requested):
Could you add a unittest?

william-billaud:
After some looks on how zip generate these timestamp, I suggest a more global approach by catching all ValueError and setting a default value:
* [zipfile.py#L1394](https://github.com/python/cpython/blob/57f5981d6260ed21266e0c26951b8564cc252bc2/Lib/zipfile.py#L1394) : Tuple consistency is not validated on ZipFile side so this tuple should be considered as unreliable/subject to errors.
* Zipfile timestamp are forensically speaking not really valuable, and I thinks that a clear default value is better than a crash.

dissect/target/filesystems/zip.py
```diff
@@ -161,11 +161,12 @@ def lstat(self) -> fsutil.stat_result:
...
-        if date_time[0] < 1980 or date_time[:3] == (1980, 0, 0):
-            date_time = (1980, 1, 1, 0, 0, 0)
-        elif date_time[0] > 2107:
-            date_time = (2107, 12, 31, 23, 59, 59)
+        try:
+            timestamp = datetime(*self.entry.date_time, tzinfo=timezone.utc).timestamp()
+        except ValueError:
+            # 4354819199 ==  datetime(*(2107, 12, 31, 23, 59, 59), tzinfo=timezone.utc).timestamp()
+            # 315532800 ==  datetime(*(1980, 1, 1, 0, 0, 0), tzinfo=timezone.utc).timestamp()
+            timestamp = 4354819199 if self.entry.date_time[0] >= 2107 else 315532800
```
Schamper, line 169:
```suggestion
        try:
            mtime = datetime(*self.entry.date_time, tzinfo=timezone.utc)
        except ValueError:
            mtime = datetime(*(2107, 12, 31, 23, 59, 59) if self.entry.date_time[0] >= 2107 else *(1980, 1, 1, 0, 0, 0), tzinfo=timezone.utc)
```
Maybe a little nicer than hardcoding a magic number.

dissect/target/filesystems/zip.py
```diff
@@ -177,7 +178,7 @@ def lstat(self) -> fsutil.stat_result:
                 0,
                 self.entry.file_size,
                 0,
-                datetime(*date_time, tzinfo=timezone.utc).timestamp(),
+                timestamp,
```
Schamper, line 181:
```suggestion
                mtime.timestamp(),
```

Schamper (changes requested):

PR #1327, opened 2025-09-11
Add Support for KAPEs that are in the VHD format

### LINKED ISSUE #1326: Support KAPEs that use VHDs
Hey, we export our KAPEs as VHDs instead of VHDXs. Hence, we would like to have VHD KAPEs also be supported by Dissect. To this end, I implemented a new loader that should be able to handle this circumstance.

## Semi Relevant Links
- https://ericzimmerman.github.io/KapeDocs/#!Pages\3.-Using-KAPE.md#vhd

Aevyz:
So this is my POC for supporting VHD KAPEs. Everything seems to work on my end. 

You can test this by creating a KAPE with the vhd flag and trying to run queries over it. The queries I ran, e.g. `target-query bla.kape -f os` all seem to work fine.

What I did was copy and paste the existing KAPE module and change all references from VHDX to VHD. That being said, I do have some questions regarding the code:

1. Do you prefer me to integrate the Code into the existing KAPE module or should I keep them seperate
2. Do I need to also add a PR to the docs?

~I'll get to fixing the commit message with a force push in a second~

closes #1326


Schamper:
Hi @Aevyz, thanks for the contribution! Do you think you could try to extend the existing KAPE loader with this, instead of having a separate one?

Aevyz:
@Schamper Sure, I'll modify it to be in the same file, give me a couple hours

Schamper (approved):

PR #1329, opened 2025-09-17
Take disk gaps in account when scraping

### LINKED ISSUE #1115: Scraper bugs in regionmap
Following the discussion at https://github.com/fox-it/dissect.target/pull/1103/files#r2043869029:

It seems `fh.read()` can throw `EOFError` exceptions. Unsure if this should be filed in `dissect.utils` or `dissect.hypervisor`.

```
target-qfind /tmp/example-ubuntu-server-24.04.1.vmx --needles "Ubuntu\s24\.04\.\d\sLTS" --regex --unique
<Target /tmp/example-ubuntu-server-24.04.1.vmx>
[Current disk: <VmdkContainer size=21474836480 vs=<DissectVolumeSystem serial=None>>]
8.87%Traceback (most recent call last):
  File "/tmp/dissect.target/venv/bin/target-qfind", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/tmp/dissect.target/dissect/target/tools/utils.py", line 237, in wrapper
    return func(\*args, \*\*kwargs)
  File "/tmp/dissect.target/dissect/target/tools/qfind.py", line 44, in main
    for _ in target.qfind(
             ~~~~~~~~~~~~^
        args.needles,
        ^^^^^^^^^^^^^
    ...<8 lines>...
        args.window,
        ^^^^^^^^^^^^
    ):
    ^
  File "/tmp/dissect.target/dissect/target/plugins/scrape/qfind.py", line 149, in qfind
    for _, stream, needle, offset in self.target.scrape.find(
                                     ~~~~~~~~~~~~~~~~~~~~~~~^
        list(needle_lookup.keys()), progress=progress(self.target) if not record else None
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/tmp/dissect.target/dissect/target/plugins/scrape/scrape.py", line 128, in find
    for needle, offset in find_needles(
                          ~~~~~~~~~~~~^
        stream,
        ^^^^^^^
    ...<3 lines>...
        progress=(lambda current: progress(disk, current, stream.size)) if progress else None,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/tmp/dissect.target/dissect/target/helpers/scrape.py", line 66, in find_needles
    next_block = fh.read(read_size)
  File "/tmp/dissect.target/venv/lib/python3.13/site-packages/dissect/util/stream.py", line 135, in read
    r.append(self._read(self._pos, read_len))
             ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/dissect.target/venv/lib/python3.13/site-packages/dissect/util/stream.py", line 313, in _read
    run_idx = self._get_run_idx(offset)
  File "/tmp/dissect.target/venv/lib/python3.13/site-packages/dissect/util/stream.py", line 308, in _get_run_idx
    raise EOFError(f"No mapping for offset {offset}")
EOFError: No mapping for offset 1904214016
```

This reference VM was created on VMware Workstation 17.5.2 using `ubuntu-24.04.1-live-server-amd64.iso`, following a default installation path with `Virtual disk contents are stored in a single file` and `Disk space is not preallocated for this virtual disk`.

twiggler:
closes #1115 

~~Fixed as proposed in the issue. This also requires changes to `MappingStream`.~~

~~Not ideal to have a special case for `MappingStream` in the scraper.~~

- Fixed by creating a separate mapping stream per contiguous region.
- Rewrote region mapper to handle nesting (LVM-LUKS-LVM, LUKS-LVM-LUKS)
- `disk` now always returns the backing physical disk


Schamper:
If we implement `SEEK_HOLE` in `AlignedStream` we can have a common implementation here and a specific implementation in `MappingStream`.

dissect/target/helpers/scrape.py
```diff
@@ -53,13 +56,44 @@ def find_needles(
...
+            return
+        run_idx = 0
+
     while offset < end if end is not None else True:
+        # Special case for MappingStream to skip gaps
+        if is_mapping_stream:
+            # Find the next run that covers the current offset
+            while not (runs[run_idx][0] <= offset < runs[run_idx][0] + runs[run_idx][1]):
+                if offset < runs[run_idx][0]:  # There is a gap, jump
+                    current_block = b""  # No overlap when jumping to a new run
```
twiggler, line 74: This means we won´t  detect non-contiguous needles, i.e. spread over multiple mappings.

Schamper (changes requested):

Schamper (dismissed):

Schamper (changes requested):

dissect/target/plugins/scrape/scrape.py
```diff
@@ -86,10 +87,14 @@ def create_streams(
 
             if lvm and isinstance(volume.vs, LogicalVolumeSystem):
                 if not all:
-                    # Remove the base volumes from the map
                     for source_volume in volume.disk:
                         source_disk = source_volume.disk
-                        del scrape_map[source_disk][(source_volume.offset, source_volume.size)]
+                        if isinstance(source_volume.vs, EncryptedVolumeSystem):
+                            # For encrypted volumes, delete the underlying disk and encrypted size
+                            physical_disk = source_disk.disk
```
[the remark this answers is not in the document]
twiggler: > Why is this? 

There is another level of indirection in the case of using a LVM with an `EncryptedVolumeSystem`

> Shouldn't we just change what disk is on encrypted volumes and have it be the correct object?

Can you elaborate?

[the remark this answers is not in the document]
twiggler: In the case of a  default Ubuntu installation with "encryption" enabled, the encrypted volume is part of a logical volume.

The `disk` (`source_volume`) has a decrypted volume with `vs` equal to a `EncryptedVolumeSystem`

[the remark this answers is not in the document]
Schamper: I mean like this, where we take certain attributes from a potential "parent" volume to ensure consistency:
https://github.com/fox-it/dissect.target/blob/131686cb5e9b92bb15207659c99e0e694ecafa3a/dissect/target/volumes/luks.py#L37-L43

Arguably `.disk` should be the original source disk at all times.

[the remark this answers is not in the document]
twiggler: Yes, the class of `disk` (`source_volume.disk`) is `Volume`.
I thought this was intentional, so that you get a "tree" of nested volumes, with the physical disks at the leaves.

Perhaps both use-cases are valid? 
I would have to look how the `disk` property is used throughout the code base to have an opinion on this.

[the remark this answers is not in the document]
twiggler: I am now working on a new solution which does not process LVM and LUKS in fixed batches, but processes them iteratively, resolving dependencies as we go,

JSCU-CNI:
Does this also fix qfind for raid volumes (ddf, md)?

twiggler:
> Does this also fix qfind for raid volumes (ddf, md)?

In theory, yes. 

I will handcraft a qcow2 to verify and add a unit test.

dissect/target/plugins/scrape/scrape.py
```diff
@@ -204,3 +286,96 @@ def scrape_needles_from_disks(
...
+            if backing_vol_obj not in context.volume_region_map:
+                dependencies_met = False
+                break
+
+            backing_regions_info.append(context.volume_region_map[backing_vol_obj])
+
+    except (AttributeError, TypeError):
+        dependencies_met = False
+
+    if dependencies_met:
```
Miauwkeru:
```suggestion
    if not dependencies_met:
        return False
```
if you invert the condition here, you can reduce the indentation down below. 
Tho you can also invert the `dependencies_met` logic and rename it as `dependencies_not_met` it depends in what reads better

dissect/target/plugins/scrape/scrape.py
```diff
@@ -204,3 +286,96 @@ def scrape_needles_from_disks(
...
+    if not (context.encrypted_flag and isinstance(volume.vs, EncryptedVolumeSystem)):
+        return False
+
+    backing_vol_obj = None
+    try:
+        backing_vol_obj = volume.vs.fh
+    except (AttributeError, TypeError):
+        return False
+
+    if backing_vol_obj in context.volume_region_map:
```
Miauwkeru:
```suggestion
    if backing_vol_obj not in context.volume_region_map:
        return False
```
same with the indentation allergy above.

Miauwkeru (changes requested):

Miauwkeru (dismissed):

Schamper:
Before I look at the scraping changes themselves, I suggested a change to how you determine the backing disk. Instead of trying to "fix it in post", I'd prefer if we just more accurately set it to begin with. Do you agree with this approach or not?

Schamper (changes requested):

twiggler:
> Before I look at the scraping changes themselves, I suggested a change to how you determine the backing disk. Instead of trying to "fix it in post", I'd prefer if we just more accurately set it to begin with. Do you agree with this approach or not?

I also implemented that: https://github.com/fox-it/dissect.target/pull/1329/commits/60d489bbfffb1462a1e8812a1e711f7f46581be4

Schamper:
> I think the physical disk(s) provides a view into the underlying data structure. If we move this to the constructor, the definition of "what disk we are on" is now stored in two places: inside `fh` and inside `self.disk`.

That's not how I view it, `fh` is "whatever file-like object we read this from", `disk` is the source disk. It's not guaranteed  that `fh == disk`, and there's no convention whatsoever about `fh` or what it looks like. It could be a socket too.

https://github.com/fox-it/dissect.target/commit/60d489bbfffb1462a1e8812a1e711f7f46581be4 seemed to only work out of pure chance on how some objects currently look in `dissect.volume` implementations, but that's not guaranteed.

Schamper:
> So the core issue is that the type of fh is effectively Any (Stream | Device | List[Device]), since Stream and Device don't seem to have any common ancestors. That's the root of the confusion: the variable is named fh (implying a stream), but the code requires us to handle objects that are definitely not streams.

Note that the original intent was that fh would be a stream, but for LVM-like volume systems this is a micro optimization to prevent having to parse "device" information twice.

dissect/target/plugins/scrape/scrape.py
```diff
@@ -204,3 +281,109 @@ def scrape_needles_from_disks(
...
+    context.scrape_map[volume][new_region_key] = (volume, 0)
+
+    # Add this new LV to our index
+    context.volume_region_map[volume] = (volume, new_region_key)
+    return True
+
+
+def process_encrypted_layer(context: ScrapeContext, volume: Volume) -> bool:
+    """
+    Processes an encrypted volume layer in the scraping context.
```
Schamper:
```suggestion
    """Processes an encrypted volume layer in the scraping context.
```

dissect/target/volume.py
```diff
@@ -121,12 +121,21 @@ def _volumes(self) -> Iterator[Volume]:
         raise NotImplementedError
 
     @property
+    def backing_objects(self) -> Iterator[Any]:
```
Schamper: In my opinion properties should never be generators. Why not return a list here instead?

Schamper (changes requested):

dissect/target/volume.py
twiggler: List or method is probably better because of the exhaustion, yes.

dissect/target/volume.py
```diff
@@ -121,12 +121,21 @@ def _volumes(self) -> Iterator[Volume]:
         raise NotImplementedError
 
     @property
+    def backing_objects(self) -> list[Any]:
```
Schamper: With the new approach, is this property still necessary? Especially since almost all implementations are basically the same. Saves on file changes and another property to implement for new volumes that'll basically be a copy paste of all other implementations.

Perhaps introduce a `self.fhs` and `self.disks` that's always normalized to a list, and just solve this in the constructor of this class?

dissect/target/volumes/ddf.py
```diff
@@ -16,24 +16,40 @@
...
+            disk = vol.disk if isinstance(vol, Volume) else vol
+            source_disks.setdefault(ddf_disk.anchor.DDF_Header_GUID, set()).add(disk)
+
+        for guid, devs in sets.items():
             try:
-                yield cls(devs)
+                disks = list(source_disks[guid])
+                source_fhs = [dev.fh for dev in devs]
+                yield cls(
+                    source_fhs[0] if len(source_fhs) == 1 else source_fhs,
```
Schamper: Is this distinction necessary? The constructor can take a list of `fh` too per the type hint. Same goes for `disk` and the other classes where you do something similar.

Schamper (changes requested):

dissect/target/volumes/ddf.py
twiggler: I honestly don't know. 

`disk=disks[0] if len(disks) == 1 else disks` was introduced in 83c264d4a109455a08b24b85381d055e67dec449

do you remember the reason?

Additionally, `source_disks.setdefault(ddf_disk.anchor.DDF_Header_GUID, set()).add(disk)` probably doesn't work when disk is a list, because lists are not hashable? 
So this condition seems to be hiding a latent crash — it papers over the single-PV nested case but the multi-PV nested case (e.g. LVM on a 2-disk MD RAID) still hits it.
We can use `update` and normalise to a list first

I rather enforce the invariant that `fh` and `disks` are always lists (which also tackles your comment on `backing_objects`). 

(update: rewritten for clarity)

Schamper: Oh, oops, it was me 😮‍💨. I think I thought it'd be "nice" that if there was only one disk, that `.disk` would be just that object. But in hindsight I agree that it's better than, especially in the case of volume systems that usually span multiple disks, it's better to have this always be a list.

dissect/target/volumes/lvm.py
```diff
@@ -49,13 +50,13 @@ def open_all(cls, volumes: list[BinaryIO]) -> Iterator[Self]:
                 vg_name = next(key for key, value in metadata.items() if isinstance(value, dict))
                 devices.setdefault(vg_name, []).append(dev)
 
-                disk = vol.disk if isinstance(vol, Volume) else vol
-                source_disks.setdefault(vg_name, set()).add(disk)
+                disks = source_disks.setdefault(vg_name, {})
+                for disk in to_list(vol.disk if isinstance(vol, Volume) else vol):
+                    disks[disk] = None
```
twiggler, line 61: Reviewer: dict to preserve ordering

dissect/target/volumes/ddf.py
twiggler: @Schamper I can also enforce the invariant that `fh` is always a list.

For now, I moved `backing_objects` to the base class

Schamper:
It looks like some previously applied changes were reverted (`encrypted_flag` -> `encrypted`, `_get_disk_vols` -> `_get_disk_volumes`).

.gitattributes
```diff
@@ -1 +1,5 @@
 tests/_data/** filter=lfs diff=lfs merge=lfs -text
+
+# Generator scripts are human-readable text, not binary test fixtures
+tests/_data/volumes/ddf/make_ddf.sh !filter !diff !merge text=auto
```
JSCU-CNI: What about `tests/_data/**/make_*.sh !filter !diff !merge text=auto` ?

Schamper: There's also some precedence to put things like this in `tests/_tools`: https://github.com/fox-it/dissect.evidence/blob/main/tests/_tools/ewf/generate.sh

Which would make the glob even simpler.

dissect/target/volumes/ddf.py
Schamper: For interactive use I don't think that's nice. See my last comment.

Schamper:
With https://github.com/fox-it/dissect.target/pull/1329/commits/8ebecb0cf46cb6f3b0c361dc4e28125999af4555 I suggested what I meant. `.fh` and `.disk` remains the "original input", which is also convenient for interactive use. `.fhs` and `.disks` are always list variants.

twiggler:
> With [8ebecb0](https://github.com/fox-it/dissect.target/commit/8ebecb0cf46cb6f3b0c361dc4e28125999af4555) I suggested what I meant. `.fh` and `.disk` remains the "original input", which is also convenient for interactive use. `.fhs` and `.disks` are always list variants.

With self.fhs = to_list(fh) and self.disks = to_list(self.disk) we now have two public attributes each holding the same fact: fh/fhs and disk/disks. fhs is a derived snapshot of fh, but they're stored as peer attributes with nothing keeping them in sync. Today that's safe because a VolumeSystem isn't mutated after construction, so in practice they can't diverge. My hesitation is more about communication of intent: I'd like the code to express that fh is the source of truth and fhs is derived from it, rather than leaving them as two independent-looking attributes that just happen to agree. 

_**In the interest of moving forward, and because it is a low-severity issue here, I am ok with the proposed changes**_. Generally speaking, I strongly prefer the derived-property shape. 

(Zooming out, this is a pattern I keep running into in the codebase: invariants that we rely on but don't make explicit — held by convention, comments, or "we know it won't change" rather than enforced by the code itself.)

Schamper:
Alright, that specific distinction wasn't clear to me. I mostly disagreed with the naming of the property. Take https://github.com/fox-it/dissect.target/pull/1329/changes/e2cd6e6e73d561c89cddc2c22162a1d5b459d4f9 into consideration then with that in mind.

twiggler:
> Alright, that specific distinction wasn't clear to me. I mostly disagreed with the naming of the property. Take [e2cd6e6](https://github.com/fox-it/dissect.target/commit/e2cd6e6e73d561c89cddc2c22162a1d5b459d4f9) into consideration then with that in mind.

Nice. 

I was writing something similar where `fhs` and `disks` were cached properties but I think this is better since the cost of `to_list` is negligible.

I think @Miauwkeru wanted to do some more QA. 
If it works out, i will push one final commit tomorrow, which is an integration test (real images, no mock)  of a lvm nested inside an encrypted volume.

tests/plugins/scrape/test_scrape_images.py
Schamper, line 1: Why not put this in `test_scrape.py`?

Schamper (changes requested):

tests/plugins/scrape/test_scrape_images.py
twiggler: My reasoning for the split was principled: `test_scrape.py` is unit tests (mock volume hierarchies, patched `MappingStream.add`, asserting on call args — fast, no fixtures), while `test_scrape_images.py` is integration tests (real disk images, `t.apply()`, asserting on bytes read back — slower, fixture- and dependency-bound). Keeping them apart makes it easy to run/skip the slower suite separately.

As of now, the repo doesn't draw that line though: there's no unit/integration marker to hook a split into.

**I am willing to merge them into test_scrape.py and group the integration cases with a section comment.**

For the longer term, I would advise to consider splitting the tests into unit tests and integration tests.

Schamper: The current style fits a merge better, if we want to split I'd rather do it uniformly rather than just for this one time.

`-k` exists in pytest, by the way.

Schamper (approved):

Schamper:
🥳

PR #1330, opened 2025-09-17
fix: str_to_isoformat for unformatted datetime

0xAlcidius:
### Summary
Introduce `str_to_isoformat` to format datetime strings
encountered in Windows Task Scheduler XML. This ensures
non-padded months/days and space-separated datetimes are
converted into strict ISO 8601 strings.

### Motivation
Parsing a velo package could result in the following error:
```
  File "/usr/lib/pypy/lib/pypy3.11/datetime.py", line 1852, in fromisoformat
    raise ValueError(
ValueError: Invalid isoformat string: '2025-7-29T16:34:00'
```
This PR solves that issue.

### DIFF
--- a/dissect/target/plugins/os/windows/tasks/xml.py
+++ b/dissect/target/plugins/os/windows/tasks/xml.py
@@ -97,6 +97,34 @@ def str_to_bool(string_to_convert: str) -> bool | None:
     raise ValueError(f"Invalid boolean string: '{string_to_convert}' (expected 'true' or 'false')")
 
 
+def str_to_isoformat(string_to_convert: str) -> str | None:
+    """Convert a datetime string to ISO 8601 format.
+
+    Accepts strings in the form YYYY-MM-DDThh:mm:ss
+    and returns a zero-padded ISO format string.
+
+    Args:
+        string_to_convert: The input datetime string.
+
+    Returns:
+        None for an empty string, otherwise a ISO 8601 formatted string.
+    """
+    if not string_to_convert:
+        return None
+    if "T" in string_to_convert:
+        date_part, time_part = string_to_convert.split("T")
+    elif " " in string_to_convert:
+        date_part, time_part = string_to_convert.split(" ")
+    else:
+        raise ValueError(f"Invalid datetime string: '{string_to_convert}' (expected 'YYYY-MM-DDThh:mm:ss')")
+
+    try:
+        year, month, day = date_part.split("-")
+        return f"{year}-{int(month):02d}-{int(day):02d}T{time_part}"
+    except Exception:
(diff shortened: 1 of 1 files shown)


bobkarreman:
Hi @0xAlcidius thanks for your contribution.

Can you please add some unit-tests for this case?

Maybe the fix can be much easier by parsing the datetime with datetime.strptime something like this:
```
datetime.strptime("2025-7-9T16:34:00", "%Y-%m-%dT%H:%M:%S")
```
But I'm not sure if the format is always the same else some variants could be added to the parse function. Besides that you can simply return a datetime object, no need to format it as a string.

0xAlcidius:
Thank you @bobkarreman for your comment and suggestion! I’ve noticed that Windows Scheduled Tasks can be a bit inconsistent when it comes to the StartBoundary. I did try using `datetime.strptime`, but given the ambiguity in the formats, I leaned towards string manipulation instead. Returning a datetime object would definitely be a cleaner approach, though.

Could you point me to the right place in the codebase to add the unit tests?

Schamper:
> Could you point me to the right place in the codebase to add the unit tests?

https://github.com/fox-it/dissect.target/blob/main/tests/plugins/os/windows/test_tasks.py

B0TAxy:
Should we also update the records and change the return type to datetime?

bobkarreman:
> Added a few test cases with timestamps observed in scheduled tasks. `str_to_isoformat` now also returns a datetime object as was suggested.

Please also include a test case for the case you described in the this PR ('2025-7-29T16:34:00')

bobkarreman:
> Thank you @bobkarreman for your comment and suggestion! I’ve noticed that Windows Scheduled Tasks can be a bit inconsistent when it comes to the StartBoundary. I did try using `datetime.strptime`, but given the ambiguity in the formats, I leaned towards string manipulation instead. Returning a datetime object would definitely be a cleaner approach.
> 
> Could you point me to the right place in the codebase to add the unit tests?

Parsing it with datetime.fromisoformat with some fallback logic is a cleaner way in my opinion (also fromisoformat has been improved in 3.11 so will work in almost all of the test cases).

Could be something like this:
```python
def parse(s):
    try:
        d = datetime.fromisoformat(s)
    except ValueError:
        s = s.replace(" ", "T")
        if 'Z' in s:
            d = datetime.strptime(s, "%Y-%m-%dT%H:%M:%SZ")
            return d.replace(tzinfo=timezone.utc)
        d = datetime.strptime(s, "%Y-%m-%dT%H:%M:%S")
    return d
```

0xAlcidius:
@bobkarreman Following your suggestion, I’ve refactored the function to make it cleaner while maintaining the same functionality. Additionally, I’ve added the missing test case you mentioned earlier.

yunzheng (changes requested):
Do you know if the timestamps in the tasks XML file are in UTC or local time?

PR #1332, opened 2025-09-19
Fix an issue where the ESXiPlugin shows a confusing warning when constructing the local fs

Miauwkeru:
The following warning was logged every time that the local_tgz tar file
was successfully decrypted

> "local.tgz is encrypted but static decryption failed and no dynamic
decryption available!"

Now this warning is moved so it only shows when it isn't run on a local
target.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,6 @@
 <!--
 Thank you for submitting a Pull Request. Please:
+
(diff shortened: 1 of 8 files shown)


tests/plugins/os/unix/esxi/test__os.py
Schamper: Can you assert that the correct log lines show up here

PR #1342, opened 2025-09-28
Feature/zone identifier

loaflover:
so, i cooked for a long time with this. i have a few things id like to say:

1) extracting the mft file is hard. oof

2) im not sure if i added enough test files. while i accommodated for all Zone identifier possible values, in my tests i only populated 3 of those, the (by far) most common ones.

3) i used this:
https://www.digital-detective.net/forensic-analysis-of-zone-identifier-stream/
for the possible values, but i also made it somewhat easy to add more

4) i contemplated adding this to the MFT plugin, since they are similar in behavior, but i saw that all the mft plugin records are timestamp based, while this isnt

anyway, its been a blast writing, except for the hour in which i tried mounting an MFT file like a dumbass, hope this looks good (:

### DIFF
--- a/dissect/target/plugins/filesystem/ntfs/zone.py
+++ b/dissect/target/plugins/filesystem/ntfs/zone.py
@@ -20,7 +20,7 @@ if TYPE_CHECKING:
 
 
 ZoneIdentifierRecord = TargetRecordDescriptor(
-    "filesystem/ntfs/ads/zone_identifier",
+    "filesystem/ntfs/zone",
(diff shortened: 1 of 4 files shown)


Schamper (changes requested):
It definitely fell through the cracks 😅 apologies!

loaflover:
i just finished, hope this is what you meant 🙏 

i also changed the exceptions and improved the code a bit.

thanks (:

PR #1344, opened 2025-09-30
Add timestamp to yara plugin

Matthijsy:
This PR adds a timestamp to the Yara plugin, by using the modification timestamp of the file. By adding this we can show yara results in a timeline method, allowing for easier analysis.

If there is any feedback please let me know!


Matthijsy (commented):
Rename record field name

Schamper (changes requested):
Can you update the tests? They're failing right now.

Schamper (approved):

PR #1347, opened 2025-09-30
NTDS.dit Pkugin

B0TAxy:
### Description
This merge request introduces an initial implementation of an NTDS.dit Plugin.
The parser can parse and decrypt secret records (e.g., user hashes, supplemental credentials) from an Active Directory NTDS database.

At this stage, the parser is in a late development preview:

Core functionality for parsing and decrypting secret records is available.

Support for additional record types (e.g., groups, ACLs, domain data) is not yet implemented.

Output support for BloodHound-compatible JSON is planned.

Automated tests have not been added yet — these will be included in a follow-up update.

The goal of creating this MR now is to gather early feedback and design review before finalizing the remaining features.

### Checklist
- [x] Implementation (core)
- [ ] Tests
- [ ] BloodHound export
- [ ] Full record support


joost-j:
Regarding this PR, I've managed to get it to work with my own PR (https://github.com/fox-it/dissect.database/pull/8) by changing a few things:

In the `__init__()`:
```python
self.ntds = NTDS(self.target.fs.path("/sysvol/Windows/NTDS/ntds.dit").open())
```

Changing all the `record[NAME_TO_INTERNAL["XXX"]]` references to `record.XXX` where XXX refers to the official `ldapDisplayName`, since these names are already automatically converted. Lastly, by changing the database data collection part:
```python
def _collect_pek_and_user_records(self) -> tuple[bytes, Generator[User]]:
    return next(self.ntds.lookup(objectCategory="domainDNS")).pekList, self.ntds.users()
```

and some other minor adjustments. Even on a small dataset that already seemed to result in a performance boost, since the `NTDS` class uses indexes to look for the `pekList` and `User` records. 

Note: not all accounts seem to have entries for `dBCSPwd`, `ntPwdHistory`, `lmPwdHistory` and even `unicodePwd`, so that probably needs some try/exception wrapping.

Please let me know what you think of this and if you need any help with refactoring & editing the code. You can also DM me on the [Dissect discord ](https://discord.gg/tS48YbzXN7)

B0TAxy:
Hi @joost-j ,  
Thanks for the suggestion! I’ve converted the PR to use your approach, and it’s working well with the changes you mentioned. I’ve also wrapped the optional attributes (dBCSPwd, ntPwdHistory, lmPwdHistory, unicodePwd etc) as you noted. Really appreciate the guidance — the adjustments to record.XXX and the NTDS data collection logic make the code cleaner and improve performance.

joost-j:
Cool! Good to hear - a quick check from my side also confirms that it works well. When running your `ntds.user_accounts` plugin I can see the NTDS `User` class attributes, as well as the hashes, really nice! Hoping to wrap up the `NTDS` parser this year, so that this PR can be finished as well. Just a small remark; I did need to change the `from Cryptodome.Cipher` on line 9 to `from Crypto.Cipher`.

joost-j:
PR https://github.com/fox-it/dissect.database/pull/8 is now ready to be reviewed, and probably will be reviewed somewhere in the upcoming weeks. The PR now also includes most of the decryption routines for PEK lists etc. Once that PR is merged, you can edit your plugin once more to focus mainly on the plugin functionality and record publishing parts, leaving out most of the internals because you can now mostly wrap around those functions from `dissect.database`. I'll ping you once the PR has been merged!

B0TAxy:
> > I’d be happy to run a few tests on my end and give you some feedback on how it handles the extraction.
> 
> If you have any test sets of old domains that are suitable for open-source sharing, that would be greatly appreciated!

Unfortunately, I don't currently have any test data from old domains that would be suitable for open-source sharing.

Schamper:
Can you add the NTDS.dit file you're using as a test file under `tests/_data/plugins/os/windows/ad/ntds.dit`? GitHub does not allow LFS uploading to forks for maintainers (us).

That way, we can add a unit test. Maybe use the GOAD one as that one has credential history?

B0TAxy:
Hello, apologies for the delay. I encountered several challenges with the tests, but I have made progress nonetheless. Here is a summary of the status and open items:

1. **Fixture Mapping:** I initially had trouble getting the fixture to work directly with `fs_win` (persisting "file not found" errors). I have currently worked around this by mapping through the target, but we may want to review this approach.

2. **Test Data Needed:**
* I need the **known/ground truth data** for the `goad` `ntds.dit` file to verify that my parser is extracting the correct information.
* I also need the **known/ground truth data** for Computer Accounts for both the `large` and `goad` datasets.

3. **Bugs Requiring Review:**
* **Hash History Issue:** There appears to be a bug in the hash history extraction. Occasionally, a hash is longer than 16 bytes, causing the DES removal step to fail. I have added a `TODO` and temporarily set it to return `b""` to allow the rest of the code to run.
* **`member_of` Exceptions:** The `member_of` extraction is throwing multiple exceptions. I have added a temporary catch-all block and a `TODO` there as well. I would appreciate clarification on whether this behavior is intended or if it is a bug.

**Could you please look into both the hash history length issue and the `member_of` exceptions?** I would really appreciate your help debugging both of these specific issues. If you do not have time, let me know, and I can investigate further next week.

Schamper:
Except for the changed passwords (which I unfortunately don't remember what I changed them too :)) all data is the same as here: https://github.com/Orange-Cyberdefense/GOAD/blob/main/ad/GOAD/data/config.json

PR #1349, opened 2025-10-01
Add support for unarchived Acquire directories

qmadev:
Doing this does not work:

```
(dissect.target) $ tar -xvf MSEDGEWIN10_20220708124036.tar

# Before changes
(dissect.target) $ target-query -f hostname,domain,ips,version fs                                                             [4:58:57]
2025-10-01T02:59:14.990461Z [warning  ] <Target fs>: Failed to find OS plugin, falling back to default [dissect.target.target]
<Target fs> None [] None

# After changes
(dissect.target) $ target-query -f hostname,domain,ips,version fs                                                             [5:01:58]
2025-10-01T03:02:02.396263Z [warning  ] <Target fs>: Unknown drive letter for sysvol [dissect.target.target]
<Target fs> MSEDGEWIN10 ['192.168.150.128'] Windows 10 Enterprise Evaluation (NT 10.0) 17763.379
```

This PR adds a fix for that. The reason seems to be that Acquire still has both a "windows" and "Windows" folder.


Miauwkeru:
Hi @qmadev, I was unavailable for a bit. I have been looking into this issue and i think it might be a bit deeper than expected. One solution I can think of at the top of my head is adding layers to the root filesystem if it finds more than one entry with the "same" name. But I am not yet sure this is the best way to tackle this issue. I think it requires a bit more research. I will try to get it to be discussed soon

PR #1351, opened 2025-10-07
Add output redirection to target-shell

twiggler:
Support output direction in target-shell

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -290,8 +290,7 @@ class ExtendedCmd(cmd.Cmd):
                 print("Syntax error: missing filename after '>'")
                 return False
             # If there are pipes, > must be after the last |
-            last_pipe_idx = max((i for i, v in enumerate(argparts) if v == "|"), default=-1)
-            if last_pipe_idx != -1 and redirect_idx < last_pipe_idx:
+            if "|" in argparts[redirect_idx:]:
                 print("Syntax error: output redirection must come after the last pipe")
                 return False
             redirect_file = argparts[redirect_idx + 1]


dissect/target/tools/shell.py
```diff
@@ -424,7 +464,11 @@ def _get_targetrc_path(self) -> pathlib.Path:
         """Get the path to the run commands file."""
 
         return pathlib.Path(
-            getattr(self.target._config, self.CONFIG_KEY_RUNCOMMANDS_FILE, self.DEFAULT_RUNCOMMANDS_FILE)
+            getattr(
+                self.target._config,
+                self.CONFIG_KEY_RUNCOMMANDS_FILE,
+                self.DEFAULT_RUNCOMMANDS_FILE,
+            )
```
Schamper, line 471: What happened with all these reformats? Can you leave those out?

dissect/target/tools/shell.py
```diff
@@ -268,30 +269,66 @@ def emptyline(self) -> None:
...
         if "|" in argparts:
             pipeidx = argparts.index("|")
             argparts, pipeparts = argparts[:pipeidx], argparts[pipeidx + 1 :]
             try:
-                with build_pipe_stdout(pipeparts) as pipe_stdin:
-                    return func(argparts, pipe_stdin)
+                # If redirect, open file for writing and pass as stdout
+                if redirect_file:
+                    with (
+                        Path(redirect_file).open("w") as f,
```
Schamper, line 310: Not binary mode?

Schamper (changes requested):

dissect/target/tools/shell.py
twiggler: Good catch

dissect/target/tools/shell.py
```diff
@@ -269,18 +270,49 @@ def emptyline(self) -> None:
...
 
+        # Enforce that output redirection (>) only appears after the last pipe (|)
+        redirect_idx = None
+        redirect_file = None
+        if ">" in argparts:
+            redirect_indices = [i for i, v in enumerate(argparts) if v == ">"]
+            # Only support single output redirection
+            if len(redirect_indices) > 1:
+                print("Syntax error: multiple output redirections specified")
+                return False
```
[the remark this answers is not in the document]
twiggler: Thanks for the suggestion!

 I see what you mean about using a slice to check what comes after. 
 
 I'm going to stick with the current implementation, though, mainly for clarity. 
 
 I feel the len(redirect_indices) > 1 check corresponds very directly to the "multiple redirections" error message, which makes the code's intent a bit more self-documenting. 
 
Also, this approach determines the count before I try to access any specific index, which feels a little more robust."

PR #1360, opened 2025-10-14
Support environment variable for keychain values

### LINKED ISSUE #898: Add ability to register keychain keys through an environment variable
This allows scripts and tools that do not explicitly use the keychain API to benefit from transparent Bitlocker and LUKS decryption.

twiggler:
Close #898


Schamper:
> The usecase that you have in mind would require that the environment variable is parsed in the `keychain` module. 
> 
> Or do you see another way?

That is the intended solution, yes.

Schamper (approved):

PR #1380, opened 2025-10-29
Add support to zstd compress vmtar file and replace zstandard with backport.zstd

### LINKED ISSUE #1296: Add support for zstandard compressed ESXi boot modules
ESXi9 started compressing boot modules with zstandard (with a signing footer) instead of LZMA. We should add support for that: https://github.com/fox-it/dissect.target/blob/eaa344004e3fe4437a35f4321ae54dc6534a8d87/dissect/target/plugins/os/unix/esxi/_os.py#L282

Related to #1191.

william-billaud:
* Add support to zstd compressed vmtar files (and XZ, as fsutil.open_decompress is now used to manage decompress)
* Use VmtarFilesystem inside esxi._os, instead of manual tar uncompress
* Add the backport.zstd dependency. In Python 3.14 zstd as been added to standard library (pep-0784), author as backported the lib.
  * Using this lib, we do not need to fully read data in memory when uncompressing zstd, which is more convenient.
* Add an UncompressionError, mainly raised when required dependency is missing. This allows logging when a file couldn't be decompressed due to an error/missing deps
* Initially, using zstandard I encounter uncompression Errors of few zstd file, but this error is not present using backport.zstd (and files from "faulty" archive are present when mounting the disk).
* Add test regarding a zstd compressed vmtar file.

Some note regarding somme "issues" I found into the code
* Perhaps zstandard usage should be fully replaced by backport.zstd/zstd in a dedicated issue
* Perhaps a decompression error should be generated whenever a compressed stream is detected using `open_decompress`, but the required library is not present (or when decompression has failed). This would warn the user of something abnormal without causing a crash.

I can provide an esxi 8 and esxi 9 disk (30Go, less than 1Go compressed) if needed for testing purposes.

close #1296

### DIFF
--- a/dissect/target/plugins/os/windows/defender/_plugin.py
+++ b/dissect/target/plugins/os/windows/defender/_plugin.py
@@ -39,6 +39,8 @@ if TYPE_CHECKING:
 
     from flow.record import Record
 
+    from dissect.target.target import Target
+
 DEFENDER_EVTX_FIELDS = [
     ("datetime", "ts"),
     ("uint32", "EventID"),
@@ -102,6 +104,12 @@ EVTX_PROVIDER_NAME = "Microsoft-Windows-Windows Defender"
 
 DEFENDER_QUARANTINE_DIR = "sysvol/programdata/microsoft/windows defender/quarantine"
 DEFENDER_MPLOG_DIR = "sysvol/programdata/microsoft/windows defender/support"
+DEFENDER_MPCMDRUN_SYSTEM_DIRS = (
+    "sysvol/Windows/Temp",
+    "sysvol/Windows.old/Windows/Temp",
+    "sysvol/Windows/Microsoft Antimalware/Tmp",
+)
+DEFENDER_MPCMDRUN_USER_DIRS = ("AppData/Local/Temp",)
 DEFENDER_KNOWN_DETECTION_TYPES = [b"internalbehavior", b"regkey", b"runkey"]
 
 DEFENDER_EXCLUSION_KEY = "HKLM\\SOFTWARE\\Microsoft\\Windows Defender\\Exclusions"
@@ -117,6 +125,16 @@ DefenderExclusionRecord = TargetRecordDescriptor(
         ("datetime", "regf_mtime"),
         ("string", "type"),
         ("string", "value"),
+    ],
+)
+
+DefenderMpCmdRunLogRecord = TargetRecordDescriptor(
+    "filesystem/windows/defender/mpcmdrunlog",
+    [
+        ("datetime", "ts_start"),
+        ("datetime", "ts_end"),
+        ("string", "command"),
+        ("path", "source"),
     ],
 )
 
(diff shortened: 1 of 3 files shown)


tests/_data/filesystems/vmtar/esxi9_zstd_intelgpi.v00
Schamper: Could you make a separate test file with some dummy data? If I remember correctly, the `vmtar` utility also allows creation of files. It'd be nice if we didn't have to include (probably) copyrighted data as a test file 😄.

dissect/target/helpers/fsutil.py
```diff
@@ -564,11 +571,10 @@ def open_decompress(
...
 
-    if HAS_ZSTD and magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
-        # stream_reader is not seekable, so we have to resort to the less
-        # efficient decompressor which returns bytes.
-        return io.BytesIO(zstandard.decompress(file.read()))
-
+    if magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
+        if HAS_ZSTD:
+            return zstd.open(file, mode=mode, encoding=encoding, errors=errors, newline=newline)
+        raise UncompressionError("fail to uncompress zstd file, missing backport.zstd dependency")
```
Schamper:
```suggestion
    if HAS_ZSTD and magic[:4] in [b"\xfd\x2f\xb5\x28", b"\x28\xb5\x2f\xfd"]:
        return zstd.open(file, mode=mode, encoding=encoding, errors=errors, newline=newline)

```
Maybe just keep this like this for the time being (similar to the other branches) and remove the separate error. I've made a separate issue to consider making this consistent with all other compression algorithms: https://github.com/fox-it/dissect.target/issues/1410

dissect/target/filesystems/vmtar.py
```diff
@@ -11,7 +11,15 @@
...
-    def __init__(self, fh: BinaryIO, base: str | None = None, *args, **kwargs):
+    def __init__(self, fh: BinaryIO, base: str | None = None, uncompress: bool = True, *args, **kwargs):
+        """
+        Load a vmtar files (modified tar version used on ESXi)
+        In most case vmtar files are compressed using zstd, xz or LZMA, and then using gzip
+
+        Args:
+            fh: Handler to vmtar data stream
+            base: TarFilesystem base
+        """
```
Schamper:
```suggestion
        """Load a vmtar file (modified tar version used on ESXi).
        
        In most case vmtar files are compressed using zstd, xz or LZMA, and then using gzip.

        Args:
            fh: File-like object of a vmtar file.
            base: Base path in the tar file to open the filesystem on.
        """
```

tests/_data/filesystems/vmtar/esxi9_zstd_intelgpi.v00
william-billaud: I will take a look, but vmtar doesn't really support zstd (see. https://github.com/fox-it/dissect.target/issues/1296#issuecomment-3460654431).

william-billaud:
Regarding zstandard replacement it should be good. For the plocate plugins it is a bit more complicated as it need to read frame by frame, which is only supported for byte-like objects.

zstandard is still required due to dependencies in `dissect.btrfs` in `dissect.squashfs`

PR #1385, opened 2025-11-03
Add support for ESXI auth, hostd and shell logs.

### LINKED ISSUE #1218: Explore ESXi logs for new parsers
## References

* [https://www.aon.com/en/insights/cyber-labs/parsing-esxi-logs-for-incident-response?collection=5b76135e-4196-415b-ab1d-f42b6f0abb10](https://www.aon.com/en/insights/cyber-labs/parsing-esxi-logs-for-incident-response?collection=5b76135e-4196-415b-ab1d-f42b6f0abb10)
* [https://github.com/strozfriedberg/qelp](https://github.com/strozfriedberg/qelp)
* [https://www.synacktiv.com/en/publications/vmware-esxi-forensic-with-velociraptor](https://www.synacktiv.com/en/publications/vmware-esxi-forensic-with-velociraptor)

william-billaud:
Add support for three ESXi logs with related test data from ESXi 6/7/8/9:
* auth.log
* shell.log
* hostd.log

I may take a look on some others files later (vodb, vmk*, vpxa...)  but it may be easier to validate the structure before adding more files. I'm not sure the plugin inheritance is the best solution, but if each log type is just an entry of the same plugins, is that we can't have a `check_compatible` function by log type.

One of the main problems is that even if most files seems to have the same format, some small differences exists (e.g in shell.log). Regarding plugins name I'm not really sure how to work with name conflict (eg : auth).

Regarding log file location, as it may change regarding version and how log were collected (ESXi does some weird thing with symlinks) plugins look for usual location, as well as in the osdata partition.

Log location may also be found in the `/etc/vmsyslog.conf` but I have not implemented, neither worked on how it works.

Furthermore : 

* Add files related to pycharm, uv and coverage to .gitignore

This issue partially resolve #1218


Schamper (changes requested):

dissect/target/plugins/os/unix/esxi/log.py
```diff
@@ -0,0 +1,226 @@
...
+        ("string", "op_id"),
+        ("string", "user"),
+        ("string", "event_metadata"),
+        ("string", "message"),
+        ("path", "source"),
+    ],
+)
+
+
+class EsxiLogBasePlugin(Plugin, ABC):
```
Schamper: Would you mind changing this around to a more functional approach, rather than ABC? So e.g. a function in the global scope `find_logs(target: Target, name: str)` or something like that. As well the parser code.

Schamper (changes requested):

dissect/target/plugins/os/unix/esxi/log.py
william-billaud: done. Let me know if it's OK for you

dissect/target/plugins/os/unix/esxi/_os.py
```diff
@@ -144,6 +144,16 @@ def version(self) -> str | None:
...
 
+    @internal
+    def osdata_fs(self) -> TargetPath | None:
+        """
+        return path to osdata vmfs, present in esxi 7+
+        """
+        os_data = list(self.target.fs.path("/vmfs/volumes/").glob("OSDATA-*"))
+        if os_data:
+            return os_data[0]
+        return None
```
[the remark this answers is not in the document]
william-billaud: See comment below. 
Removed.

dissect/target/plugins/os/unix/esxi/esxi_log/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                target.info.warning("Fail to resolve path to %s : %s", path, str(e))
+    return
+
+
+def yield_log_records(
+    target: Target, log_paths: list[Path], re_log_format: re.Pattern, logname: str
+) -> Iterator[ESXiLogRecord]:
+    """Yield parsed log entries, iterate on identified log files."""
+    for path in log_paths:
+        try:
```
cochaviz: Maybe it would be better to put this try catch outside this function? Currently the function is quite deeply nested making the function hard to read.

cochaviz (changes requested):
Thanks all the work on this huge plugin! I'm new to the code base, so if there are any things that jump out as obviously incorrect let me know.

With that, I focused mainly on code styling:

- The nesting depth making `yield_log_records` really hard to read: try to 'shortcut' various paths by starting the for loop with a number of cases in which the line can be skipped, and we should consider moving the try-catch block out of that function. 
- Really large regular expressions: I would split these up into separate chunks (check requested changes).

dissect/target/plugins/os/unix/esxi/esxi_log/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                            else datetime.strptime(ts, "%Y-%m-%dT%H:%M:%S%z"),
+                            source=path,
+                        )
+                else:
+                    target.log.warning("log file contains unrecognized format in %s", path)
+                    target.log.debug("log file contains unrecognized format in %s : %s", path, line)
+                    continue
+            if current_record:
+                yield current_record
+        except Exception as e:
```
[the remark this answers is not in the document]
william-billaud: regarding the generic catch on Exception, it's a pattern widely used on the code base (235 occurrence over 118 files).
It's allows to process others files, even in an unexpected error occur, but yes it's not really necessary, I won't object to remove it.

cochaviz (approved):
Super, nice! I think the code is much more readable.

Edit: The only thing left that I can tell is code coverage, but I'll check if I can contribute to those tests tonight!

Edit Ep.2: The only missing test should have to do with the various exception and warnings that could be raised: https://app.codecov.io/gh/fox-it/dissect.target/pull/1385

PR #1387, opened 2025-11-06
Add magic helper module

JSCU-CNI:
This PR open sources our MIME type identification helper functions for dissect using magic bytes and path suffix identification. The knowledgebase is derived from [FreeDesktop's shared mime info project](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in) which provides a somewhat structured base of file signatures. Tests are based on [Wikipedia's list of file signatures](https://en.wikipedia.org/wiki/List_of_file_signatures).

This implementation provides a base to build future improvements on. The current implementation does not support nested or logical (AND/OR) magic header definitions or `io.SEEK_END` offsets.

Once this module is merged this functionality can be embedded in plugins such as `walkfs` to return a mimetype for each file.


dissect/target/helpers/magic/magic.py
```diff
@@ -0,0 +1,102 @@
...
+                    return mimetypes.TYPES[index][res_attr]
+
+        return None
+
+
+def from_file(path: TargetPath | FilesystemEntry, *, mime: bool = False) -> MagicResult:
+    """Detect file type from a :class:`TargetPath` or :class:`FilesystemEntry` instance."""
+
+    if not isinstance(path, (TargetPath, FilesystemEntry)):
+        raise TypeError("Provided path is not a TargetPath or FilesystemEntry")
```
Schamper: This can just be any `Path`-like, right? Not specifically `TargetPath`.

dissect/target/tools/build_magic.py
```diff
@@ -0,0 +1,210 @@
...
+    parser.add_argument("path", metavar="PATH", type=Path, help="Path to freedesktop.org.xml.in file")
+    parser.add_argument("--output", "-o", type=Path, help="Optional output path for generated types document")
+    parser.add_argument("--format", action="store_true", default=False, help="Format output using ruff")
+    args = parser.parse_args()
+
+    if not args.path.is_file():
+        print(f"Provided file does not exist: {args.path}")
+        return 1
+
+    output = (args.output or Path("dissect/target/helpers/magic/mimetypes/freedesktop.py")).resolve()
```
Schamper: This assumes that the tool is executed from the root of the repository. While this is a "development" tool, maybe a one line sanity check of that would be nice to add. I.e. checking that the target directory/path exists, and printing are you sure you're in the repository root. Or basing it on the path of `__file__`.

Schamper (changes requested):

Schamper (changes requested):

dissect/target/helpers/magic/magic.py
```diff
@@ -0,0 +1,112 @@
...
+        raise TypeError("Provided fh does not have a read or seek method")
+
+    return from_buffer(fh, suffix, mime=mime)
+
+
+# Convenience alias, not present in python-magic.
+from_fh = from_descriptor
+
+
+def from_buffer(buf: bytes | BinaryIO, suffix: str | None = None, *, mime: bool = False) -> MagicResult:
```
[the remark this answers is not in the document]
JSCU-CNI: Strange indeed but that is because we follow the `python-magic` implementation (https://github.com/ahupp/python-magic/blob/master/magic/__init__.py#L224). Feel free to change this as you see fit.

Schamper (approved):

PR #1390, opened 2025-11-06
Fix TargetPath once and for all (famous last words)

### LINKED ISSUE #1376: Add TargetPath support for Python 3.14
TargetPath is currently incompatible with 3.14.

Schamper:
Closes https://github.com/fox-it/dissect.target/issues/1376, depends on #1389.

Globbing is still a pain in the ass to fix, but at least it looks like 3.15 will have a nicer interface for this.


dissect/target/helpers/compat/path_314.py
```diff
@@ -0,0 +1,586 @@
...
+        return f"{cls.__name__}.{attribute} is unsupported"
+
+    def get(self) -> FilesystemEntry:
+        """Return the :class:`FilesystemEntry` for this path."""
+        if not hasattr(self, "_entry"):
+            self._entry = self._direntry.get() if hasattr(self, "_direntry") else self._fs.get(str(self))
+        return self._entry
+
+    @property
+    def info(self) -> TargetPathInfo:
```
Miauwkeru:
```suggestion
    def info(self) -> TargetPathInfo | DirEntryInfo:
```

Miauwkeru (changes requested):
I tried out the changes a bit, didn't see anything that broke compared to before. I did notice `TargetPath.from_uri` doesn't work, but that is also the case in older versions. Where the question ofc is, when would you ever use it.

Schamper:
I've decided to solve this problem with a nuke and just rewrite all path parsing logic and the core principle of how we did `TargetPath`. I'd appreciate critical eyes on this to verify if I've missed any common or expected path handling convention.

To quote the docstring in the new `compat/pathlib.py`:

> Historically, we replied on subclassing pathlib classes directly and applying some patches to make it work.
However, this approach has become increasingly difficult to maintain as the pathlib implementation in the
standard library changes dramatically between CPython versions, and there's often a big delay between the
release of a new CPython version and the update of our code to support it.
> 
> (Yes, I burned my hands on the fire I was playing with)
> 
> For that reason, we've decided to basically copy-paste the entirety of pathlib's implementation from CPython's main branch, and then apply the necessary changes to make it work with dissect.target. This has several benefits:
> - Since we no longer rely on subclassed standard library code, there should be no more issues with running on newer CPython versions
> - Newer pathlib features will become available even on older CPython versions, as we will always be running the latest pathlib code on all CPython versions
> 
> There will still be the burden of having to manually update our code with changes from the standard library,
but at least in this way it _shouldn't_ break on newer CPython versions. Worst case scenario, we miss some
new features, optimizations or bug fixes, but at least it won't break entirely.
> 
> To ease updating this code, the order of functions, comments and code style is kept largely the same as the
original pathlib code, except for linting and formatting rules.
> 
> Changes will be marked with a comment starting with `# PATCH`. The largest structural changes being:
> - Instead of having just two variants (Posix and Windows), we have a polypath parser that deals with both styles of separators and case sensitivity
> - The "pure" path (PureTargetPath) is mostly copied straight from PurePath, but has a few important distinctions (most of which derive from polypath rules):
>   - All paths are normalized to POSIX-style paths internally
>   - For Windows-style paths, we also consider paths starting with a separator to be absolute, and we have a few more possible "drive letter" names (see polypath.ALLOWED_DRIVE_NAMES)
> - All variations of path (i.e. the product of all case sensitivity and separator styles) are pre-created as subclasses, and the correct class will be selected during TargetPath initialization

The remaining 3.14 test failures are because of https://github.com/fox-it/dissect.evidence/pull/42

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+    @property
+    def stem(self) -> str:
+        """The final path component, minus its last suffix."""
+        name = self.name
+        i = name.rfind(".")
+        if i != -1:
+            stem = name[:i]
+            # Stem must contain at least one non-dot character.
+            if stem.lstrip("."):
+                return stem
```
Miauwkeru: # FIXME: What happens to a dotfile?

dissect/target/filesystem.py
```diff
@@ -50,29 +50,46 @@ class Filesystem:
     def __init__(
         self,
         volume: BinaryIO | list[BinaryIO] | None = None,
-        alt_separator: str = "",
+        *,
+        sep: str = "/",
+        altsep: str | None = None,
```
JSCU-CNI: Could breaking changes like these (and perhaps others, i.e. `fsutil.*` kwargs) be added to some form of a `CHANGELOG.md` file? Perhaps a summary of (any) behavioral changes in `TargetPath` as well.

dissect/target/filesystem.py
```diff
@@ -50,29 +50,46 @@ class Filesystem:
...
 
     def __repr__(self) -> str:
         return f"<Filesystem type={self.__type__}>"
 
+    @property
+    def altsep(self) -> str:
+        """The alternative separator used to distinguish between directories in a path."""
+        if self._altsep is None:
+            return "/" if self.sep == "\\" else ""
+        return self._altsep
```
JSCU-CNI: Is the `altsep` value of `""` native pathlib behavior?

tests/_data/plugins/os/unix/linux/android/another.prop
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ed459e01c1c895deb4f259649e7d9f4135a53fdfc1d153a6d794f88ce3f89e4f
-size 13
+oid sha256:b5a2753336112f49251becf688e2e1a059be7c4efbeaf45ef22eacd1ce3a01f1
+size 11
```
JSCU-CNI, line 3: This looks like an unrelated change for the scope of this PR?

tests/helpers/test_fsutil.py
```diff
@@ -31,186 +31,399 @@
...
+        pytest.param(
+            "/some/dir/some/file", "/SOME/DIR", "/", True, "../../some/dir/some/file", id="posix-case-sensitive"
+        ),
+        pytest.param("/some/dir/some/file", "/SOME/DIR", "/", False, "some/file", id="posix-case-insensitive"),
+        pytest.param(
+            "/some/dir/some/file", "/SOME/DIR", "\\", True, "..\\..\\some\\dir\\some\\file", id="windows-case-sensitive"
+        ),
+        pytest.param("/some/dir/some/file", "/SOME/DIR", "\\", False, "some\\file", id="windows-case-insensitive"),
+        # Ensure that drive letter paths are handled according to POSIX rules
+        pytest.param("C:/some/file", "D:/some/file", "\\", None, "..\\..\\..\\C:\\some\\file", id="windows-drive"),
```
JSCU-CNI: It might be worth pointing out this non-pathlib behavior in the docstring of relpath.

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+    def __next__(self, *args) -> FilesystemEntry:
+        return next(self._iterator, *args)
+
+    def close(self) -> None:
+        # close() is not defined in the various filesystem implementations. The
+        # python ScandirIterator does define the interface however.
+        pass
+
+
+class _PolyParser:
```
JSCU-CNI, line 129: Does this not belong in `polypath.py`?

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+    def __repr__(self) -> str:
+        # PATCH: hardcode TargetPath as class name if it's a subclass of TargetPath
+        if isinstance(self._path, TargetPath):
+            return "<TargetPath.parents>"
+        # Otherwise leave the original class name to ease debugging
+        return f"<{type(self._path).__name__}.parents>"
+
+
+# To be compatible with things like isinstance checks, we _do_ actually subclass from stdlib pathlib
+# However, since we basically override everything that makes PurePath works, it shouldn't cause any issues
```
JSCU-CNI, line 203: Famous last words 😄 But an understandable tradeoff nonetheless.

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+
+    def _mod_time_ns(self, *, follow_symlinks: bool = True) -> int:
+        """Return the modify time in nanoseconds."""
+        return self._stat(follow_symlinks=follow_symlinks).st_mtime_ns
+
+    def _bsd_flags(self, *, follow_symlinks: bool = True) -> int:
+        """Return the flags."""
+        return self._stat(follow_symlinks=follow_symlinks).st_flags
+
+    # PATCH: remove xattr support for now
```
JSCU-CNI: Could you create a ticket for this? Seems like something we'd want down the line for at least NTFS and EXTFS (and perhaps xfs, btrfs, apfs). Perhaps use `UnsupportedOperation` for now?

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+        """Return the path as a URI."""
+        return super().as_uri()
+
+    @classmethod
+    def from_uri(cls, uri: str) -> Self:
+        """Return a new path from the given 'file' URI."""
+        f = f"{type(cls).__name__}.from_uri()"
+        raise UnsupportedOperation(f"{f} is unsupported in Dissect")
+
+    # Substitute methods and properties that we do not support with ones that raise errors
```
JSCU-CNI:
```suggestion
    # PATCH: Substitute methods and properties that we do not support with ones that raise errors
```

JSCU-CNI (changes requested):
Overall this looks like a better approach for `TargetPath` in the long term. I agree with the POSIX-first approach for NT pathlib behavior. The performance improvement is a nice bonus.

dissect/target/filesystem.py
Schamper: This was an assumption from me but upon further inspection, it's actually `None` 🙃:
https://github.com/python/cpython/blob/b3cfd36ba4f17c19d587427b90ca9378ad009978/Lib/posixpath.py#L22

Will change that.

tests/_data/plugins/os/unix/linux/android/another.prop
Schamper: Yes and no. This test line actually never triggered because `TargetPath` did not support string comparison before:
https://github.com/fox-it/dissect.target/blob/660b0102ea26868d5b9b3549181f9765eeeccb01/tests/plugins/os/unix/linux/android/test__os.py#L58-L59

So to make this assert pass, I had to fix the test data. (`s/"bar"/bar/`).

dissect/target/filesystem.py
Schamper: Yes that's the plan. I was all out of energy for that at the moment though.

dissect/target/helpers/compat/pathlib.py
Schamper: It's more closely associated with the pathlib implementation (an evolution of the "flavours" in the past) and only really makes sense to be used within the pathlib code. I personally think it's better here.

dissect/target/helpers/compat/pathlib.py
Schamper: Yeah... but hopefully this way it's at least easier to fix compatibility issues. Instead of having to try to get the exact same path behavior out of 5 different pathlib implementations.

dissect/target/helpers/compat/pathlib.py
Schamper: Yes, most definitely.

dissect/target/helpers/compat/fnmatch.py
```diff
@@ -0,0 +1,93 @@
+"""The necessary fnmatch internals for our pathlib implementation. Copied from Lib/fnmatch.py.
+
+Update periodically.
+
+Commit hash we're in sync with: add0ca9
```
william-billaud: Maybe use the full commit hash add0ca9ea00ab02fd3a58d059e8370c2d0a1d32c (easier to go on this specific commit, especially from GitHub WebUI) + add the date when the sync was done (like for [build magic](https://github.com/fox-it/dissect.target/blob/660b0102ea26868d5b9b3549181f9765eeeccb01/dissect/target/tools/build_magic.py#L20), it's may be help to track how up to date current implementation is.

And also specify in which branch/major release (here 3.15 & 3.14) this commit correspond to the version used.

dissect/target/helpers/compat/glob.py
```diff
@@ -0,0 +1,331 @@
+"""The necessary glob internals for our pathlib implementation. Copied from Lib/glob.py.
+
+Update periodically.
+
+Commit hash we're in sync with: e542255
```
william-billaud: Same as above : e54225545866d780b12d8e70c33d25fc13b2c3a9 / python version 3.15

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1487 @@
...
+  (most of which derive from polypath rules):
+
+  - All paths are normalized to POSIX-style paths internally
+  - For Windows-style paths, we also consider paths starting with a separator to be absolute, and we have
+    a few more possible "drive letter" names (see polypath.ALLOWED_DRIVE_NAMES)
+
+- All variations of path (i.e. the product of all case sensitivity and separator styles) are pre-created as subclasses,
+  and the correct class will be selected during TargetPath initialization
+
+Commit hash we're in sync with: 1bfe86c
```
william-billaud: 1bfe86caaaaca3ec16351d69fa778f1212f1dd84 (python version : v3.15)

dissect/target/helpers/compat/pathlib.py
Schamper: https://github.com/fox-it/dissect.target/issues/1748

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1492 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterator
+
+    from typing_extensions import Self
+
+    from dissect.target.filesystem import DirEntry, FilesystemEntry
+    from dissect.target.helpers.fsutil import stat_result
+
+
+class UnsupportedOperation(NotImplementedError):
```
william-billaud: I think that this exception must be redefined only if we cannot import the original pathlib.UnsupportedOperation (python < [3.13](https://docs.python.org/3/library/pathlib.html#pathlib.UnsupportedOperation)).
I feel it's a bit missleading having two exception with the same name, but depending of from where this is imported, they won't catch each others :

```python
from pathlib import UnsupportedOperation as PathlibUnsupportedOperation
from dissect.target.helpers.compat.pathlib import UnsupportedOperation

try:
    raise UnsupportedOperation
except PathlibUnsupportedOperation as e:
    print("Catched as PathlibUnsupportedOperation")
except UnsupportedOperation as e:
    print("Catched as UnsupportedOperation")

> Catched as UnsupportedOperation
```

This maybe not a big problem yet as minimum supported version is 3.10, but end of support for 3.12 is in "just" 2.5 year.

Furthermore regarding python backward compatibility policy (pep387), there is nearly never any breaking change in documented exceptions and the semantics which lead to their raising.

tests/helpers/test_fsutil.py
```diff
@@ -922,57 +1735,6 @@ def test_fs_attrs_no_os_listxattr() -> None:
...
-        results = list(target_win.fs.path("/").glob("C:\\windows\\system32\\some*.txt"))
-        assert len(results) == 1
-
-        results = list(target_win.fs.path("/").glob("sysvol/windows/system32/some*.txt"))
-        assert len(results) == 1
-
-
 @pytest.fixture
 def glob_fs() -> VirtualFilesystem:
     vfs = VirtualFilesystem()
```
william-billaud, line 1838: Maybe add a test that ensure that every public method of Path is overrides, it could help identify change in future version. :

```
import inspect
from pathlib import Path
from dissect.target.helpers.compat.pathlib import TargetPath

def check_overrides(parent_class, child_class):
    parent_methods = {
        name for name, _ in inspect.getmembers(parent_class, predicate=inspect.isfunction)
        if (not name.startswith('_')) or name.startswith('__')  # public methods only
    }
    child_methods = {
        name: getattr(child_class, name)
        for name in parent_methods
    }
    not_overridden = [
        name for name, method in child_methods.items()
        if method is getattr(parent_class, name)  # still points to parent's method
    ]
    return not_overridden

# Online redefine one method for tests.
class TestPath(Path):
    def get():
        return None
print(check_overrides(Path, TargetPath))
> []
assert len(check_overrides(Path, TargetPath)) == 0
print(check_overrides(Path, TestPath))
> ['is_relative_to', 'rmdir', 'write_bytes', 'hardlink_to', 'is_char_device', 'relative_to', 'is_dir', 'iterdir', 'full_match', 'replace', '__fspath__', 'write_text', 'is_junction', 'with_stem', 'lchmod', '__eq__', 'samefile', 'is_block_device', 'as_uri', 'is_file', 'expanduser', 'unlink', '__le__', 'with_segments', 'open', 'symlink_to', 'is_symlink', 'rglob', '__ge__', 'is_mount', 'as_posix', 'group', 'lstat', '__lt__', 'glob', 'is_absolute', 'is_fifo', 'mkdir', 'readlink', 'walk', 'match', 'resolve', 'touch', 'joinpath', '__str__', '__truediv__', '__bytes__', 'owner', '__gt__', '__rtruediv__', 'is_reserved', 'with_name', '__init__', 'rename', '__hash__', 'with_suffix', 'is_socket', 'absolute', 'read_bytes', 'read_text', '__reduce__', 'chmod', '__repr__', 'stat', '__new__', 'exists']
```

william-billaud (commented):
Also agree that it's look like a better approach, few minor comment/suggestion.

tests/helpers/test_fsutil.py
Schamper: I updated the docstrings throughout.

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1500 @@
...
+            return NotImplemented
+        return self._str_normcase == other._str_normcase and self.parser is other.parser
+
+    @property
+    def _parts_normcase(self) -> list[str]:
+        # Cached parts with normalized case, for comparisons.
+        try:
+            return self._parts_normcase_cached
+        except AttributeError:
+            self._parts_normcase_cached = self._str_normcase.split(self.parser.sep)
```
Miauwkeru: for windows type paths this part will behave unexpectedly. As `_str_normcase` normalizes it using unix style separators with `/` and the parser.sep here is `\\`.

This will result in `_parts_normcase` to behave like this:
```
 c:/windows/system32   -> ['c:', 'windows', 'system32'] # unix

 c:\\windows\\system32 -> ['c:/windows/system32'] # windows
 c:/windows/system32   -> ['c:/windows/system32'] # windows
```

So wouldn't it be better to split on `/` so the same behaviour is used across the different path types

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1500 @@
...
+        return other == self or other in self.parents
+
+    def is_absolute(self) -> bool:
+        """True if the path is absolute (has both a root and, if applicable, a drive)."""
+        # PATCH: we consider paths absolute if they've been constructed with a leading slash,
+        # or if we've already determined that they have a drive (for Windows-style paths)
+        seps = (self.parser.sep, self.parser.altsep) if self.parser.altsep else (self.parser.sep,)
+        return self.drive != "" or any(path[:1] in seps for path in self._raw_paths)
+
+    def as_uri(self) -> str:
```
Miauwkeru, line 658: I was wondering, if `as_uri` functions, wouldn't it make sense that `from_uri` would also work?

dissect/target/helpers/polypath.py
```diff
@@ -1,80 +1,575 @@
...
+        if part.startswith(seps) or not path:
+            path = part
+        elif path.endswith(seps):
+            path += part
+        else:
+            path += sep + part
+    return path
+
+
+def split(path: str, *, sep: str = "/") -> str:
```
Miauwkeru:
```suggestion
def split(path: str, *, sep: str = "/") -> tuple[str, str]:
```

dissect/target/filesystem.py
```diff
@@ -50,29 +53,46 @@ class Filesystem:
     def __init__(
         self,
         volume: BinaryIO | list[BinaryIO] | None = None,
-        alt_separator: str = "",
+        *,
+        sep: str = "/",
+        altsep: str | None | DEFAULT_ALTSEP = DEFAULT_ALTSEP,
```
Miauwkeru:
```suggestion
        altsep: str | None | type[DEFAULT_ALTSEP] = DEFAULT_ALTSEP,
```
as `DEFAULT_ALTSEP` is not a instance

dissect/target/helpers/compat/pathlib.py
```diff
@@ -0,0 +1,1528 @@
...
+        if not parts:
+            raise ValueError(f"Unacceptable pattern: {str(pattern)!r}")
+        if rel.endswith(sep):
+            # GH-65238: preserve trailing slash in glob patterns.
+            parts.append("")
+        return parts
+
+    def as_posix(self) -> str:
+        """Return the string representation of the path with forward (/) slashes."""
+        return str(self).replace(self.parser.sep, "/")
```
Miauwkeru: is the replace needed since the str is already normalized as a posix path?

dissect/target/helpers/compat/glob.py
```diff
@@ -0,0 +1,335 @@
...
+class _PathGlobber(_GlobberBase):
+    """Provides shell-style pattern matching and globbing for ReadablePath."""
+
+    @staticmethod
+    def lexists(path: Path) -> bool:
+        return path.info.exists(follow_symlinks=False)
+
+    @staticmethod
+    def scandir(path: Path) -> Iterator[tuple[os.DirEntry, str, Path]]:
+        return ((child.info, child.name, child) for child in path.iterdir())
```
Miauwkeru, line 329: technically the `child.info` would be a python 3.14 thing and would not be a thing in `Path` until that time. Is it an idea to update this to use `TargetPath`?

Schamper: I don't really care enough for this to make that change. Let's just leave it.

Miauwkeru (commented):
LGTM, I also checked any issues with acquire and only found that some entries were missing. However, the acquires themselves have the same data. So that was a bit odd.

Are you still planning to add the CHANGELOG.md?

Schamper:
> LGTM, I also checked any issues with acquire and only found that some entries were missing. However, the acquires themselves have the same data. So that was a bit odd.
> 
> Are you still planning to add the CHANGELOG.md?

Can you describe in more detail what was missing?

And yes, I am. Just haven’t found the time yet to formulate something.

Miauwkeru:
> > LGTM, I also checked any issues with acquire and only found that some entries were missing. However, the acquires themselves have the same data. So that was a bit odd.
> > Are you still planning to add the CHANGELOG.md?
> 
> Can you describe in more detail what was missing?

From some looking it seems that a directory entry is absent from the `report.json` file while the directory contents are present

Some modules where I observed this on one of my vms:
```
ActivitiesCache
DPAPI
Powershell
Recents
Startup
```

All of these "missing" entries contain "Documents and Settings" so maybe somethings goes wrong when resolving the path containing a symlink of some sort? 

As an example of an acquire log message:
> [2026-05-13 13:37:05,675] [DEBUG] - Collecting path sysvol/Documents and Settings/Admin/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup
[2026-05-13 13:37:05,678] [INFO ] - Collecting path C:/Users/Admin/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup: Skipped (DEDUP)
[2026-05-13 13:37:05,678] [INFO ] - Collecting symlink branch path C:/Documents and Settings
[2026-05-13 13:37:05,678] [INFO ] - Collecting symlink branch path C:/Documents and Settings: Skipped (DEDUP)
[2026-05-13 13:37:05,678] [DEBUG] - Collecting path sysvol/Documents and Settings/Admin/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup succeeded
[2026-05-13 13:37:05,678] [DEBUG] - Collecting path sysvol/Documents and Settings/All Users/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup
[2026-05-13 13:37:05,683] [ERROR] - Path C:/Documents and Settings/All Users/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup is not found (while collecting sysvol/Documents and Settings/All Users/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup)

Schamper:
@Miauwkeru "Documents and Settings" is a junction/"symlink" to "AppData". Are the following paths correctly collected?

- `C:/Users/Admin/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup`
- `C:/Documents and Settings` (as symlink)

If so, I believe that's path normalization working as intended in acquire. If not, then it's a problem.

Miauwkeru:
> If so, I believe that's path normalization working as intended in acquire. If not, then it's a problem.

I believe the path normalization is working as intended. The additional error message is new but that should be fixed in acquire itself.

I did find a very specific edge case with `target-shell` if you change the case sensitivity:
image

Schamper:
@Miauwkeru this is because of this paragraph in Python's data model (https://docs.python.org/3/reference/datamodel.html#object.__new__)

> If `__new__()` is invoked during object construction and it returns an instance of _cls_, then the new instance’s [`__init__()`](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will be invoked like `__init__(self[, ...])`, where _self_ is the new instance and the remaining arguments are the same as were passed to the object constructor.
> 
> If `__new__()` does not return an instance of _cls_, then the new instance’s [`__init__()`](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will **not** be invoked.

The change in case sensitivity will cause `TargetPath.__new__` to return a different class, so we skip calling `__init__`.

Technically this is fixable by making `with_segments` look like this:
```python
    def with_segments(self, *pathsegments) -> Self:
        obj = type(self).__new__(type(self), self._fs, *pathsegments)
        obj.__init__(self._fs, *pathsegments)
        return obj
````
or
```python
    def with_segments(self, *pathsegments) -> Self:
        obj = type(self)(self._fs, *pathsegments)
        if not isinstance(obj, type(self)):
            obj.__init__(self._fs, *pathsegments)
        return obj
````

However, any operations on the old `TargetPath` object (that still has the "old" parser associated with it) will still use the old case sensitivity, which might cause unexpected behavior. Doing `cd /` in `target-shell` will resolve the problem.

I'm not sure what exactly is the best way forward. I could have a rethink on the pre-generated class approach and go a more dynamic route with the parser. Will cost a little bit in performance though. Or, just put something like this, brand it as an edge case and call it a day.
```python
    def with_segments(self, *pathsegments) -> Self:
        obj = type(self)(self._fs, *pathsegments)
        if not isinstance(obj, type(self)):
            raise TypeError("filesystem variant changed, this TargetPath object is stale")
        return obj
```

Miauwkeru:
> Or, just put something like this, brand it as an edge case and call it a day.

I would say we should just do this and maybe create a ticket to change the case_sensitive with a target-shell command

Schamper:
> create a ticket to change the case_sensitive with a target-shell command

I don't think that's a very common use case. I'm fine leaving it as a "known issue" for now, as I assume that when you do stuff like this, you know enough on how to workaround this.

Miauwkeru (dismissed):
LGTM

Schamper:
@Miauwkeru I rebased to main. With your ack, do we need another ack from twiggler, or is this safe to merge?

Miauwkeru:
> @Miauwkeru I rebased to main. With your ack, do we need another ack from twiggler, or is this safe to merge?

I think it is safe to merge, I already checked it with integration tests and did not see anything weird there

Miauwkeru (approved):

PR #1393, opened 2025-11-10
Fix MSSQL log parser bug

### LINKED ISSUE #1331: MSSQL incorrect timestamp parsing
I have a raw MSSQL ERRORLOG file. Parsing this file using the existing MSSQL plugin results in an issue with the timestamps. The timestamps do not match the message as they do in the raw log.

```bash
$ rg xp_cmdshell fs/C\:/Program\ Files/Microsoft\ SQL\ Server/MSSQL15.MSSQLSERVER/MSSQL/LOG/ERRORLOG
4987:2025-09-15 09:49:56.54 rpid03      SQL Server blocked access to procedure 'sys.xp_cmdshell' of component 'xp_cmdshell' because this component is turned off as part of the security configuration for this server. A system administrator can enable the use of 'xp_cmdshell' by using sp_configure. For more information about enabling 'xp_cmdshell', search for 'xp_cmdshell' in SQL Server Books Online.
4989:2025-09-15 09:50:37.20 rpid03      Configuration option 'xp_cmdshell' changed from 0 to 1. Run the RECONFIGURE statement to install.
4994:2025-09-15 11:26:01.31 rpid03      Configuration option 'xp_cmdshell' changed from 1 to 0. Run the RECONFIGURE statement to install.

$ target-query -f mssql.errorlog TARGET > mssql_errorlog.rec

$ rdump mssql_errorlog.rec ~~f '{ts} -~~> {message}' | grep xp_cmdshell
2025-09-15 09:50:28.880000+00:00 --> SQL Server blocked access to procedure 'sys.xp_cmdshell' of component 'xp_cmdshell' because this component is turned off as part of the security configuration for this server. A system administrator can enable the use of 'xp_cmdshell' by using sp_configure. For more information about enabling 'xp_cmdshell', search for 'xp_cmdshell' in SQL Server Books Online.
2025-09-15 10:15:02.320000+00:00 --> Configuration option 'xp_cmdshell' changed from 0 to 1. Run the RECONFIGURE statement to install.
2025-09-15 11:26:06.290000+00:00 --> Configuration option 'xp_cmdshell' changed from 1 to 0. Run the RECONFIGURE statement to install.
```

qmadev:
closes #1331

### DIFF
--- a/dissect/target/plugins/filesystem/yara.py
+++ b/dissect/target/plugins/filesystem/yara.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import contextlib
 import hashlib
 from io import BytesIO
 from pathlib import Path
 
 from dissect.target.helpers import hashutil
+from dissect.target.helpers.fsutil import open_decompress
 from dissect.target.helpers.logging import get_logger
 
 try:
@@ -54,6 +56,7 @@ class YaraPlugin(Plugin):
     @arg("-p", "--path", default="/", help="path on target(s) to recursively scan")
     @arg("-m", "--max-size", type=int, default=DEFAULT_MAX_SCAN_SIZE, help="maximum file size in bytes to scan")
     @arg("-c", "--check", action="store_true", help="check if every YARA rule is valid")
+    @arg("--no-decompress", action="store_true", help="do not automatically decompress files for extra scanning")
     @export(record=YaraMatchRecord)
     def yara(
         self,
@@ -61,6 +64,7 @@ class YaraPlugin(Plugin):
         path: str = "/",
         max_size: int = DEFAULT_MAX_SCAN_SIZE,
         check: bool = False,
+        no_decompress: bool = False,
     ) -> Iterator[YaraMatchRecord]:
         """Scan files inside the target up to a given maximum size with YARA rule file(s).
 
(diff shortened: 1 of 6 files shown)


dissect/target/plugins/os/windows/log/mssql.py
```diff
@@ -81,19 +81,38 @@ def errorlog(self) -> Iterator[MssqlErrorlogRecord]:
...
+                                _target=self.target,
+                            )
+
+                            buf = ""
+
+                    buf += line
+
+                # For the last line
+                if buf:
+                    if current_ts := RE_TIMESTAMP_PATTERN.match(line):
```
twiggler: This only works if the last line is timestamped. If it is text, it will fail and the last record will be dropped.

The solution is to use `current_ts`

twiggler:
> @twiggler I took a second look at your logic but I could not get it to work. Had issues with messages that had newlines and tabs in it. I added a check to see if a logfile starts with a timestamp and skip it if it does not. Would be very weird for this log format.
> 
> Added some unit tests and changed the test file a little bit to make sure that the last log entry contains a newline which we can test on.
> 
> Lmk what u think.

Hi, can you share (parts of, and/or redacted) the logfile that leads to problems?

qmadev:
Try the test data file

PR #1407, opened 2025-11-24
Improve Citrix Netscaler support

JSCU-CNI:
This PR improves Citrix Netscaler support in Dissect.

* Add support for mounting the kernel root filesystem ("ramdisk") embedded in the kernel
* Improve Netscaler OS detection (filesystem detection, hostname, config parsing, language, install date, shell history)


JSCU-CNI:
> Is there some unit test you could add that at least tests a subset of the added features?

At this time, unfortunately not.

Schamper (changes requested):

PR #1416, opened 2025-11-27
Add fingerprints to ssh plugin records

JSCU-CNI:
This PR adds fingerprint calculation to existing ssh plugin records.


Schamper (changes requested):
Can you fix the tests?

PR #1419, opened 2025-12-02
[for discussion, POC]: make loader arguments discoverable

twiggler:


dissect/target/tools/query.py
```diff
@@ -104,8 +105,19 @@ def main() -> int:
...
+            parser.print_help()
+            return 0
+
+        if (loader_cls := LOADERS_BY_SCHEME.get(args.loader, None)) is None:
+            print(f"Error: Loader '{args.loader}' not found.")
+            return 1
+
+        # The loader now knows how to print its own help.
+        # We just need to instantiate it. Note that the path is not important here.
+        loader_instance = loader_cls(pathlib.Path())
```
twiggler: Passing a dummy path is a bit dubious

dissect/target/loader.py
```diff
@@ -115,6 +130,20 @@ def find_all(path: Path, parsed_path: urllib.parse.ParseResult | None = None) ->
         yield path
 
     def map(self, target: Target) -> None:
+        """Wrapper around the _map function that handles argument passing."""
+        # The loader now parses its own arguments
+        loader_options = self._parser.parse_args(self.loader_args)
```
twiggler: parse_known_args

dissect/target/loader.py
```diff
@@ -66,7 +67,8 @@ class Loader:
     """
 
     def __init__(
-        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True, **kwargs
+        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True,
+        **kwargs
```
Miauwkeru:
```suggestion
        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True, **kwargs
```

dissect/target/loader.py
```diff
@@ -115,6 +136,20 @@ def find_all(path: Path, parsed_path: urllib.parse.ParseResult | None = None) ->
...
+        # The query string can act as a default for any arguments not provided on the command line
+        for key, value in self.parsed_query.items():
+            # argparse sets missing optional arguments to None. We only want to override if it's None.
+            if getattr(loader_options, key, None) is None:
+                setattr(loader_options, key, value)
+
+        # Pass the parsed options directly to the implementation map function
+        return self._map(target, **vars(loader_options))
+
+    def _map(self) -> None:
```
Miauwkeru:
```suggestion
    def _map(self, **kwargs) -> None:
```

dissect/target/loaders/local.py
```diff
@@ -49,21 +56,18 @@ def __init__(self, path: Path, **kwargs):
...
+    def _map(self, target: Target, force_directory_fs: bool = False, fallback_to_directory_fs: bool = False) -> None:
         # For the local loader we abuse the path/URI parsing a bit, so fix it up here
         target.parsed_path = urllib.parse.urlparse(str(target.path))
         target.path_query = self.parsed_query
         target.path = Path("local")
 
         os_name = _get_os_name()
 
-        force_dirfs = "force-directory-fs" in self.parsed_query
-        fallback_to_dirfs = "fallback-to-directory-fs" in self.parsed_query
```
Miauwkeru: this would force this new method of doing these command line arguments where acquire will stop functioning properly as it uses these query parameters. https://github.com/fox-it/acquire/blob/c109fe4f5e642078931fc8c3d80ab7f39e1c496f/acquire/acquire.py#L2433

Miauwkeru (changes requested):

dissect/target/tools/utils/cli.py
```diff
@@ -233,22 +233,20 @@ def open_target(args: argparse.Namespace, *, apply: bool = True) -> Target:
     return target
```
Miauwkeru: Probably a good idea to just remove your changes from this specific file

Miauwkeru (changes requested):

PR #1456, opened 2025-12-15
Add generic Compression Loader

### LINKED ISSUE #1455: Add generic support for compression format in loaders
Currently some loaders implement specific compression formats. For example the Acquire loader supports `.tar.gz`. However, if I have a VMDK which is compressed using gzip this won't work. It would be nice if we could have a generic layer of decompression which handles the gzip (and maybe others) layer, and than let's the other loaders do the heavy lifting. By doing this all the loaders support the gzip format, and we no longer have the need to decompress before using dissect.

Matthijsy:
This PR adds a loader that allows reading gzipped files. It will pack the gzip file into a `VirtualFileSystem`, this way we can access the content without the need to decompress. Than we can use this to pass the new Path (which now no longer has the gzip) to all loaders again. This way you can compress any kind of files currently support by dissect without the need to adapt all loaders for it.  

From a performance perspective this is not the best, it is considerbly slower than processing a de-compressed version. If someone has a better idea than using the VirtualFileSystem which might perform better let me know!

closes #1455

### DIFF
--- a/dissect/target/loaders/tar.py
+++ b/dissect/target/loaders/tar.py
@@ -20,43 +20,9 @@ if TYPE_CHECKING:
 
     from dissect.target import target
 
-
 log = get_logger(__name__)
 
-TAR_EXT_COMP = (
-    ".tar.gz",
-    ".tar.xz",
-    ".tar.bz",
-    ".tar.bz2",
-    ".tar.lzma",
-    ".tar.lz",
-    ".tgz",
-    ".txz",
-    ".tbz",
-    ".tbz2",
-    ".tlz",
-    ".tlzma",
-)
 TAR_EXT = (".tar",)
-
-TAR_MAGIC_COMP = (
-    # gzip
-    b"\x1f\x8b",
-    # bzip2
-    b"\x42\x5a\x68",
-    # xz
-    b"\xfd\x37\x7a\x58\x5a\x00",
-    # lzma
-    b"\x5d\x00\x00\x01\x00",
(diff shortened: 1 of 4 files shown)


Schamper (changes requested):
Could you take some cues from the tar loader on dealing with the compression (and magic), as well as dealing with a few more compression formats besides gzip? Possibly look into `fsutil.open_decompress` too.

If we want to incorporate this nicely, it would be nice to remove any dealing with compression that other loaders currently do themselves, as this loader would be dealing with that.

Matthijsy:
Thank you @Schamper! I reworked the PR to use the MiddelwareLoader. It now supports gz, lzma, bz2 and zst. These are all formats currently accepted by `fsutil.open_decompress`, which we now also use to open the file itself.

dissect/target/loaders/tar.py
```diff
@@ -20,43 +20,9 @@
...
-    b"\x5d\x00\x00\x01\x00",
-    b"\x5d\x00\x00\x10\x00",
-    b"\x5d\x00\x00\x08\x00",
-    b"\x5d\x00\x00\x10\x00",
-    b"\x5d\x00\x00\x20\x00",
-    b"\x5d\x00\x00\x40\x00",
-    b"\x5d\x00\x00\x80\x00",
-    b"\x5d\x00\x00\x00\x01",
-    b"\x5d\x00\x00\x00\x02",
-)
```
Schamper: We seem to have lost a few magics from here, why is that?

PR #1470, opened 2025-12-19
Prevent target info crash

### LINKED ISSUE #1463: target-info should not crash and burn on an error in a single retrieval function
Currently if there's an error in a single piece of information being retrieved (for example, the installation date), `target-info` will crash and burn in its entirety. This should not be acceptable, all the other info can be printed just fine. It should log the error and show "Error" or whatever for the missing piece of information.

william-billaud:
Wrap every function call within target info with a try catch. This allows to ensure that even if one plugin crash for an unknown reason, target info will produces information

On exemple is running `uv run --extra full target-info -r tests/_data/loaders/acquire/test-windows-fs-c-absolute.tar` on the current branch. 
* closes #1463

### DIFF
--- a/dissect/target/tools/info.py
+++ b/dissect/target/tools/info.py
@@ -132,7 +132,8 @@ def get_property(target: Target, func: str) -> str | None:
         if target.has_function(func):
             return getattr(target, func)
     except Exception as e:
-        log.warning("Error executing %s : %s", str(func), e)
+        log.warning("Error executing %s: %s", func, e)
+        log.debug("", exc_info=e)
     return None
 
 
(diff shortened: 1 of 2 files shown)


dissect/target/tools/info.py
```diff
@@ -130,6 +133,14 @@ def get_optional_func(target: Target, func: str) -> str | None:
     return None
 
 
+def get_func_safe(target: Target, func: str) -> str | None:
```
Schamper: Maybe re-use `get_optional_func`? Throw a `try/except` in there, rename it to `get_property`, and use it for the ones you currently wrapped in `get_func_safe`.

dissect/target/tools/info.py
```diff
@@ -111,22 +114,25 @@ def get_target_info(target: Target, recursive: bool = False) -> dict[str, str |
...
 
-def get_optional_func(target: Target, func: str) -> str | None:
-    if target.has_function(func):
-        return getattr(target, func)
+def get_property(target: Target, func: str) -> str | None:
+    try:
+        if target.has_function(func):
+            return getattr(target, func)
+    except Exception as e:
+        log.warning("Error executing %s : %s", str(func), e)
```
Schamper:
```suggestion
        log.warning("Error executing %s: %s", func, e)
        log.debug("", exc_info=e)
```

Schamper (approved):

PR #1473, opened 2025-12-23
Add Surge loader

### LINKED ISSUE #1472: Add Surge loader
Add the Volexity Surge loader to Dissect.

The loader supports Surge image data in:
- zip file
- directory
- subdirectory

Politie-SOC:
Adds the Volexity Surge loader.

The loader supports Surge image data in zip file,
directory and subdirectory.

Unit tests with test data/fixtures are present.

Closes #1472


dissect/target/loaders/surge.py
```diff
@@ -0,0 +1,208 @@
...
+            return path
+
+        try:
+            for subdir in path.iterdir():
+                if subdir.is_dir() and SurgeLoader.contains_all_required_files(path=subdir):
+                    return subdir
+        except Exception:
+            return None
+
+        return None
```
Schamper: Can you move all of these out of the class?

Schamper (changes requested):

Schamper (changes requested):
The map functions are almost identical, can you see if you can deduplicate some of that code? The docstrings are duplicated too but don't reflect the reality of the content of their function though 😄.

The point of splitting the loaders into a directory loader and a zip loader was so that we don't have to parse ZIP files many times - we only parse it once and we can apply different loaders to it. The logic between these loaders in this specific case should still be shared as much as possible. It's very hard to point out the (subtle) differences now and keep changes synchronized.

Politie-SOC:
Updated code as requested by @Schamper. Most of the changes origin from @EarthGrob, so thank you for your effort.

Schamper:
I have some outstanding suggestions for this, but I didn't like the direction it was going so I'm making some more thorough changes (including to the `ZipFilesystem` implementation) to more easily accommodate this loader. Hang tight.

Issue #1477, opened 2025-12-30
On Windows systems t.users returns fewer users/sids vs users found in SAM hive


lhaagsma:
Currently working on parsing groups and groupmembership from the SAM hive. While correlating some information I noticed that `t.users` and UserPlugin.find(sid=sid) returns/can only find users with a home path. As `t.users` is populated from `"HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\ProfileList"`

As a result I cannot associate a username to a SID found in groupmemberships, while these are known to the local system and present in the SAM hive.

My suggestion would be to change https://github.com/fox-it/dissect.target/blob/b805239bc31b488cbbf3da9c9f3a07498e025d8e/dissect/target/plugins/os/windows/_os.py#L306 to yield both (the unique) local users returned by the `sam` (after my PR `sam.users`) plugin as well as the local and remote users found in `ProfileList.`

I might try to include this as part of PR for `sam.groups`.

lhaagsma:
I initially added some code for this, but that PR is currently in draft and turned into different PRs.

A working implementation can be found here that replaces `ProfileList` with `sam.users`:

https://github.com/fox-it/dissect.target/compare/main...lhaagsma:dissect.target:sam-groups#diff-fea9f5c538e75aaaaec1c41d46bb7ecf7d87f43d1d25912c8b1703cff0a33005

PR #1478, opened 2026-01-01
Add sam.groups plugin and reworking of t.users (feedback required)

### LINKED ISSUE #1490: SAM plugin changes

lhaagsma:
This PR changes the `sam` plugin to a namespace of `sam.users` and adds a `sam.groups` function that parses local groups from SAM. 

Some changes are made to `t.users` to also parse users from `sam` as the currently implementation seem to fail to return some local users as mentioned in: https://github.com/fox-it/dissect.target/issues/1477

Feedback requested: 

- Additionally an attempt is made at more user friendly test-case by using exported `.reg` files - including tooling to export registry keys and construction of `VirtualHives` from `.reg` files.  
- I'd like to add `local_admin=Boel` to user records somewhere, either in `SAM.users` or `target.users` as this would provide a user friendly way to identify local admin users. Because the 'administrators' group name can be changed and is language specific.

Still some work to do, such as fix a test i'm failing (`tests/plugins/os/windows/test__os.py::test_windows_user_from_sam`) because the `Mock` target does not get properly parsed by the new `target.users` implementation that makes use of `SAM`. A solution is likely a change to the test, rather than a change in the code as I have tested this scenario on a live machine

### DIFF
--- a/dissect/target/helpers/reg_export.py
+++ b/dissect/target/helpers/reg_export.py
@@ -2,6 +2,7 @@
 """
 reg-export.py - Export Windows registry hives to .reg format using dissect.target
 """
+
 import argparse
 import sys
 import time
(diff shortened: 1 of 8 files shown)


lhaagsma:
Sun-setting this PR (for now in Draft) and creating new ones for:

- adding write logic to `RegFlex` (the read part of reg_export already existed)
- parsing `sam.groups` (and namespace change from `sam` to `sam.users`
- changing `target.users` to use `sam.users` -> though this might just be an issue for now as it would depend on namespace change part of `sam.groups`

PR #1480, opened 2026-01-05
Skip nested objects of Velociraptor records

respondersGY:
The objects contain additional metadata that is not necessary to output as records. In addition, some of the dynamic (nested) objects result in conflicts with other plugins, which can't be determined beforehand depending on which artefacts are collected.


respondersGY:
@Schamper can you review the PR?

tests/plugins/apps/edr/test_velociraptor.py
```diff
@@ -35,8 +35,10 @@ def test_windows_velociraptor(mock_velociraptor_dir: Path, target_win: Target) -
     target_win.add_plugin(VelociraptorPlugin)
 
     results = list(target_win.velociraptor())
+    results_extract = list(target_win.velociraptor.results(extract=True))
```
Schamper:
```suggestion
    results_extract = list(target_win.velociraptor.results(extract_nested=True))
```

respondersGY:
@Schamper  Can you **please** review the changes?

PR #1482, opened 2026-01-07
Make the cache compatible with iterators

twiggler:
When the cache is enabled, calling `next` on a supposed iterator from a plugin function causes an error, because the cache is iterable but not an iterator.

### DIFF
--- a/tests/helpers/test_cache.py
+++ b/tests/helpers/test_cache.py
@@ -64,8 +64,8 @@ def test_cache_filename(target_win: Target) -> None:
 
 
 def test_cache_write_failure_behavior(target_bare: Target, tmp_path: Path) -> None:
-    """
-    Specifically tests the 'Write Path' (Cache Miss) which returns a CacheWriter.
+    """Specifically tests the 'Write Path' (Cache Miss) which returns a CacheWriter.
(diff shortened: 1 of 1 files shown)


tests/helpers/test_cache.py
```diff
@@ -56,3 +61,65 @@ def test_cache_filename(target_win: Target) -> None:
     assert (
         cache4.cache_path(target_win, ()).stem == "dissect.target.plugins.os.windows.ual.UalPlugin.client_access.KCk="
     )
+
+
+def test_cache_write_failure_behavior(target_bare: Target, tmp_path: Path) -> None:
+    """
+    Specifically tests the 'Write Path' (Cache Miss) which returns a CacheWriter.
+    We verify that CacheWriter acts as an Iterator even when the underlying
```
Schamper:
```suggestion
    """Specifically tests the 'Write Path' (Cache Miss) which returns a CacheWriter.

    We verify that CacheWriter acts as an Iterator even when the underlying
```

Schamper: Nitpicking

Schamper (dismissed):

PR #1483, opened 2026-01-08
Add NTDS filesystem

Schamper:
Depends on https://github.com/fox-it/dissect.database/pull/8

Allows you to browse NTDS.dit files like a filesystem:
image


joost-j (dismissed):
LGTM; existing code in shell.py is mostly rewritten to use `with` clauses and the rest consists of new code. The new filesystem is indeed a really nice flex and a great cli-based NTDS.dit browser!

PR #1485, opened 2026-01-08
Improve target-shell autocompletion for dirs with spaces

JSCU-CNI:
This PR attempts to solve a long-standing annoyance with target-shell, the handling of directories with spaces in the shell autocomplete function. We attempt to solve this by escaping spaces with "\ ", as most real shells do too. This seems to work reasonably well.


Schamper (changes requested):
Is it possible to extend the existing autocompletion tests with this?

Schamper (changes requested):
Can you add a test (or at least a comment) with what exactly still breaks as discussed in https://github.com/fox-it/dissect.target/issues/1621#issuecomment-4054555379?

JSCU-CNI:
I think this is now fixed with https://github.com/fox-it/dissect.target/issues/1720

PR #1415, opened 2025-11-27
Add certificate parsing to webserver plugins

JSCU-CNI:
This PR adds basic x509 certificate parsing to the webserver plugins of apache and nginx.

PR #1498, opened 2026-01-14
Normalize plugins Certificates output

### LINKED ISSUE #1452: Normalize Windows CertLog Certificates output
PR #1415 introduced a helper function to parse X509 certificates. The function `certlog.certificates` introduced in #1443 performs it's own rudimentary certificate parsing. Ideally the output of this function should contain normalized fields (e.g. `serial_number` as `varint`, `issuer_dn`, etc.).

william-billaud:
Following #1415

Normalize serial_number output to string format (hex), which is the common way serial number are shown (e.g by browser or openssl).
Rename some field for consistency between plugins that manipu.

 * fixes #1452
 
cc  @JSCU-CNI

### DIFF
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -278,6 +278,7 @@ register("overlay2", "Overlay2Loader")
 register("overlay", "OverlayLoader")
 register("phobos", "PhobosLoader")
 register("uac", "UacLoader")
+register("vmsupport", "VmSupportLoader")
 # tar and zip loaders should be low priority to give other loaders a chance first
 register("tar", "TarLoader")
 register("zip", "ZipLoader")
--- a/dissect/target/loaders/nscollector.py
+++ b/dissect/target/loaders/nscollector.py
@@ -31,7 +31,7 @@ class NsCollectorTarSubLoader(TarSubLoader):
         return required_paths.issubset(names)
 
     def map(self, target: Target) -> None:
-        fs = TarFilesystem(tarfile=self.tar, base=self.tar.getnames()[0], fh=self.tar.fileobj)
+        fs = TarFilesystem(self.tar.fileobj, base=self.tar.getnames()[0])
         target.filesystems.add(fs)
 
         # Symlink /nsconfig to /flash/nsconfig to make Citrix parsers compatible
--- a/dissect/target/loaders/tar.py
+++ b/dissect/target/loaders/tar.py
@@ -137,6 +137,7 @@ class TarLoader(Loader):
         import_lazy("dissect.target.loaders.acquire").AcquireTarSubLoader,
         import_lazy("dissect.target.loaders.uac").UacTarSubloader,
         import_lazy("dissect.target.loaders.nscollector").NsCollectorTarSubLoader,
+        import_lazy("dissect.target.loaders.vmsupport").VmSupportTarSubloader,
         GenericTarSubLoader,  # should be last
     )
 
--- a/dissect/target/loaders/vmsupport.py
+++ b/dissect/target/loaders/vmsupport.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from dissect.target.filesystems.tar import TarFilesystem
+from dissect.target.loader import Loader
+from dissect.target.loaders.dir import map_dirs
+from dissect.target.loaders.tar import TarSubLoader
+from dissect.target.plugin import OperatingSystem
+
+if TYPE_CHECKING:
+    import tarfile as tf
+    from pathlib import Path
+
+    from dissect.target.target import Target
+
+
+# From tech support readme
+# Files/Directories of Interest:
+# ------------------------------
+#
+# error.log: log containing errors that ocurred while running vm-support
+# action.log: a log of all commands, and/or actions run during vm-support
+# commands: a directory containing output files for all commands run
+#
+# All other directories and files should be mirrors of the the ESXi system
+# vm-support was run on.
+
+EXPECTED_FILES_OR_DIR = ["etc/vmware/esx.conf", "error.log", "action.log"]
+
+
+class VmSupportLoader(Loader):
+    """Loader for extracted ESXi vm-support.
+
+    References:
+        - https://knowledge.broadcom.com/external/article/313542
+    """
+
+    @staticmethod
+    def detect(path: Path) -> bool:
+        if not path.is_dir():
+            return False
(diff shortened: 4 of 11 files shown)


dissect/target/plugins/os/windows/certlog.py
```diff
@@ -172,13 +172,14 @@
     "$AttributeValue": "common_name",
     "$CRLPublishError": "crl_publish_error",
     "$CallerName": "caller_name",
-    "$CertificateHash2": "certificate_hash2",
+    "$CertificateHash": "fingerprint",
+    "$CertificateHash2": "fingerprint",
```
Schamper: Won't this override each other?

Schamper (changes requested):

JSCU-CNI:
Thanks for the PR, this is a good step in normalizing certificate fields across the project.

> Normalize serial_number output to string format (hex), which is the common way serial number are shown (e.g by browser or openssl).

Unfortunately serial numbers can also be negative (signed) integers (4.1.2.2, https://www.ietf.org/rfc/rfc2459.txt, `-set_serial n` in OpenSSL), which is difficult to properly communicate in hex notation (`-1337` -> `FA C7`).

We prefer the notation of serial numbers as integers, as most cryptography libraries do by default as well (BoringSSL, asn1crypto, pyca/cryptography). 

Perhaps we can add two fields to accommodate both use cases, `serial_number` and `serial_hex` or `serial_number_hex`.

Schamper:
@william-billaud I agree with @JSCU-CNI here. Even though it's only a data representation problem of the same data, I can see the value of having a readily available/searchable hex variant too.

Maybe opt for `serial_number_hex`?

> which is difficult to properly communicate in hex notation (-1337 -> FA C7).

Even more so because Python by default will make that `-0x539`.

dissect/target/plugins/os/windows/certlog.py
william-billaud: MS docs only mention $CertificateHash, but on my sample (generate and real cases), I only ever had $CertificateHash2. . it was an error during the initial plugin release. 
My guess is that CertificateHash2 support sha256 (but it was sha1 on my sample...) and that it replaces CertificateHash. I have added warning if this would cause an override. 

I have found some reference online of $CertificateHash : 

* Ms docs : https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-csra/5f06c74c-1a29-4fdf-b8dd-ae3300d1b90d
* Article related to certlog from 2012 : https://sysengblog.wordpress.com/2012/04/03/complete-microsoft-certificate-authority-maintenance-procedure/

dissect/target/plugins/os/windows/certlog.py
```diff
@@ -283,9 +285,15 @@ def format_serial_number(serial_number_as_hex: str | None, target: Target) -> st
     return serial_number_as_hex.replace(" ", "")
 
 
+def serial_number_as_int(serial_number_as_hex: str | None) -> int | None:
+    if not serial_number_as_hex:
+        return None
+    return int(serial_number_as_hex, 16)
```
JSCU-CNI: I fear that this does not correctly convert negative hex encoded serial numbers as they would need `byteorder="big" and signed=True`.

dissect/target/plugins/os/windows/certlog.py
```diff
@@ -57,7 +57,9 @@
         ("string", "organization"),
         ("string", "organizational_unit"),
         ("string", "public_key_algorithm"),
+        ("string", "serial_number"),
         ("string", "serial_number_hex"),
+        ("varint", "serial_number"),
```
Schamper: Double serial_number?

tests/plugins/apps/webserver/test_apache.py
```diff
@@ -501,8 +501,96 @@ def test_apache_hosts_certificates(target_unix: Target, fs_unix: VirtualFilesyst
     assert records[0].fingerprint.sha1 == "6566d8ebea1feb4eb3d12d9486cddb69e4e9e827"
     assert records[0].fingerprint.sha256 == "7221d881743505f13b7bfe854bdf800d7f0cd22d34307ed7157808a295299471"
     assert records[0].serial_number == 21067204948278457910649605551283467908287726794
+    assert records[0].serial_number_hex == "0x03b0afa702c33e37fffd40e0c402b2120c1284ca"
```
Schamper: I don't think the `0x` prefix is necessary? At least Safari doesn't show it.

Schamper: I agree, the browser one seems more consistent and easier to understand.

dissect/target/helpers/certificate.py
```diff
@@ -75,6 +75,32 @@ def compute_pem_fingerprints(pem: str | bytes) -> tuple[str, str, str]:
...
 
+def format_serial_number_as_hex(serial_number: int | None) -> str | None:
+    """Format serial_number from integer to hex.
+
+    Add a prefix 0 if output length is not pair, in order to be consistent with usual serial_number representation
+    (navigator, openssl etc...).
+    For negative number use the same representation as navigator, which differ from openssl
+    E.g for -1337 :
+        open ssl : Serial Number: -1337 (-0x539)
+        Navigator : FA C7
```
Schamper:
```suggestion
    Add a prefix 0 if output length is not pair, in order to be consistent with usual serial_number representation
    (navigator, openssl etc...). 
    For negative number use the same representation as navigator, which differ from OpenSSL.
    
    For example for -1337::
    
        OpenSSL : Serial Number: -1337 (-0x539)
        Navigator : FA C7
```

dissect/target/helpers/certificate.py
```diff
@@ -75,6 +75,32 @@ def compute_pem_fingerprints(pem: str | bytes) -> tuple[str, str, str]:
...
+    if serial_number is None:
+        return serial_number
+
+    if serial_number > 0:
+        serial_number_as_hex = f"{serial_number:x}"
+        if len(serial_number_as_hex) % 2 == 1:
+            serial_number_as_hex = f"0{serial_number_as_hex}"
+        return f"{serial_number_as_hex}"
+    # Representation is always a multiple of 8 bits, we need to compute this size
+    output_bin_len = (8 - (len(bin(serial_number)) - 3) % 8) + len(bin(serial_number)) - 3
```
Schamper:
```suggestion
    output_bin_len = (8 - (serial_number.bit_length() % 8) + serial_number.bit_length()
```
Untested but this should give the same result.

Schamper (approved):

PR #1499, opened 2026-01-15
Fix ESXi configstore call in OS create method

JSCU-CNI:
Since `_configstore` is no longer populated in the `ESXiPlugin`, we have to update the reference to the config store in the `create` method.

The `check_compatible` method of `ConfigStorePlugin` is also updated to not use `target.os` since that would cause recursion.

### DIFF
--- a/tests/plugins/os/unix/esxi/test__os.py
+++ b/tests/plugins/os/unix/esxi/test__os.py
@@ -90,7 +90,10 @@ def test_esxi_os_creation_version_7(target_bare: Target) -> None:
 
     fs2 = VirtualFilesystem()
     fs2.map_file_fh("/etc/vmware/esx.conf", BytesIO(b'/resourceGroups/version = "7.13.37"\n'))
-    fs2.map_file("/var/lib/vmware/configstore/backup/current-store-1", absolute_path("_data/plugins/os/unix/esxi/current-store-1"))
+    fs2.map_file(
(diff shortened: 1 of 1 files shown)


JSCU-CNI:
@william-billaud That would definitely be helpful for the test coverage of ESXi functionality. Perhaps we can add that in a separate PR.

@Schamper I've added a small test in [c8a2a8e](https://github.com/fox-it/dissect.target/pull/1499/commits/c8a2a8e7b50c04296e12c77949d4d1e6329cf4b5) for the meantime.

tests/plugins/os/unix/esxi/test__os.py
```diff
@@ -73,3 +75,23 @@ def test_esxi_os_detection(target_bare: Target, fs_esxi: VirtualFilesystem) -> N
     assert target_bare.hostname == "localhost"
     assert target_bare.version == "6.7.0"
     assert target_bare.ips == ["192.168.56.101"]
+
+
+def test_esxi_os_creation_version_7(target_bare: Target) -> None:
+    """Test handling of ``ESXiPlugin.create`` for ESXi 7 with separate partitions.
+    Indirectly tests the ESXi configstore plugin."""
```
Schamper:
```suggestion

    Indirectly tests the ESXi configstore plugin.
    """
```

Schamper (changes requested):

PR #1504, opened 2026-01-18
Fix crash in `linux.environ` plugin when `environ` file is missing

B0TAxy:
This MR fixes a bug in the `linux.environ` plugin where attempting to parse the environment of a process would crash if the `/proc/<pid>/environ` file does not exist (e.g., kernel threads or race conditions during process termination).

**Changes:**

* **`dissect/target/plugins/os/unix/linux/proc.py`**: Added an explicit `.exists()` check in `_parse_environ` before attempting to read the file. It now gracefully returns/skips instead of raising an exception.
* **`tests/conftest.py`**: Updated the `fs_linux_proc` fixture to simulate a scenario where a process in `/proc` lacks an `environ` file (conditional mapping).
* **`tests/.../test_environ.py`**:
* Updated the expected result count from `4` to `3` (since one process is now correctly skipped due to the missing file).
* Added a linter suppression comment for the fixture usage.

**Testing:**

* Updated existing tests to verify that the plugin no longer crashes on missing files and correctly parses the remaining valid processes.

### DIFF
--- a/tests/plugins/os/unix/linux/test_environ.py
+++ b/tests/plugins/os/unix/linux/test_environ.py
@@ -10,9 +10,6 @@ if TYPE_CHECKING:
 
 
 def test_environ(target_linux_users: Target, fs_linux_proc: VirtualFilesystem) -> None:
-    # Assigned to `_` to suppress "unused variable" linter warnings.
-    _ = fs_linux_proc
-
     target_linux_users.add_plugin(ProcPlugin)
     results = list(target_linux_users.environ())
     assert len(results) == 3


tests/plugins/os/unix/linux/test_environ.py
```diff
@@ -10,6 +10,9 @@
 
 
 def test_environ(target_linux_users: Target, fs_linux_proc: VirtualFilesystem) -> None:
+    # Assigned to `_` to suppress "unused variable" linter warnings.
+    _ = fs_linux_proc
+
```
Schamper:
```suggestion
```

PR #1517, opened 2026-01-26
Add individual hash command to target-shell

### LINKED ISSUE #1516: Add md5sum, sha1sum and sha256sum to target-shell
It would be nice to have the commands `md5sum`, `sha1sum` and `sha256sum` in `target-shell`. Now, hashes can be retrieved by running the custom `hash`, which is nice, but not obvious.

skepppy:
This PR adds the commands `md5sum`, `sha1sum` and `sha256sum` to `target-shell`.

Closes #1516.


Schamper (changes requested):
Same for the rest.

The docstrings in this file as specifically formatted like this to look the most "CLI-like" when used as help messages. Perhaps there should be a comment explaining that in this file.

PR #1531, opened 2026-02-06
Add support for VDI changes

Schamper:
Add support for https://github.com/fox-it/dissect.hypervisor/pull/72

### DIFF
--- a/tests/plugins/filesystem/test_walkfs.py
+++ b/tests/plugins/filesystem/test_walkfs.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import itertools
+import platform
 import stat
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dissect/target/containers/vdi.py
```diff
@@ -17,11 +17,9 @@ class VdiContainer(Container):
     __type__ = "vdi"
 
     def __init__(self, fh: BinaryIO | Path, *args, **kwargs):
-        f = fh
-        if not hasattr(fh, "read"):
-            f = fh.open("rb")
-        self.vdi = vdi.VDI(f)
+        self.vdi = vdi.VDI(fh)
 
+        self._stream = self.vdi.open()
```
twiggler: The API suggests a symmetry between vdi.open() and vdi.close() that doesn't actually exist. In reality, the code operates on different levels of abstraction.

The VDI acts as a manager that you shut down using vdi.close() (Level 1).
However, it also acts as a factory to generate results (streams) via vdi.open() (Level 2).

One option would be to rename vdi.open() to vdi.get_stream() or vdi.create_stream().

twiggler (approved):

Issue #1487, opened 2026-01-12
Standardize plugin input files


Matthijsy:
Most plugins use some kind of input file, parse them and then output it as records. It would be nice if we can have a standard method on every plugin that exports a list/object which defined all the input files. This way we can ask every plugin the input file it depends on, and for example also create a `target-extract` which dumps the raw files instead of the records. This would allow to obtain raw artifacts more easily.

If you combine this with #1471 every Plugin effectively just becomes a list of input files which are passed into the parser.

Would this be something more people would be interessted in?

Issue #1471, opened 2025-12-19
Dissecting plugins: cutting out the parser


twiggler:
It is surprisingly difficult to run a plugin function on a single file on disk.

This is our current architecture:

<img width="818" height="290" alt="Image" src="https://github.com/user-attachments/assets/47324a06-aae1-404c-ad5f-ed5fd9badcd0" />

**Now let's cut out the "parse" part into a parser:**

<img width="963" height="422" alt="Image" src="https://github.com/user-attachments/assets/b85967e8-103c-4c94-b405-20172a05b7de" />

What are the advantages:

- Separates parsing, which is a crucial part of our core domain, from secondary and tertiary concerns.
- Provides a path  to run parsers on single files without instantiating the jungle, perhaps by creating a CLI tool named `run-parser`
- Enables the use of dissect parsers in other tools and libraries.
- Gradual migration: plugins are free to embrace the concept.

Some considerations:

- Some plugins already moved the parsing part to parser classes.
- Ideally, parsers are pure: they only depend on (a set of) input streams.
- They will not directly depend on the target API for auxiliary information.
- Plugins mediate between the target API and the separate parsers.

_Todo: for exposition, it is a good idea to add an example parser_

Let's discusss!

PR #1532, opened 2026-02-06
POC: Implement artifact spec

Matthijsy:
This PR adds a POC for defining needed artifacts as artifact spec (with a bit of inspiration from acquire). Currently multiple plugins have a repetative pattern of loading related artifacts which can be handled by `Plugin` after this PR. Furtheremore this makes it possible to have acquire collect everything that is parsed by target, without the need to specify the spec on both sides. 

This PR adds a `__spec__` attribute to each plugin. This property is then used to collect all the needed artifacts from the target, and can be obtained by the plugins by using the new `artifacts` property. This way it is no longer needed to obtain the relevant files within the `__init__`.  

Currently there is already `Plugin.get_paths()` which is aimed to return all the relevant paths for a certain plugin. However, if we would use this in acquire we loose the option to log that we aimed to collect a certain file but it's not present. Acquire simply does not have the information about which paths are checked. After this PR we can use `__spec__` as spec definition within Acquire.

For now I only implemented it for the Anydesk plugin, mainly since I wanted to know if this is a way in which we want to move forward. Tests, and further implementation can follow but since this touches the base of Plugin it might be good to have some discussion first. 

Relates to #1082 and #1487

### DIFF
--- a/dissect/target/artifact_spec.py
+++ b/dissect/target/artifact_spec.py
@@ -1,4 +1,8 @@
 from collections.abc import Iterator
+from pathlib import Path
+
+from dissect.target import Target
+from dissect.target.plugins.general.users import UserRecord
 
 
 class Artifact:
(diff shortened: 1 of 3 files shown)


Schamper:
Could be relevant for https://github.com/fox-it/dissect.target/issues/1471.

PR #1533, opened 2026-02-06
Add file type to walkfs

Matthijsy:
This PR adds a field called `name` and `type` to the walkfs records.


Schamper:
Can you explain what the added benefit is of adding `name`? It's also accessible as `path.name`.

Also keep in mind that `is_dir` will return `True` if it's a symlink pointing to a directory. Not sure if that's what you intended. Also please add unit tests.

Matthijsy:
I am trying to extract all the records into a database via json files. I would like to have a field with the name (and not the full path) in there. What is the normal workflow for that? I just figured out that doing ` rdump -E 'import stat; name = path.name; t = "dir" if stat.S_ISDIR(mode) else "file"'` gives the same result as this PR. Would that be a better way to achieve this result?

dissect/target/plugins/filesystem/walkfs.py
```diff
@@ -118,17 +120,25 @@ def generate_record(target: Target, entry: FilesystemEntry, capability: bool) ->
     else:
         fs_types = [entry.fs.__type__]
 
+    ftype = "unknown"
+    if entry.is_dir():
+        ftype = "dir"
+    elif entry.is_file():
+        ftype = "file"
```
yunzheng: Can you also add a check for `is_symlink()`?

tests/plugins/filesystem/test_walkfs.py
```diff
@@ -47,6 +47,24 @@ def test_walkfs_plugin(target_unix: Target, fs_unix: VirtualFilesystem) -> None:
...
+        "to",
+        "some",
+        "file",
+        "other",
+        "file.ext",
+        "root_file",
+        "var",
+    ]
+    assert results[0].type == "dir"
+    assert results[3].type == "file"
```
yunzheng: also add a test for symlink

Schamper (changes requested):

PR #1552, opened 2026-02-16
Automatically YARA scan compressed files by decompressing them

### LINKED ISSUE #1542: Add option for YARA plugin to scan compressed files
Would be nice if during the iteration of files, the YARA plugin automatically scans the contents of compressed files. Perhaps `open_decompress` can be used here.

yunzheng:
This PR ensures that the contents of compressed files are also scanned in it's decompressed form.

By default it will scan both the original and decompressed form. If `--no-decompress` is specified, only the original file is scanned. 

resolves #1542

### DIFF
--- a/dissect/target/plugins/filesystem/yara.py
+++ b/dissect/target/plugins/filesystem/yara.py
@@ -95,7 +95,10 @@ class YaraPlugin(Plugin):
         if no_decompress:
             self.target.log.info("Will not automatically check in compressed files")
         else:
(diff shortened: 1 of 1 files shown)


yunzheng:
> > I was thinking we can limit the read to fh.read(max_size) so it stays within the given limits.
> 
> That might be a good idea, yes.

done. but it won't throw a warning for decompressed files if the size is bigger like it would for a `.stat().st_size` check on normal files.

Schamper (approved):

PR #1559, opened 2026-02-19
Add NTDS Group Policy records

Matthijsy:
This PR adds a method that exports all Group Policies which can be fetched from the NTDS.

We might have to consider how this relates to [adpolicy](https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/windows/adpolicy.py) plugin. They are kinda related, but depend on completly different artifacts. 

Depends on https://github.com/fox-it/dissect.database/pull/38
Depends on: https://github.com/fox-it/dissect.database/pull/41

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -269,7 +269,7 @@ class Filesystem:
         """
         return self.get(path).walk_ext(topdown, onerror, followlinks)
 
-    def recurse(self, path: str) -> Iterator[DirEntry]:
+    def recurse(self, path: str) -> Iterator[FilesystemEntry]:
         """Recursively walk a directory and yield contents as :class:`FilesystemEntry`.
 
         Does not follow symbolic links.
@@ -633,7 +633,7 @@ class FilesystemEntry:
         """
         yield from fsutil.walk_ext(self, topdown, onerror, followlinks)
 
-    def recurse(self) -> Iterator[DirEntry]:
+    def recurse(self) -> Iterator[FilesystemEntry]:
         """Recursively walk a directory and yield its contents as :class:`DirEntry`.
 
         Does not follow symbolic links.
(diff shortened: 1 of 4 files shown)


dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -92,6 +105,7 @@ def __init__(self, target: Target):
         super().__init__(target)
         self.path = None
 
+        # Fallback path
```
Schamper: I don't really understand this comment?

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -138,6 +152,22 @@ def computers(self) -> Iterator[NtdsComputerRecord]:
...
             )
 
+    @export(record=NtdsGPORecord)
+    def group_policy(self) -> Iterator[NtdsGPORecord]:
+        """Extract all group policy objects (GPO) NTDS.dit database."""
+
+        for gpo in self.ntds.group_policies():
+            yield NtdsGPORecord(
+                cn=gpo.cn,
+                distinguished_name=gpo.distinguishedName,
```
Schamper: Could you use either the "typed"/"defined" properties for these attributes on the objects' class, or the `.get` method for the attributes that don't yet have a property defined on their Python classes?

Schamper: Some of these exist in the base `Object` class already:

https://github.com/fox-it/dissect.database/blob/357f64263d1733330bb250fb50bdd019ad438e52/dissect/database/ese/ntds/objects/object.py#L198

The intention is to have most of the attributes added as Python properties over time in their classes, but for some it's fine to just use `.get()` for now. If you're willing to add the proper attributes to the proper classes, that's of course appreciated!

My point is more, while "magic attributes" exist on these objects (e.g. doing `.whenCreated` works), I'd prefer to use properly typed/available properties and methods in library code, such as `.when_created` or `.get("whenCreated")`. The magic attributes are more nice for interactive use.

PR #1214, opened 2025-07-01
Add ShellFilesystem and friends

Schamper:
Initial PR, still needs some work, mainly docstrings and unit tests.

Closes https://github.com/fox-it/dissect.target/issues/1295.

PR #1589, opened 2026-02-26
Add middleware loader

Schamper:
Consider this a proposal/experiment for a middleware loader, which should facilitate making loaders like https://github.com/fox-it/dissect.target/pull/1456 and #1214 easier. I've adjusted the VBK loader as test (which also immediately serves as a unit test).

@qmadev I remember you looked a bit at VBK too, besides the middleware loader itself, does the VBK loader change make sense to you?

### DIFF
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -152,6 +152,42 @@ class SubLoader(Loader, Generic[T]):
         raise NotImplementedError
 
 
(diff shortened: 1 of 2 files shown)


qmadev:
Other than that, it looks good to me. Seems like this would make sure you never have to look for a new loader within a loader anymore.

PR #1592, opened 2026-03-01
BugFix/small fixes

loaflover:
just some minor fixes, with varying importance.

openssh plugin:
    removed unused variable.
    improved ``find_sshd_directory`` function, removing a redundent check and allowing it to return None.

opensshd plugin:
    updated to work with new ``find_sshd_directory`` function.
    
iis plugin:
    moved variable assignment inside the If block it is used in. very slightly better for performance.

clsid plugin:
    changed key paths from dict to standalone consts. 
    fixed ``create_records`` function docstring
    fixed plugin functions and ``create_records`` function to support the new CONST types.

firewall plugin:
    added missing commas to a list. 

ual plugin:
    changed the way the log and db dir were stored.
    
test_opensshd:
    made work with the new ``find_sshd_directory`` function

### DIFF
--- a/dissect/target/plugins/os/windows/ual.py
+++ b/dissect/target/plugins/os/windows/ual.py
@@ -178,7 +178,7 @@ class UalPlugin(Plugin):
         return [
             path
             for path in self.target.resolve(self.LOG_DIR).glob(self.LOG_DB_GLOB)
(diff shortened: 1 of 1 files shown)


loaflover:
i fixed all failed tests, i tried to do it in the least invasive way.

hope its good (:

PR #1599, opened 2026-03-03
NTDS: Enhanced records

B0TAxy:
This PR expands the parser's coverage by adding support for several new record types, ensuring a more comprehensive map of the target environment.  
Also, added better handling in case the pek cannot be unlocked


Schamper (changes requested):

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -93,6 +170,18 @@
...
+MS_DS_MANAGED_PASSWORD_DEF = """
+struct gms_managed_password {
+    uint16 version;
+    uint16 reserved;
+    uint32 length;
+    char password[length];
+};
+"""
+
+MS_DS_MANAGED_PASSWORD = cstruct().load(MS_DS_MANAGED_PASSWORD_DEF)
```
Schamper: I'd prefer this field (`msDS-ManagedPassword`) to be an "enhanced field" in `dissect.database` directly. So that you don't have to parse this data yourself as a "user" of the NTDS parser. Could you make a separate PR for that in `dissect.database`?

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -213,60 +344,203 @@ def secretsdump(self) -> Iterator[str]:
...
+
+
+def get_supplemental_credentials(account: User | Computer) -> Iterator[str]:
+    """Extract supplemental credentials from a user object.
+
+    Args:
+        account (User | Computer): The user or computer object to extract supplemental credentials from.
+
+    Yields:
+        supplemental_credential (str): Supplemental credential in the secrets dump format
```
Schamper:
```suggestion
        account: The user or computer object to extract supplemental credentials from.

    Yields:
        Supplemental credentials in the secrets dump format.
```
Typing is already provided from the function signature.

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -213,60 +344,203 @@ def secretsdump(self) -> Iterator[str]:
...
+        "telephone_number": account.get("telephoneNumber"),
+        "home_directory": account.get("homeDirectory"),
+        "logon_script": account.get("scriptPath"),
+        "service_principal_names": account.get("servicePrincipalName"),
+    }
+
+
+def extract_laps(computer: Computer) -> dict[str, Any]:
+    """Extract LAPS information from the NTDS.dit database.
+    LAPS information can be stored in multiple attributes depending on the Windows version and LAPS configuration.
```
Schamper, line 490:
```suggestion

    LAPS information can be stored in multiple attributes depending on the Windows version and LAPS configuration.
```

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -213,60 +344,203 @@ def secretsdump(self) -> Iterator[str]:
...
+
+    Args:
+        computer (Computer): The computer object to extract LAPS information from.
+
+    Returns:
+        laps (dict[str, Any]): Dictionary containing LAPS information
```
Schamper:
```suggestion
        computer: The computer object to extract LAPS information from.

    Returns:
        Dictionary containing LAPS information.
```

Schamper (changes requested):

dissect/target/plugins/os/windows/ad/ntds.py
B0TAxy: Sure, but I will say we currently don't have a data set that contains this field, so it will be hard to test if it works.

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -213,60 +330,204 @@ def secretsdump(self) -> Iterator[str]:
...
+    except Exception:
+        sfu_password = None
+
+    try:
+        user_password = account.get("userPassword")
+    except Exception:
+        user_password = None
+
+    try:
+        unix_password = account.get("unixPassword")
```
Schamper, line 450: I can't find any evidence that this field exists in active directory. Please provide a source for your information or manually verify accuracy of your code and remove hallucinations before submitting a PR for review.

dissect/target/plugins/os/windows/ad/ntds.py
```diff
@@ -213,60 +330,204 @@ def secretsdump(self) -> Iterator[str]:
...
+        "unix_password": unix_password if unix_password else None,
+        "sfu_password": sfu_password,
+        "supplemental_credentials": list(get_supplemental_credentials(account)),
+        "user_account_control": account.user_account_control.name.split("|"),
+        "primary_group_id": account.primary_group_id,
         "member_of": member_of,
-        "service_principal_name": user.get("servicePrincipalName"),
+        "allowed_to_delegate": account.get("msDS-AllowedToDelegateTo"),
+        "info": account.get("info"),
+        "comment": account.get("comment"),
```
Schamper, line 477: I can't find evidence that this field exists on user objects. Please provide a source for your information or manually verify accuracy of your code and remove hallucinations before submitting a PR for review.

Schamper (changes requested):

dissect/target/plugins/os/windows/ad/ntds.py
B0TAxy: Hi @Schamper,

Thank you for catching that! You are completely right—I made a typo in the field name.

The field does exist in the Active Directory schema, but it should be `unixUserPassword` instead of `unixPassword`. You can find the Microsoft documentation confirming the schema attribute here:
[https://learn.microsoft.com/en-us/windows/win32/adschema/a-unixuserpassword](https://learn.microsoft.com/en-us/windows/win32/adschema/a-unixuserpassword)

I will push a fix shortly to correct `unixPassword` to `unixUserPassword`. Thanks again for the review!

dissect/target/plugins/os/windows/ad/ntds.py
B0TAxy: Hi @Schamper,

Thanks for taking a look at this.

The `info` attribute exists on user objects in the Active Directory schema and represents the "Comment" field you see in the Active Directory Users and Computers (ADUC) GUI. You can find the Microsoft schema documentation for it here:
[https://learn.microsoft.com/en-us/windows/win32/adschema/a-info](https://learn.microsoft.com/en-us/windows/win32/adschema/a-info)

However, you are entirely right to call this out—I made a mistake trying to query it via `"comment"`. The actual LDAP attribute name is `info`. I'll remove the redundant `"comment"` line and ensure the data maps correctly.

Fix incoming shortly!

B0TAxy:
Heads up on the latest commit—I stripped out the parsing logic for msDS-ManagedPassword.

It turns out this is a dynamically constructed attribute generated by the DC on the fly, so it doesn't physically exist in the offline ntds.dit file. No need for us to parse a BLOB that will always be empty offline! We still get all the access control info we need from msDS-GroupMSAMembership. The remaining PR enhancements are ready for review.

PR #1605, opened 2026-03-05
Improve debug message and added postmortem mode

### LINKED ISSUE #1604: Improve `debug` message and options in `target-shell`
We can all agree that the current message in `target-shell` is a bit unclear when an unhandled exception happens:

```python
*** Unhandled error: 'utf-8' codec can't encode character '\udce6' in position 57: surrogates not allowed
If you wish to see the full debug trace, enable debug mode.
```

It doesn't really tell the user how to enable debug mode, so the message could be improved. Some suggestions:

This is self-documenting by telling them exactly what to type:
```python
"If you wish to see the full debug trace, type 'debug' to enable debug mode."
```

Or even more direct:
```python
"Tip: type 'debug' to enable debug mode and see full tracebacks."
```

```python
"For a full traceback, enable debug mode with 'debug on'."
```

Other than improving the message, I also want to propose to add a [pdb.post_mortem](https://docs.python.org/3/library/pdb.html#pdb.post_mortem) mode, eg via: `debug post-mortem` or `debug pm` for short. Which drops you into a Pdb (or ipdb where available) debugger after an exception.

yunzheng:
This PR improves the message on how to turn on debug mode in `target-shell`.

It also adds another debug mode using `debug pm` or `debug postmortem`, to enable postmortem mode. This will drop you into a (i)Pdb shell after an exception.

The current message now looks like this:

```python
*** Unhandled error: can only concatenate str (not "NoneType") to str

Tip: type 'debug on' to enable debug mode and see full tracebacks.
     type 'debug pm' to drop into a post-mortem debugger on unhandled exceptions.

trixie:/$ 
```

resolves #1604

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -1618,8 +1618,8 @@ def create_cli(targets: list[Target], cli_cls: type[TargetCmd]) -> cmd.Cmd | Non
     return cli
 
 
-def run_cli(cli: cmd.Cmd) -> None:
-    """Helper method for running a cmd.Cmd cli.
+def run_cli(cli: ExtendedCmd) -> None:
+    """Helper method for running a :class:`ExtendedCmd` cli.
 
     Loops cli.cmdloop(), skipping KeyboardInterrupts. This is done so
     that ctrl+c doesn't exit the shell but only resets the current line.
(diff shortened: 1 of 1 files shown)


dissect/target/tools/shell.py
```diff
@@ -1597,12 +1627,20 @@ def run_cli(cli: cmd.Cmd) -> None:
             print()
 
         except Exception as e:
-            if cli.debug:
+            debug_mode = getattr(cli, "debug", DebugMode.OFF)
```
Schamper: Why the change to `getattr`?

Schamper: The error message doesn't really make sense anymore if this is a bare `cmd.Cmd` (there's no debug mode on those). Maybe just change the typing. I don't think we use this anywhere else where we don't use `ExtendedCmd`.

Issue #1362, opened 2025-10-16
Specific order of args specified to Yara hits on itself.


nmhawry:
When running the following command, if an argument is placed between --rules and the target path, yara will run the rule file against itself instead of file.txt in the example below.

target-query -f yara --rules <rule_path> --direct file.txt

however, when adding the equal sign to rules, it works correctly:

target-query -f yara --rules=<rule_path> --direct file.txt

PR #1619, opened 2026-03-13
Improve parsing of plugin arguments in target-query

### LINKED ISSUE #1508: Target-query runtime arguments interpreted as TARGET
The `target-query` command below incorrectly parses `value` as a `TARGET`.
```
target-query --function function --argument value /path/to/target
```
This is unexpected as `target-query -h` states that it expects `target-query [n.. arguments] [TARGETS ...]`. The `--argument` in question can be any argument from a plugin's `@arg` decorator.

Changing the command to `--argument=value` fixes this for now, but perhaps the argparser needs to be adjusted accordingly.

yunzheng:
This PR implements the suggested solution described in #1508, the described edgecase is left open.

It mainly gets the supported argparse arguments from the requested plugins to `target-query`, and then reparses the arguments again so it properly parses known plugin flags. This essentially fixes the problem described in the issue.

Unknown arguments/flags now also throw an error, which before would go unnoticed.

resolves #1508, #1362

### DIFF
--- a/dissect/target/tools/query.py
+++ b/dissect/target/tools/query.py
@@ -71,8 +71,10 @@ def main() -> int:
         fromfile_prefix_chars="@",
         formatter_class=help_formatter,
         add_help=False,
+        conflict_handler="resolve",
     )
     parser.add_argument("targets", metavar="TARGETS", nargs="*", help="Targets to load")
+    parser.add_argument("-h", "--help", action="store_true", help="show this help message and exit")
     parser.add_argument("--direct", action="store_true", help="treat TARGETS as paths to pass to plugins directly")
 
     configure_plugin_arguments(parser)
@@ -113,10 +115,31 @@ def main() -> int:
 
     args, rest = parser.parse_known_args()
 
-    # Show help for target-query
-    if not args.function and ("-h" in rest or "--help" in rest):
+    # Show help for target-query (when --help is specified without a function)
+    if not args.function and args.help:
         parser.print_help()
         return 0
+
+    # Dynamically add plugin arguments for the specified function(s)
+    for func_desc in find_and_filter_plugins(
+        functions=args.function or "",
+        target=None,
+        excluded_func_paths=args.excluded_functions,
+    ):
+        plugin_args = func_desc.args
+        for arg in plugin_args:
+            opts, kwargs = arg
+            kwargs.pop("group", None)  # Remove 'group' if it exists, as this is a special argument used in dissect.
+            parser.add_argument(*opts, **kwargs)
+
+    # Re-parse arguments now that requested plugin arguments have been added.
+    # Unknown args will automatically be shown to the user here to catch any misspelled or invalid arguments early on.
+    args = parser.parse_args()
+
+    # Propagate --help argument to plugin to show the help message of the plugin.
+    rest = []
+    if args.help:
+        rest.append("--help")
 
     process_generic_arguments(parser, args)
 
--- a/dissect/target/tools/query.py
+++ b/dissect/target/tools/query.py
@@ -127,9 +127,9 @@ def main() -> int:
         excluded_func_paths=args.excluded_functions,
     ):
         plugin_args = func_desc.args
-        for arg in plugin_args:
-            opts, kwargs = arg
-            kwargs.pop("group", None)  # Remove 'group' if it exists, as this is a special argument used in dissect.
+        for opts, kwargs in plugin_args:
+            # Skip 'group' if it exists, as this is a special argument used in dissect.
+            kwargs = {k: v for k, v in kwargs.items() if k != "group"}
             parser.add_argument(*opts, **kwargs)
 
     # Re-parse arguments now that requested plugin arguments have been added.
--- a/dissect/target/tools/query.py
+++ b/dissect/target/tools/query.py
@@ -125,6 +125,7 @@ def main() -> int:
         functions=args.function or "",
         target=None,
         excluded_func_paths=args.excluded_functions,
+        plugin_paths=args.plugin_path or [],
     ):
         plugin_args = func_desc.args
         for opts, kwargs in plugin_args:
--- a/dissect/target/tools/query.py
+++ b/dissect/target/tools/query.py
@@ -128,8 +128,6 @@ def main() -> int:
     ):
         plugin_args = func_desc.args
         for opts, kwargs in plugin_args:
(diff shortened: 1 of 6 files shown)


yunzheng:
@Schamper wonder what your thoughts are on this method of parsing the cli arguments.

This PR mainly fixes the argument parsing of where to put the `<TARGET>` in your command line. Some people just made a habit to always put it as the first argument so that it always works, but that shouldn't have to be.

dissect/target/tools/query.py
```diff
@@ -113,11 +115,32 @@ def main() -> int:
...
+    # Dynamically add plugin arguments for the specified function(s)
+    for func_desc in find_and_filter_plugins(
+        functions=args.function or "",
+        target=None,
+        excluded_func_paths=args.excluded_functions,
+    ):
+        plugin_args = func_desc.args
+        for opts, kwargs in plugin_args:
+            # Skip 'group' if it exists, as this is a special argument used in dissect.
+            kwargs = {k: v for k, v in kwargs.items() if k != "group"}
```
Schamper: It is? 😄

Schamper (commented):
> @Schamper wonder what your thoughts are on this method of parsing the cli arguments.

Overall looks fine I suppose, my only concern is that I don't recall whether `find_and_filter_plugins` was super slow still or not. Maybe we can add a benchmark test for something like "time from invocation to first record output".

dissect/target/tools/query.py
yunzheng: yea man.. go dig deeper :D

yunzheng:
> > @Schamper wonder what your thoughts are on this method of parsing the cli arguments.
> 
> Overall looks fine I suppose, my only concern is that I don't recall whether `find_and_filter_plugins` was super slow still or not. Maybe we can add a benchmark test for something like "time from invocation to first record output".

`find_and_filter_plugins()` will still be called eventually. So maybe if we add caching at places (if it doesn't do this already), it should not matter much.

dissect/target/tools/query.py
yunzheng: See the `@arg` decorator of dissect and being used here:

https://github.com/fox-it/dissect.target/blob/0fbd59f443e69c01caabeada1b962ab3dcb317d8/dissect/target/plugins/filesystem/ntfs/mft.py#L157-L162


the `group=` is not a supported argument to [argparse.add_argument()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument)

So it's a dissect hack. It costed me some time to debug this as well from what I remember. Maybe prefix this with `dissect_` -> `dissect_group`, so it's easier to relate where it comes from.

Schamper: I git blamed and I decided it's best to remove this feature in it's entirety as it's undocumented, not intuitive and with this PR would cause more workarounds than it provides convenience.

yunzheng: Alright, I will check how the removal of "group" support will affect the MFT plugin, and do some tests.

yunzheng: It seems the `group="fmt"` was mainly meant to create a mutually exclusive option group in dissect. So far it's only used in the MFT plugin.

help output of `mft` plugin:

```
usage: target-query -f records [-h] [--compact] [--fs FS] [--start START] [--end END] [--macb]

Return the MFT records of all NTFS filesystems.

The Master File Table (MFT) contains primarily metadata about every file and folder on a NFTS filesystem.

If the filesystem is part of a virtual NTFS filesystem (a ``VirtualFilesystem`` with the MFT properties
added to it through a "fake" ``NtfsFilesystem``), the paths returned in the MFT records are based on the
mount point of the ``VirtualFilesystem``. This ensures that the proper original drive letter is used when
available.
When no drive letter can be determined, the path will show as e.g. ``\$fs$\fs0``.

References:
    - https://docs.microsoft.com/en-us/windows/win32/fileio/master-file-table

options:
  -h, --help     show this help message and exit
  --compact      compacts the MFT entry timestamps into a single record
  --fs FS        optional filesystem index, zero indexed
  --start START  the first MFT segment number
  --end END      the last MFT segment number
  --macb         compacts MFT timestamps into MACB bitfield (format: MACB[standard|ads]/MACB[filename])
```


If you would specify both `--macb` and `--compact` (they have both `group="fmt"`), it will give the error:

```
usage: target-query  [-h] [--compact] [--fs FS] [--start START] [--end END] [--macb]
target-query: error: argument --compact: not allowed with argument --macb
```


If we remove the custom `group` option you can specify both and no error will be raised. I think that's fine right, until we add better explicit support for this in dissect.

Let me know, then i'll update this PR with the removal of this custom option support all together.

Schamper: > I think that's fine right, until we add better explicit support for this in dissect.

In my opinion it is. While it's a nice UX to explicitly disallow it, I'd assume anyone using it would notice quickly enough. I'd say remove it now and let's create a ticket for the grouping idea we had.

yunzheng:
The windows tests fail with:

```
ERROR: Could not install packages due to an OSError: ('Connection broken: IncompleteRead(6936 bytes read, -23 more expected)', IncompleteRead(6936 bytes read, -23 more expected))
```

is that a known issue?

dissect/target/tools/query.py
yunzheng: Changes done in the PR

Schamper:
It's been happening for a while now. I deleted the CI cache, looks like that fixes it.

tests/tools/test_query.py
```diff
@@ -454,6 +477,12 @@ def mock_function(self, json: bool, compact: bool) -> None:
         )
 
         with (
+            patch.object(
+                FunctionDescriptor,
+                "args",
+                new_callable=PropertyMock,
+                side_effect=mock_query_plugin_args,
+            ),
```
Schamper, line 485: Just curious as to why this patch is necessary. Does it not work automatically?

yunzheng: it's because we mock `find_functions` to use a mocked version that returns fake plugin functions (note that this already existed before my PR). So the `args` property on the fake function needs to be mocked  or else it doesn't work.

Schamper: Ah yes. Thanks for explaining.

Schamper (approved):

yunzheng:
Nice, thanks for the review, with this merged, finally the position of your TARGET in your CLI args works as intended!

Issue #1577, opened 2026-02-25
Add sorting by time to `ls` command


yunzheng:
It would be nice to be able to sort the directory listing by time. The default flags for standard `ls` are `ls -t` and `ls -r`, so implement these.

PR #1622, opened 2026-03-16
Add benchmark for `ls_cmd` using a ext4 test image of around 1000 files

yunzheng:
This PR adds a benchmark test for the `ls_cmd` in `target-shell` and a new test fixture providing an ext4 image with around 1000 files for doing the directory listing.

The benchmark is in preparation of implementing #1577 and some other related issues, which will contain some refactoring regarding efficiency.

The test image is derived from `debian-live-13.3.0-amd64-standard.iso` and only the `/bin` dir and some files from `/etc` are copied over. The file contents in `/bin` are also all sparse (NULL bytes) so the image compresses well. The test data mainly focuses on what is useful for directory listings, such as filenames, permissions, owner and group.

The compressed test image itself is only 100kb.

### DIFF
--- a/tests/tools/test_shell.py
+++ b/tests/tools/test_shell.py
@@ -617,9 +617,11 @@ def test_shell_prompt_tab_autocomplete() -> None:
         pytest.param("-l"),  # long listing
     ],
 )
+@pytest.mark.parametrize("warm_cache", [True, False], ids=["warm_cache", "cold_cache"])
 def test_benchmark_ls_bin(
     benchmark: BenchmarkFixture,
     args: str,
+    warm_cache: bool,
     capsys: pytest.CaptureFixture,
 ) -> None:
     """Benchmark ls command with different parameters with a /bin directory containing ~1000 files.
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -626,3 +627,31 @@ def guarded_keychain() -> Iterator[None]:
     keychain.KEYCHAIN.clear()
     yield
     keychain.KEYCHAIN.clear()
+
+
+@pytest.fixture(scope="session")
+def path_debian_ext4_raw(tmp_path_factory: pytest.TempPathFactory) -> Iterator[pathlib.Path]:
```
Schamper: Is there a specific reason over decompressing this to a temporary directory vs just holding it in memory? (other than the fact that our CI runs already seem to commonly exceed GitHub Runner memory 😄)

Schamper: My only comment against this method is that it's not very uniform with all the other test data and way we deal with that. I.e. I'd like there to be a generic helper for deflating test data. Maybe for the time being we can just keep it in memory, and create an issue to track the wider problem and think of a good uniform solution for that? What do you think.

PR #1628, opened 2026-03-17
Add ProcFdPlugin for Linux process FD analysis

Torres773:
Depends on #1609

Add ProcFdPlugin for Linux process FD analysis

Implement a new plugin to analyze open file descriptors by parsing
/proc/[pid]/fd and fdinfo. This provides forensic visibility into
files, sockets, and pipes used by running processes.

The plugin yields FileDescriptorRecord which includes file offsets
and access flags. This builds upon the recent FileDescriptor
iterator addition in the proc parser.

### DIFF
--- a/dissect/target/plugins/os/unix/linux/fd.py
+++ b/dissect/target/plugins/os/unix/linux/fd.py
@@ -17,8 +17,8 @@ FileDescriptorRecord = TargetRecordDescriptor(
         ("varint", "pid"),
         ("string", "name"),
         ("varint", "fd"),
-        ("string", "path"),
-        ("uint64", "pos"),
+        ("string", "link"),
+        ("varint", "pos"),
         ("string", "flags"),
     ],
 )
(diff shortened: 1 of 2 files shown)


dissect/target/plugins/os/unix/linux/fd.py
```diff
@@ -0,0 +1,67 @@
...
+
+class ProcFdPlugin(Plugin):
+    """Linux process file descriptor plugin."""
+
+    def check_compatible(self) -> None:
+        if not self.target.has_function("proc"):
+            raise UnsupportedPluginError("proc filesystem not available")
+
+    @export(record=FileDescriptorRecord)
+    def fd(self) -> Iterator[FileDescriptorRecord]:
```
Horofic: So you can access the plugin using `filedescriptor` as well. Requires the necessary imports.
```suggestion
    @alias("filedescriptor")
    def fd(self) -> Iterator[FileDescriptorRecord]:
```

PR #1634, opened 2026-03-19
[POC]: loader args decoupled

Miauwkeru:
(The following is written by @twiggler, not @Miauwkeru)

This PR:

- Adds support for displaying help when invoking with `target-query -L local --help`
- Fixes a regression when invoking with `target-query -f hostname -L local --force-directory-fs local` failed to pick up on the `--force-directory-fs`
- refactor loader selection into `infer_loader`

Originally, I had bigger plans, namely to move the argument parsing of plugins to the top-level and constructing the target as `with_loader(..., loader_instance)`.

 I walked this back, because:
1. Loaders are tied to a specific target path, so using a single instance on multiple targets does not work.
1. We dynamically open child targets. I forgot if they if they also need loader arguments, but the bottom line is that it is hard to create loaders with a bound argument map in advance. 

Now this is solvable by using dependency injection in the form of a  `LoaderFactory` and Loader adapters which support  construction from a uri, cli, or programmatically using a loader spec:

```
┌────────────────────────────────────────────────────────┐
│ Layer A — Inputs (adapters)                            │
│   • URI adapter:   "local?force-dirfs=1"  → LoaderSpec │
│   • CLI adapter:   argparse Namespace      → LoaderSpec │
│   • Programmatic:  LoaderSpec(...)         → LoaderSpec │
└──────────────────────────┬─────────────────────────────┘
                           │  LoaderSpec (value object)
                           ▼
┌────────────────────────────────────────────────────────┐
│ Layer B — Loader options                               │
│   • Typed `Options` dataclass per loader               │
│   • No argparse, no URI, no strings                    │
└──────────────────────────┬─────────────────────────────┘
                           │  Options instance
                           ▼
┌────────────────────────────────────────────────────────┐
│ Layer C — Loaders (business logic)                     │
│   • Loader(path, options) → maps target                │
│   • Pure: knows nothing about CLI or URIs              │
└──────────────────────────┬─────────────────────────────┘
                           │  emits child LoaderSpecs
                           ▼
┌────────────────────────────────────────────────────────┐
│ Layer D — LoaderFactory (dependency inversion)         │
│   • Target.open*(factory=...)                          │
│   • Children call factory.open(child_spec)             │
└────────────────────────────────────────────────────────┘
```
This would also remove the necessity of the `args_to_uri` hack which stringifies parsed arguments back to an uri.

However, I guess such as solution wouldn't be  "pythonic" anymore.

### DIFF
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -68,8 +68,7 @@ class Loader:
     """
 
     def __init__(
-        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True,
-        **kwargs
+        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True, **kwargs
     ):
         self.path = path
         self.absolute_path = None
@@ -150,7 +149,7 @@ class Loader:
         # Pass the parsed options directly to the implementation map function
         return self._map(target, **vars(loader_options))
 
-    def _map(self) -> None:
+    def _map(self, **kwargs) -> None:
         """Maps the loaded path into a ``Target``.
 
         Args:
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -94,6 +94,7 @@ class Loader:
     @staticmethod
     def _create_parser(cls: type[Loader]) -> argparse.ArgumentParser:
         """Creates the argument parser for this loader."""
+        # Do like generate_argparse_for_method in cli.py
         parser = argparse.ArgumentParser(
             prog=f"loader:{cls.__name__.lower()}",
             description=f"Options for the '{cls.__name__}' loader.",
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -149,7 +149,7 @@ class Loader:
         # Pass the parsed options directly to the implementation map function
         return self._map(target, **vars(loader_options))
 
-    def _map(self, **kwargs) -> None:
+    def _map(self, target: Target, **kwargs) -> None:
         """Maps the loaded path into a ``Target``.
 
         Args:
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -69,7 +69,7 @@ class Loader:
 
     def __init__(
         self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True,
-        loader_args: list[str] | None = None, **kwargs
+        **kwargs
     ):
         self.path = path
         self.absolute_path = None
@@ -84,19 +84,25 @@ class Loader:
         self.parsed_query = (
             dict(urllib.parse.parse_qsl(parsed_path.query, keep_blank_values=True)) if parsed_path else {}
         )
-        self.loader_args = loader_args or []
 
-        self._parser = argparse.ArgumentParser(
-            prog=f"loader:{self.__class__.__name__.lower()}",
-            description=f"Options for the '{self.__class__.__name__}' loader.",
+        self._parser = self._create_parser(self.__class__)
+
+    @classmethod
+    def print_help(cls) -> None:
+        """Prints the help message for this loader's specific arguments."""
+        cls._create_parser(cls).print_help()
+
+    @staticmethod
+    def _create_parser(cls: type[Loader]) -> argparse.ArgumentParser:
+        """Creates the argument parser for this loader."""
+        parser = argparse.ArgumentParser(
+            prog=f"loader:{cls.__name__.lower()}",
+            description=f"Options for the '{cls.__name__}' loader.",
             add_help=False,
         )
-        for args, kwargs in  getattr(self, "__args__", []):
-            self._parser.add_argument(*args, **kwargs)
-
-    def print_help(self) -> None:
-        """Prints the help message for this loader's specific arguments."""
-        self._parser.print_help()
+        for args, kwargs in getattr(cls, "__args__", []):
+            parser.add_argument(*args, **kwargs)
+        return parser
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({str(self.path)!r})"
@@ -132,8 +138,8 @@ class Loader:
 
     def map(self, target: Target) -> None:
         """Wrapper around the _map function that handles argument passing."""
-        # The loader now parses its own arguments
-        loader_options = self._parser.parse_args(self.loader_args)
+        # The loader parses its own arguments from argv
+        loader_options, _ = self._parser.parse_known_args()
 
         # The query string can act as a default for any arguments not provided on the command line
         for key, value in self.parsed_query.items():
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -68,13 +68,7 @@ class Loader:
     """
 
     def __init__(
-        self,
-        path: Path,
-        *,
-        parsed_path: urllib.parse.ParseResult | None = None,
-        resolve: bool = True,
-        loader_kwargs: dict | None = None,
-        **kwargs,
+        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True, **kwargs
     ):
         self.path = path
         self.absolute_path = None
@@ -89,7 +83,6 @@ class Loader:
         self.parsed_query = (
             dict(urllib.parse.parse_qsl(parsed_path.query, keep_blank_values=True)) if parsed_path else {}
         )
-        self._loader_kwargs = loader_kwargs or {}
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({str(self.path)!r})"
@@ -125,19 +118,6 @@ class Loader:
 
     def map(self, target: Target) -> None:
         """Maps the loaded path into a ``Target``.
-
-        Delegates to :meth:`_map` with any loader-specific keyword arguments set at instantiation time.
-
-        Args:
-            target: The target that we're mapping into.
-        """
-        return self._map(target, **self._loader_kwargs)
-
-    def _map(self, target: Target, **kwargs) -> None:
-        """Implementation hook for :meth:`map`.
-
-        Subclasses that accept loader-specific arguments (declared via ``@arg`` decorators)
-        should override this method instead of :meth:`map`.
 
         Args:
             target: The target that we're mapping into.
--- a/dissect/target/loader.py
+++ b/dissect/target/loader.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import argparse
 import os
 import urllib.parse
 from pathlib import Path
@@ -67,7 +68,8 @@ class Loader:
     """
 
     def __init__(
-        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True, **kwargs
+        self, path: Path, *, parsed_path: urllib.parse.ParseResult | None = None, resolve: bool = True,
+        loader_args: list[str] | None = None, **kwargs
     ):
         self.path = path
         self.absolute_path = None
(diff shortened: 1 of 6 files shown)


dissect/target/loaders/local.py
```diff
@@ -39,11 +40,18 @@
...
+    action="store_true",
+    help="Fallback to DirectoryFilesystem if a filesystem cannot be opened.",
+)
 class LocalLoader(Loader):
     """Load local filesystem."""
 
     def __init__(self, path: Path, **kwargs):
-        kwargs["parsed_path"] = urllib.parse.urlparse(str(path))
+        if not kwargs.get("parsed_path"):
+            kwargs["parsed_path"] = urllib.parse.urlparse(str(path))
```
twiggler, line 54: reviewer: do not overwrite parameters from args_to_uri, if available

This is necessary to make `--fallback-to-directory-fs` work

dissect/target/tools/utils/cli.py
```diff
@@ -499,14 +520,27 @@ def args_to_uri(targets: list[str], loader_name: str, args: list[str] | None = N
...
     loader = LOADERS_BY_SCHEME.get(loader_name, None)
-
     parser = argparse.ArgumentParser(
         argument_default=argparse.SUPPRESS, description="\n".join(textwrap.wrap(get_docstring(loader)))
     )
     for load_arg in getattr(loader, "__args__", []):
-        parser.add_argument(*load_arg[0], **load_arg[1])
-    args = vars(parser.parse_known_args(args)[0])
-    return [f"{loader_name}://{target}" + (("?" + urllib.parse.urlencode(args)) if args else "") for target in targets]
+        arg_opts, arg_kwargs = load_arg
```
twiggler, line 569: Reviewer: force argparse to use the long option name to stop argparse from converrting to underscores.

Or perhaps you *want* it  converted to underscores? In any case, the tests suggest that it should work with hyphens. But who knows.

This is necessary to make `--fallback-to-directory-fs  and friends work

dissect/target/tools/query.py
```diff
@@ -103,21 +102,63 @@ def main() -> int:
...
+            if result is not None:
+                _, load_path, parsed_path = result
+            else:
+                load_path = args.targets[0]
+                parsed_path = None
+
+        try:
+            loader_instance = loader_cls(load_path, parsed_path=parsed_path, loader_kwargs=loader_kwargs)
+        except Exception as e:
+            parser.error(f"Failed to instantiate loader '{loader_cls.__name__}': {e}")
```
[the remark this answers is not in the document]
twiggler: This problem was fixable but in the end I walked it back (see PR description)

Miauwkeru (commented):
Maybe an odd expectation from me, but from a usage perspective I would expect that passing the command line argument without `-L <loader>` for a specific loader would still get passed to the loader.

So that the following case would work:
> target-query local --force-fallback-fs

with any of the `target-*` tooling.

Tho maybe that was reserved for a future iteration?

dissect/target/tools/utils/cli.py
Miauwkeru: wouldn't it be fine with underscores? besides, the reason we can access the value in the namespace is because it has underscores. so you can do `ns.fallback_to_directory_fs`
Maybe just replace the underscores in the `k` part of params?

dissect/target/loaders/local.py
Miauwkeru: yeah, i see it's because otherwise the output of args_to_uri doesn't get passed properly to the loader.

dissect/target/tools/utils/cli.py
twiggler: > wouldn't it be fine with underscores? 

The local loader and the the tests expect hyphens.

> besides, the reason we can access the value in the namespace is because it has underscores. so you can do ns.fallback_to_directory_fs

Where is `ns` accessed this way?  I don't see `ns` escaping the function?

> Maybe just replace the underscores in the k part of params?

It is simpler, but I think it suffers from two problems:

1. The current approach  copies the original kwargs first (kw = dict(arg_kwargs)), so if the author provided an explicit dest it stays untouched — you never overwrite an explicit dest.

1. If no dest was provided, it inspects arg_opts (the real option strings passed to add_argument) and picks the canonical long option (long_opts = [o for o in arg_opts if o.startswith("--")]), then sets kw["dest"] = long_opts[0].lstrip("-"). That uses the actual --long-option string the developer registered, not a heuristic post‑parse rename.

Miauwkeru: > If no dest was provided, it inspects arg_opts (the real option strings passed to add_argument) and picks the canonical long option (long_opts = [o for o in arg_opts if o.startswith("--")]), then sets kw["dest"] = long_opts[0].lstrip("-"). That uses the actual --long-option string the developer registered, not a heuristic post‑parse rename.

Fair enough, it is more in line with what the developer specified, instead of the assumption it never used a `_` in the command line argument.

> Where is ns accessed this way? I don't see ns escaping the function?

I was talking about the ArgumentNamespace that gets created with `parse_known_args`. So the `args.fallback_to_directory_fs` like its used in any of the command line tools.

But not very important anymore related to the second problem you illustrated

twiggler:
> Maybe an odd expectation from me, but from a usage perspective I would expect that passing the command line argument without `-L <loader>` for a specific loader would still get passed to the loader.
> 
> So that the following case would work:
> 
> > target-query local --force-fallback-fs
> 
> with any of the `target-*` tooling.
> 
> Tho maybe that was reserved for a future iteration?

Implemented

dissect/target/tools/utils/cli.py
```diff
@@ -488,25 +520,64 @@ def wrapper(*args, **kwargs) -> int:
...
-        argument_default=argparse.SUPPRESS, description="\n".join(textwrap.wrap(get_docstring(loader)))
+        argument_default=argparse.SUPPRESS, description="\n".join(textwrap.wrap(get_docstring(loader_cls)))
     )
-    for load_arg in getattr(loader, "__args__", []):
-        parser.add_argument(*load_arg[0], **load_arg[1])
-    args = vars(parser.parse_known_args(args)[0])
-    return [f"{loader_name}://{target}" + (("?" + urllib.parse.urlencode(args)) if args else "") for target in targets]
+    for load_arg in getattr(loader_cls, "__args__", []):
+        arg_opts, arg_kwargs = load_arg
+        kw = dict(arg_kwargs)
```
Miauwkeru: isn't arg_kwargs already a dictionary?

Miauwkeru (commented):
From what I have seen everything seems fine. Tho we need to make a note of it that you can't combine uri args and command line arguments. E.g. the following only uses the `fallback-to-directory-fs` 

> target-query local://fallback-to-directory-fs=1 -f osinfo --force-directory-fs

twiggler:
> From what I have seen everything seems fine. Tho we need to make a note of it that you can't combine uri args and command line arguments. E.g. the following only uses the `fallback-to-directory-fs`
> 
> > target-query local://fallback-to-directory-fs=1 -f osinfo --force-directory-fs

Is this what you mean with note? https://github.com/fox-it/dissect.target/blob/459f6ba50746e38c8f58c7b589329cdb0f4af8cb/dissect/target/tools/utils/cli.py#L533-L545

dissect/target/tools/utils/cli.py
twiggler: Yes it is; I am creating a defensive shallow copy, because I don't want to mutate the  arguments on the loader class as a side effect when doing `kw["dest"] = long_opts[0].lstrip("-")`. 
(Although this would be correct by accident)

Shall I add a comment?

Miauwkeru: The reason why is probably useful, otherwise someone might unknowingly remove it without understanding the consequences.

Miauwkeru:
> > From what I have seen everything seems fine. Tho we need to make a note of it that you can't combine uri args and command line arguments. E.g. the following only uses the `fallback-to-directory-fs`
> > > target-query local://fallback-to-directory-fs=1 -f osinfo --force-directory-fs
> 
> Is this what you mean with note?
> 
> https://github.com/fox-it/dissect.target/blob/459f6ba50746e38c8f58c7b589329cdb0f4af8cb/dissect/target/tools/utils/cli.py#L533-L545

yes, like that. But also in the dissect-docs and the eventual release notes. Because I don't think most of the users will look at the docstring of any non developers will look inside the code and read the docstring of specifically `args_to_uri`. So it might be good to also add it to a more visual / accessible place so people might actually read and understand it.

twiggler:
> > > From what I have seen everything seems fine. Tho we need to make a note of it that you can't combine uri args and command line arguments. E.g. the following only uses the `fallback-to-directory-fs`
> > > > target-query local://fallback-to-directory-fs=1 -f osinfo --force-directory-fs
> > 
> > 
> > Is this what you mean with note?
> > https://github.com/fox-it/dissect.target/blob/459f6ba50746e38c8f58c7b589329cdb0f4af8cb/dissect/target/tools/utils/cli.py#L533-L545
> 
> yes, like that. But also in the dissect-docs and the eventual release notes. Because I don't think most of the users will look at the docstring of any non developers will look inside the code and read the docstring of specifically `args_to_uri`. So it might be good to also add it to a more visual / accessible place so people might actually read and understand it.

Agreed, I will make a pr on dissect-docs

dissect/target/tools/query.py
```diff
@@ -113,9 +115,22 @@ def main() -> int:
 
     args, rest = parser.parse_known_args()
 
-    # Show help for target-query
+    # Show help for target-query. If a loader can be determined, show loader-specific help.
     if not args.function and ("-h" in rest or "--help" in rest):
```
Miauwkeru, line 119: Is it an idea to show help for both the plugins and the loader (if specified)?

Doesn't necessarily have to be in this pr

Miauwkeru:
I don't really see anything else, @Schamper what are your thoughts?

dissect/target/loader.py
```diff
@@ -153,6 +154,32 @@ def map(self, target: Target) -> None:
         raise NotImplementedError
 
 
+def infer_loader(
```
Schamper, line 157:
```suggestion
def infer(
```
All the other functions use verb naming, as most of them will be used as attribute on the module.

As such, I'd suggest also changing bare usage of `infer_loader` to `loader.infer`.

dissect/target/loader.py
```diff
@@ -153,6 +154,32 @@ def map(self, target: Target) -> None:
...
+    spec: str | Path,
+) -> tuple[type[Loader], Path, urllib.parse.ParseResult | None] | None:
+    """Infer the loader class for a given target specification.
+
+    Uses the same logic as :meth:`Target.open_all <dissect.target.target.Target.open_all>` to determine
+    which :class:`Loader` class would be used for the given ``spec``.
```
Schamper, line 163: This comment doesn't make sense? That logic has moved here.

dissect/target/tools/utils/cli.py
```diff
@@ -491,25 +523,64 @@ def wrapper(*args, **kwargs) -> int:
...
         ``target-query /evtxs/* -L log --log-hint="evtx" -f evtx``
     into:
         ``target-query "log:///evtxs/*?hint=evtx" -f evtx``
 
     For loaders providing ``@arg()`` arguments.
+
+    If ``target`` is already a URI with a scheme it is returned unchanged; any
+    CLI loader args are ignored because the URI is the authoritative
+    specification.  When ``-L`` is specified alongside a URI target, the
+    explicitly requested loader overrides the one encoded in the URI scheme.
```
Schamper, line 539: Can you remove the Claude-isms from all docstrings? E.g. double spaces and use of ;

dissect/target/tools/utils/cli.py
```diff
@@ -491,25 +523,64 @@ def wrapper(*args, **kwargs) -> int:
...
-    args = vars(parser.parse_known_args(args)[0])
-    return [f"{loader_name}://{target}" + (("?" + urllib.parse.urlencode(args)) if args else "") for target in targets]
+    for load_arg in getattr(loader_cls, "__args__", []):
+        arg_opts, arg_kwargs = load_arg
+        kw = dict(arg_kwargs)  # Defensive copy to avoid mutating the loader class
+        # If no explicit dest provided, prefer the long option string as dest
+        if "dest" not in kw:
+            long_opts = [o for o in arg_opts if o.startswith("--")]
+            if long_opts:
+                kw["dest"] = long_opts[0].lstrip("-")
```
Schamper, line 575: Maybe I missed it but I don't see this explicitly tested?

Schamper (changes requested):

PR #1635, opened 2026-03-19
Support preservation of symlinks for non-windows systems when using `cp`

### LINKED ISSUE #1636: CP cannot preserve symlinks
Normally, `cp` can preserve symlinks and other properties using the `--preserve` flag (among many others). There is no way to preserve these attributes currently.

cochaviz:
# Support preservation of symlinks for non-windows systems when using `cp`

As introduced in #1078, symlinks are not preserved, but this also includes other attributes which might be relevant for forensics. This PR introduces, firstly, support for preserving symlinks specifically, but could also include other attributes such as `owner`, `timestamp`, etc.

## Proposed Changes

We want to introduce the various flags and options that `cp` supports for preserving flags (which are weirdly numerous? https://www.man7.org/linux/man-pages/man1/cp.1.html):

```
       -d     same as --no-dereference --preserve=links

[...]

       -P, --no-dereference
              never follow symbolic links in SOURCE

[...]

       --preserve[=ATTR_LIST]
              preserve the specified attributes
```

This addition allows all the above variations, showing warnings when the included attribute is no supported:

```
target-fs <file> -P some-symlinked-file
```

Which preserves the `links` attribute. Here an example of an unsupported attribute:

```
$ target-fs <file> --preserve=owner some-symlinked-file
2026-03-19T15:58:17.468205Z [warning  ] <Target missing_symlink.zip>: Failed to find OS plugin, falling back to default [dissect.target.target]
2026-03-19T15:58:17.468448Z [warning  ] Unsupported --preserve attributes (will be ignored): owner [dissect.target.tools.fs]
```

Closes #1636.

## Checklist


Schamper:
I can't see what the changes were in this PR, but regardless of #1690, I think I'd still like the `cp` command in `target-fs` (or `save` in `target-shell`) to preserve symlinks where possible.

I don't much care for preservation of any other attributes.

PR #1641, opened 2026-03-28
Rename CLI Plugins to make the names more inline with the whole project

### LINKED ISSUE #1627: Rename or alias `target-query` for first-time users
Right now, if we install `dissect.target`, we get the command `target-query` while I would expect something like `dissect target-query` or `dissect-target`, etc..

Having never used Dissect, this was rather confusing. I understand this has long been the case so just renaming is unreasonable, but perhaps we can introduce an alias?

Qwekkeboom:
# Rename CLI Plugins to make the names more inline with the whole project

When installing the dissect package the cli used is mostly `target-query` which can be a bit confusing

## Proposed Changes

Prepend `dissect-` to all the cli entries. Resolves #1627.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,20 +152,20 @@ dev = [
 ]
 
 [project.scripts]
-target-build-magic = "dissect.target.tools.build_magic:main"
-target-build-pluginlist = "dissect.target.tools.build_pluginlist:main"
-target-dd = "dissect.target.tools.dd:main"
-target-inspect = "dissect.target.tools.inspect:main"
-target-diff = "dissect.target.tools.diff:main"
-target-dump = "dissect.target.tools.dump:main"
-target-fs = "dissect.target.tools.fs:main"
-target-info = "dissect.target.tools.info:main"
-target-mount = "dissect.target.tools.mount:main"
-target-qfind = "dissect.target.tools.qfind:main"
-target-query = "dissect.target.tools.query:main"
-target-reg = "dissect.target.tools.reg:main"
-target-shell = "dissect.target.tools.shell:main"
-target-yara = "dissect.target.tools.yara:main"
+dissect-target-build-magic = "dissect.target.tools.build_magic:main"
+dissect-target-build-pluginlist = "dissect.target.tools.build_pluginlist:main"
+dissect-target-dd = "dissect.target.tools.dd:main"
+dissect-target-inspect = "dissect.target.tools.inspect:main"
+dissect-target-diff = "dissect.target.tools.diff:main"
+dissect-target-dump = "dissect.target.tools.dump:main"
+dissect-target-fs = "dissect.target.tools.fs:main"
+dissect-target-info = "dissect.target.tools.info:main"
+dissect-target-mount = "dissect.target.tools.mount:main"
+dissect-target-qfind = "dissect.target.tools.qfind:main"
+dissect-target-query = "dissect.target.tools.query:main"
+dissect-target-reg = "dissect.target.tools.reg:main"
+dissect-target-shell = "dissect.target.tools.shell:main"
+dissect-target-yara = "dissect.target.tools.yara:main"
 
 [tool.ruff]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,19 +153,33 @@ dev = [
 
 [project.scripts]
 dissect-target-build-magic = "dissect.target.tools.build_magic:main"
+target-build-magic = "dissect.target.tools.build_magic:main"
 dissect-target-build-pluginlist = "dissect.target.tools.build_pluginlist:main"
+target-build-pluginlist = "dissect.target.tools.build_pluginlist:main"
 dissect-target-dd = "dissect.target.tools.dd:main"
+target-dd = "dissect.target.tools.dd:main"
 dissect-target-inspect = "dissect.target.tools.inspect:main"
+target-inspect = "dissect.target.tools.inspect:main"
 dissect-target-diff = "dissect.target.tools.diff:main"
+target-diff = "dissect.target.tools.diff:main"
 dissect-target-dump = "dissect.target.tools.dump:main"
+target-dump = "dissect.target.tools.dump:main"
 dissect-target-fs = "dissect.target.tools.fs:main"
+target-fs = "dissect.target.tools.fs:main"
 dissect-target-info = "dissect.target.tools.info:main"
+target-info = "dissect.target.tools.info:main"
 dissect-target-mount = "dissect.target.tools.mount:main"
+target-mount = "dissect.target.tools.mount:main"
 dissect-target-qfind = "dissect.target.tools.qfind:main"
+target-qfind = "dissect.target.tools.qfind:main"
 dissect-target-query = "dissect.target.tools.query:main"
+target-query = "dissect.target.tools.query:main"
 dissect-target-reg = "dissect.target.tools.reg:main"
+target-reg = "dissect.target.tools.reg:main"
 dissect-target-shell = "dissect.target.tools.shell:main"
+target-shell = "dissect.target.tools.shell:main"
 dissect-target-yara = "dissect.target.tools.yara:main"
+target-yara = "dissect.target.tools.yara:main"
 
 [tool.ruff]
 line-length = 120


cochaviz:
> There is also https://docs.astral.sh/uv/reference/cli/#uv-tool-install, perhaps this is the canonical way to make tools permanently available

I'm not too sure about all the different build methods/features in Python (there are so many), but I don't think relying on `uv` for installation is the most pip-compatible idea. Personally, I would expose the different operations (query, shell, etc) as subcommands under the module name:

```
dissect-target query/shell/etc.. <params>
``` 

Doing `dissect<TAB>` for completion would then clearly show dissect-target (the module you just installed). Furthermore, `dissect-target --help` would show all the different subcommands, usage examples, ,etc.

The advantage is that we would not need to change anything about the build process since, as far as I can see, this could be implemented by creating a `__main__.py` in the `dissect/target` folder and adding `dissect-target = "dissect.target:main"` to the `[project.scripts]` section in `pyproject.toml`.

The `dissect` command would, of course be the most elegant solution! But I think this will be much easier to do when we've got everything in the monorepo.

Schamper:
> From a user perspective I personally like this:
> 
> `dissect query ...` or `dissect shell ...` ,

That's certainly a direction we will take, but https://github.com/fox-it/dissect/issues/104 is there for that. I specifically was asking your opinion on an implementation detail of where `dissect-`-prefixed scripts should live: the `dissect` package or each individual package. But I think I've made up my mind on this, and it would make more sense for each individual package to be responsible for its own "aliases", like this PR is doing for `dissect.target`.

> I know we're not too keen on additional dependencies, but I would love y'alls opinion on `typer` (https://typer.tiangolo.com/#example)

I'm generally not a fan of these kind of libraries in Python. In Rust I love `clap`, but there it gets compiled in and it's a billion times faster. In Python, we pay the startup cost of invoking these "fancy" argument parsers on every invocation, for (IMO) little benefit. And the startup cost of `target-*` is already quite big. Additionally, our argument parsing gets quite complex (see https://github.com/fox-it/dissect.target/pull/1619 as example) and I think we'd quickly run into limitations.

It definitely looks nice, I'll probably use it for small scripts of my own, but I don't think it's a good fit in Dissect.

> Personally, I would expose the different operations (query, shell, etc) as subcommands under the module name:
> 
> ```
> dissect-target query/shell/etc.. <params>
> ```
> 
> Doing `dissect<TAB>` for completion would then clearly show dissect-target (the module you just installed). Furthermore, `dissect-target --help` would show all the different subcommands, usage examples, etc.

I hadn't thought of this idea yet, but that could be nice to have. Can you check if this works with `uvx`? As in, would `uvx dissect.target query ...` work?

Would we still want aliases like `dissect-target-query` and `dissect-target-shell` (and [`dissect-dump-nskeyedarchiver`](https://github.com/fox-it/dissect.util/blob/568a93fafaab98ad44e3cad5e7fed7189021dbb7/pyproject.toml#L57) etc)? It goes a little bit against the Python ethos of "one clear way to do something" if we have both. With your idea, instead of simply "duplicating" all commands with a `dissect-` prefix, we'd have:
- A normal, non-prefixed version of all scripts
- A single additional, per-package, utility script that (hopefully) works with `uvx`

Which seems quite elegant. 

To summarize, we could have:
- A `dissect` script entry in the `dissect` package to call into "popular" tools (https://github.com/fox-it/dissect/issues/104)
- A per-package `dissect-<package>` script entry to call into each script of that package
- Regular per-package script entries with just normal script names

This approach would have the benefit of gaining _some_ discoverability (you don't get _all_ script entries on a tab-complete), but with the benefit of not blowing up both the script entry list, but also the tab-complete list upon a `dissect-<tab>`.

@yunzheng @JSCU-CNI pulling you into this discussion for your opinions.

twiggler:
> That's certainly a direction we will take, but https://github.com/fox-it/dissect/issues/104 is there for that. I specifically was asking your opinion on an implementation detail of where dissect--prefixed scripts should live: the dissect package or each individual package. But I think I've made up my mind on this, and it would make more sense for each individual package to be responsible for its own "aliases", like this PR is doing for dissect.target.

 I would rather go for a `dissect` command immediately and bypass this step. 
 
 With regards to the implementation, I don't think it matters much, certainly not in the contex of the monorepository

cochaviz:
## On `uvx dissect.target` testing

Okay, so testing this is a bit difficult since it seems that `uvx` only works 'well' in the context of distributed packages. However, if we run:

```bash
vscode ➜ /workspaces $ uvx dissect.target --help
An executable named `dissect.target` is not provided by package `dissect-target`.
The following executables are available:
# [...]

vscode ➜ /workspaces $ uvx dissect-target --help
An executable named `dissect-target` is not provided by package `dissect-target`.
The following executables are available:
# [...]
```

Which suggests that `uvx` looks for a package with name `argv[1].replace(".", "-")` in the second argument, and uses the unedited `argv[1]` for the lookup in `[project.scripts]`. I'm very 50/50 whether this means whether it will support `uvx dissect.target`, but if you have any proposals for testing, please let me know.

**Also**: I've also added a call to `cli:main` in `dissect/target/__main__.py` so module executions are also possible (`uv run -m dissect.target`), but I'm not sure whether that's desirable?

## Other formats

> Would we still want aliases like `dissect-target-query` and `dissect-target-shell` (and [`dissect-dump-nskeyedarchiver`](https://github.com/fox-it/dissect.util/blob/568a93fafaab98ad44e3cad5e7fed7189021dbb7/pyproject.toml#L57) etc)? It goes a little bit against the Python ethos of "one clear way to do something" if we have both.

Yeah I'm personally not a big fan of this. It would overflow the auto-completion, like you suggest, making guiding new users much harder.

**Note**: The `dissect.util` package actually seems a little strange in a way? It exposes a lot of functionality in the form of a package, but only a single--rather specific--script?

## Generic `dissect` top-level command

> I would rather go for a dissect command immediately and bypass this step.
> 
> With regards to the implementation, I don't think it matters much, certainly not in the contex of the monorepository

On second thought, this might work well! I'm new to the code base, but as long as `dissect-target` is the only package that exposes a CLI, I don't see why we couldn't just make `dissect` reference that package. That is, allow:

```bash
$ dissect --help
usage: dissect[-h] {build-magic,build-pluginlist,dd,diff,dump,fs,info,inspect,mount,qfind,query,reg,shell,yara} ...

positional arguments:
  {build-magic,build-pluginlist,dd,diff,dump,fs,info,inspect,mount,qfind,query,reg,shell,yara}
    build-magic         Build magic signatures
    build-pluginlist    Build the plugin list
    dd                  Copy target data
# [...]

options:
  -h, --help            show this help message and exit
```

By introducing the `dissect` command in the `[project.scripts]` in the `dissect` metapackage and make it reference `dissect-target`:

```toml
dissect = "dissect.target.cli:main"
```

This would solve the prior `uvx` `.` or `-` issue as well, allowing `uvx dissect --help`.

Furthermore, this would be completely backwards-compatible and even _better_ if we're considering a monorepo, as it allows for updating the documentation to the use of `dissect` over `dissect-target`.

What do you think @twiggler?

Edit: re-reading the thread, we already discussed this somewhat. The question is, do we want to support all commands, or only the ones in `dissect-target` which claims to be the user-facing cli?

Schamper:
> I would rather go for a dissect command immediately and bypass this step.

But would the `dissect` command also make available everything in e.g. `dissect.database`? Or just a select set of "blessed" tools? I.e. `dissect util dump-nskeyedarchiver`, `dissect database ntds`?

Note that it's not a given that the `dissect` package is installed. Do we forego the use-case where someone only installs `dissect.database`? They may not want to have to deal with all the dependencies just to run a small utility script in `dissect.database`.

> On second thought, this might work well! I'm new to the code base, but as long as dissect-target is the only package that exposes a CLI, I don't see why we couldn't just make dissect reference that package.

It is not. See the point above.

twiggler:
> But would the dissect command also make available everything in e.g. dissect.database? Or just a select set of "blessed" tools? I.e. dissect util dump-nskeyedarchiver, dissect database ntds?

I don't think we have any usage statistics, but perhaps we indeed could make a judgement call for each utility. The database tools look a bit niche to me personally. Initially I was thinking more about the heavy hitters like query, shell, and mount, the scraper.    

> Note that it's not a given that the dissect package is installed. Do we forego the use-case where someone only installs dissect.database? They may not want to have to deal with all the dependencies just to run a small utility script in dissect.database.

I reckoned that there two types of users (which might be the same person): 

1.  the analyst who typically does a full installation and primarily wants to interact through the cli  and 
1. developers who want to develop their own tooling on top of individual packages / llibraries. 


In case this is too rigid, we could also go for another approach (which was perhaps already suggested somewhere)
By moving the  dissect cli functionality (i.e. the dissect program) in dissect-core, instead of putting it is the metapackage,  we could advertise  the entrypoints of tool logic of for example dissect.database in its pyproject.toml. In other words, we move the orchestrator to dissect-core and keep the implementation in the modules.

Then analysts can still install the dissect meta package (which includes dissect-core), and developers can install dissect-core together with dissect.database. So we maintain a lean environment if necessary, at the cost of some extra complexity. 

One win would be a more polished and professional UX because this is centralized in dissect.core by standardizing layout, standardizing global arguments, centralized error handling etc, better discoverability etc. 

(brainstorming)

Schamper:
@twiggler and 3), a senior analyst in a restricted environment. I think your separation between 1) and 2) is too simplistic. The goal for the separation is primarily senior analyst focussed, not for developers. Unless I end up being the last-man-standing with this opinion, but I still believe every individual package should be as usable-as-can-be in isolation.

yunzheng:
I agree with @twiggler that we should skip prefixing commands with `dissect-` and directly go for the `dissect` meta command. It would make more sense that this lives in the `dissect` meta package and not in `dissect.target` though.

Each package should still have the ability to have its own entrypoints defined where applicable, just like `dissect.target` already has. So no change needed here.

However, the issue #1627 is mainly about new time users and improving discoverability of `target-` related tools. So does having a `dissect` meta command solves this problem?. I think it does. If we all agree on that the choice is pretty straightforward.

Some additional comments:
* I think we should go for blessed commands initially.
  * Although maybe not very discoverable, non blessed commands can still be called via their original entrypoints when needed.
  * Can we assume people always just use `dissect` to use the dissect suite? and if you really need one specific thing only, it would be a power user and knows how to find the alternative entrypoint of the script?
* Putting some via `util` or `database` is a nice idea that we can explore further

PR #1644, opened 2026-03-30
Expose Target arguments to Plugin instances

### LINKED ISSUE #1493: Expose plugin arguments to `Plugin.check_compatible`
For some plugins their compatibility depends on the given arguments at runtime (e.g. given paths, etc.). Currently there is no way to determine in `Plugin.check_compatible` if said plugin will be compatible at runtime. Exposing the `parser` of the function or plugin would enable that check before calling the plugin function or method.

JSCU-CNI:
This PR implements and fixes https://github.com/fox-it/dissect.target/issues/1493. The `Plugin` class is extended with a `get_args()` method, which returns a custom `argparse.Namespace` object populated with rest arguments from `Target`.

This allows plugins to determine if they are compatible based on plugin arguments provided at runtime. For an example see the included changes for the `EnvironmentFilePlugin` plugin.

Interested to hear your thoughts on this @Schamper and @yunzheng, especially in combination with the proposed changes in https://github.com/fox-it/dissect.target/pull/1619.

### DIFF
--- a/dissect/target/plugin.py
+++ b/dissect/target/plugin.py
@@ -24,11 +24,16 @@ import dissect.target.plugins.os.default as default
 from dissect.target.exceptions import PluginError, PluginNotFoundError, UnsupportedPluginError
 from dissect.target.helpers import cache
 from dissect.target.helpers.fsutil import has_glob_magic
+from dissect.target.helpers.lazy import import_lazy
 from dissect.target.helpers.logging import get_logger
 from dissect.target.helpers.record import EmptyRecord
 from dissect.target.helpers.utils import StrEnum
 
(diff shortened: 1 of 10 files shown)


JSCU-CNI:
I think I can rewrite this feature on top of https://github.com/fox-it/dissect.target/pull/1619 once that is merged.

yunzheng:
> I think I can rewrite this feature on top of #1619 once that is merged.

#1619 has been merged, let the man cook

JSCU-CNI:
> https://github.com/fox-it/dissect.target/pull/1619 has been merged, let the man cook

Dinner is ready :cook:

PR #1645, opened 2026-03-30
Fix plugin argparse help messages

JSCU-CNI:
This PR fixes some bugs in the current argparse usage template generation for plugins.

Before:
```
$ target-query -f example_namespace -h
usage: target-query -f __call__ [-h]

Return the records of all exported methods.

Raises:
    PluginError: If the subclass is not a namespace plugin.

options:
  -h, --help  show this help message and exit
```

After:
```
$ target-query -f example_namespace -h
usage: target-query -f example_namespace [-h]

`ExampleNamespacePlugin` (`dissect.target.plugins.general.example.ExampleNamespacePlugin`)

    Example namespace plugin.

    `example_namespace.example_record` (output: records)

        Example namespace export.

options:
  -h, --help  show this help message and exit
```

The last test in `test_cli.py` depends on #1644


JSCU-CNI:
Added these changes to #1644.

PR #1648, opened 2026-03-31
Change target-dump output to use plugin name for files

### LINKED ISSUE #1489: target-dump cannot handle conflicting functions
When specifying 2 functions in `target-dump` that produce the same kind of records it will error.

For example:
```
target-dump -f "webserver.access,webserver.logs" -t <target>
```

Will error with the following error

```
2026-01-12T13:36:59.250474Z [error    ] Can not write to a clean finished sink [dissect.target.tools.dump] sink="Sink(target_path='testfile.vma', func='webserver.access', path=PosixPath('output/application_log_webserver_access.jsonl'), is_dirty=False, record_count=1, size_bytes=1125)" state=output/target-dump.state.json                                                                                                                                                                              
2026-01-12T13:36:59.250728Z [error    ] Exception while running the pipeline [dissect.target.tools.dump] exc='Can not write to a clean finished sink'
```

This happens because both functions produce an `application_log_webserver_access` record type. This results in quite a lot of troubleshooting to see which functions can be run together, and which cannot.

Miauwkeru:
- **Change the output file to use FunctionDescriptor name instead of the RecordDescriptor.name**
- **Fix some consistency issues where things didn't function as intended**
- **Adjust tests for changes to target-dump**

# Change target-dump output to use plugin names

Target-dump can't write to the same file multiple times, and as the record descriptor name was used
there is sometimes a collision depending on which plugins get used.

## Suggested changes

The solution proposed is to use the name of the plugin for the file instead.

E.g for `-f webserver.access` it will change the output structure from 

- `<OUTPUT_DIR>/<TARGET_NAME>/webserver/application_webserver_log_access.jsonl`

to

- `<OUTPUT_DIR>/<TARGET_NAME>/webserver_access.jsonl`

Closes #1489

### DIFF
--- a/dissect/target/tools/dump.py
+++ b/dissect/target/tools/dump.py
@@ -546,30 +546,14 @@ def get_nested_attr(obj: Any, nested_attr: str) -> Any:
     return functools.reduce(getattr, [obj, *parts])
 
 
-@functools.lru_cache(maxsize=DEST_DIR_CACHE_SIZE)
-def get_sink_dir_by_target(target: Target, function: FunctionDescriptor) -> Path:
-    func_first_name, _, _ = function.name.partition(".")
-    return Path(target.name) / func_first_name
-
-
-@functools.lru_cache(maxsize=DEST_DIR_CACHE_SIZE)
-def get_sink_dir_by_func(target: Target, function: FunctionDescriptor) -> Path:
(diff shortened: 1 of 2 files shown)


Matthijsy (approved):
LGTM

twiggler (approved):

PR #1650, opened 2026-04-02
Fix the side effects of module imports when running tests in a specific order

Miauwkeru:
# Fix the side effects of module imports when running tests in a specific order

This issue is best illustrated with the SMBLoader.

When testing the SMBLoader, we monkeypatch the module out of
`sys.modules`. So we can whether the import behaviour works as intended.
The mocking of the loader module (`dissect.target.loaders.smb`) occurs
at the start of that test. This only happens if that loader was inside
`sys.modules`.

However, when the loader module isn't in `sys.modules` the tests still
import the `dissect.target.loaders.smb` module.
This module contains all the references to the mocked libraries that
were used to import that module. This "tainted" module will exist during
the whole test cycle.

The biggest effect observed from this behaviour was on
`test_target.py::test_exception_invalid_path()` where it created a memory
leak during test execution which crashes the runner.
This is why the tests sometimes failed.

## Proposed Changes

`tests/_utils.py::cleanup_modules` is a context manager to handle the monkeypatching and removing
the module if it was created during the test.

### DIFF
--- a/tests/_utils.py
+++ b/tests/_utils.py
@@ -26,19 +26,22 @@ def cleanup_modules(modules: list[str], monkeypatch: pytest.MonkeyPatch) -> Iter
     """Removes or reverts modules back into sys.modules.
 
     There can be cases that a "tainted" module gets imported using monkeypatch.
-    This still has references to the mock objects and effects other tests.
+    This still has references to the mock objects and affects other tests.
     """
     prev_modules: dict[str, ModuleType | None] = {}
-    with monkeypatch.context() as m:
-        for mod in modules:
-            prev_modules.update({mod: sys.modules.get(mod)})
-            # Remove the module inside this monkeypatch context
-            if mod in sys.modules:
-                m.delitem(sys.modules, mod)
 
-        yield
+    try:
+        with monkeypatch.context() as m:
+            for mod in modules:
+                prev_modules.update({mod: sys.modules.get(mod)})
+                # Remove the module inside this monkeypatch context
+                if mod in sys.modules:
+                    m.delitem(sys.modules, mod)
 
-    for name, module in prev_modules.items():
-        if module is None and name in sys.modules:
-            # Delete the module if it was created during the test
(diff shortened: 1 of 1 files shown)


tests/_utils.py
```diff
@@ -10,3 +19,26 @@ def absolute_path(filename: str) -> Path:
...
+    """
+    prev_modules: dict[str, ModuleType | None] = {}
+    with monkeypatch.context() as m:
+        for mod in modules:
+            prev_modules.update({mod: sys.modules.get(mod)})
+            # Remove the module inside this monkeypatch context
+            if mod in sys.modules:
+                m.delitem(sys.modules, mod)
+
+        yield
```
twiggler: (pre-existing)

What if the test throws?

Perhaps wrap with try / finally?

tests/_utils.py
```diff
@@ -10,3 +19,26 @@ def absolute_path(filename: str) -> Path:
...
     for path in paths:
         root.joinpath(path).mkdir(parents=True)
+
+
+@contextmanager
+def cleanup_modules(modules: list[str], monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
+    """Removes or reverts modules back into sys.modules.
+
+    There can be cases that a "tainted" module gets imported using monkeypatch.
+    This still has references to the mock objects and effects other tests.
```
twiggler:
```suggestion
    This still has references to the mock objects and affects other tests.
```

tests/_utils.py
Miauwkeru: That's a good point, fixed.

Miauwkeru:
> Very nice.
> 
> It is still a pragmatic fix; perhaps a more principled fix would be to use Dependency Injection to improve testability. Or lazy imports?

I can think of swapping the system modules in a context manager, similar to how. That would remove the need to know which module you need to remove from sys.modules. I don't know what other effects it would have tho.. which is why  rather avoid it.

PR #1692, opened 2026-04-03
Add tar command to target-shell

### LINKED ISSUE #1690: Add a tar command to target-shell
I usually want to extract/save multiple files and directories from a target, but somehow I dislike the current `save` command. Mainly because you need to create the local directory first where you want to store everything and if you don't, it will just clutter your whole local directory.

Adding a `tar` command would be a nice alternative way as it can store everything in one file, while preserving most forensic data.

This could potentially also solve the issue described in #1635. A tar container is much better for preserving metadata than on a local filesystem itself but of course depends on the use-case.

PR incoming.

yunzheng:
# Add tar command to target-shell

This PR adds a `tar` command to `target-shell`. It can be used to extract directories and files, preserving permissions, timestamps and other attributes from a target to the local filesystem.

I've been thinking of adding a `--tar` option to the current `save` command, but opted for a separate command instead.

Note that the tar spec itself doesn't preserve atime, ctime and btime, only mtime. But there are alternative tar implementations that can read it from pax_headers. This PR also adds these headers but it will not always be easy to read them using default tools.

closes #1690 

## Proposed Changes

This adds a `tar` command, mimicking the basic flags of the tar command on linux on how to create tar files:

```
trixie:/$ tar --help
usage: tar  [-h] [-f FILE] [-z] [-j] -c [-L] [-v] path [path ...]

Archive one or more files or directories into a tar archive.

By default, the archive is written to stdout, allowing it to be piped to other commands or redirected to a file.
To write to a file directly, use the -f option with the desired filename.

positional arguments:
  path

options:
  -h, --help         show this help message and exit
  -f, --file FILE    file to write the tar archive to, defaults to stdout
  -z, --gzip         compress the tar archive with gzip
  -j, --bzip2        compress the tar archive with bzip2
  -J, --xz           compress the tar archive with xz
  -c, --create       create a tar archive (only this mode is supported)
  -L, --dereference  follow symlinks and archive the files they point to instead of the symlinks themselves
  -v, --verbose
```

The only thing that is not preserved yet is the username and group for the gid and uid if there is a /etc/passwd or /etc/group file. Checking if these files exist and parsing them is not a problem, but maybe a generic approach could be implemented so the functionality can be reused. Suggestions are welcome.

## How to test

(in target-shell)

Create a tar archive and output to stdout, piping it to a local `tar` command to show the listing:

```
trixie:/$ tar -cvf- /etc/ | tar -tf -
```

Create a gzipped tar archive to /tmp/out.tar.gz:

```
trixie:/$ tar -czvf /tmp/out.tar.gz /etc
```

Note: the `-c` flag is mandatory and by default it outputs to stdout, at minimum you can use:

```
trixie:/$ tar -c /etc /home > /tmp/etc-home.tar
```

## Checklist

### DIFF
--- a/tests/tools/test_shell.py
+++ b/tests/tools/test_shell.py
@@ -17,7 +17,6 @@ from unittest.mock import MagicMock, call, mock_open, patch
 import pytest
 
 from dissect.target.containers.raw import RawContainer
-from dissect.target import Target
 from dissect.target.helpers.fsutil import TargetPath, normalize
 from dissect.target.target import Target
 from dissect.target.tools.shell import (


dissect/target/tools/shell.py
```diff
@@ -991,6 +994,164 @@ def cmd_file(self, args: argparse.Namespace, stdout: TextIO) -> bool:
...
+                tarinfo.size = statres.st_size
+            else:
+                tarinfo.size = 0
+            tarinfo.mtime = statres.st_mtime
+            tarinfo.type = type
+            tarinfo.linkname = linkname
+
+            # Store additional timestamps in PAX headers, as the standard tar format only supports mtime.
+            tarinfo.pax_headers["atime"] = str(statres.st_atime)
+            tarinfo.pax_headers["ctime"] = str(statres.st_ctime)
```
Schamper, line 1084: I don't think our own `TarFilesystem` even interprets these fields currently 😄 could you create an issue for that?

Schamper (approved):
Lovely PR. I think the tar writing logic in acquire could really benefit from some of this. For now I'm fine with keeping it within `target-shell`, but if you agree, we can open an issue/explore if we can generalize some of this logic and re-use it in acquire.

dissect/target/tools/shell.py
yunzheng: Done: #1714

yunzheng:
> Lovely PR. I think the tar writing logic in acquire could really benefit from some of this. For now I'm fine with keeping it within `target-shell`, but if you agree, we can open an issue/explore if we can generalize some of this logic and re-use it in acquire.

Thanks, yes I did not really like to inline those tar functions in the `cmd_tar` itself. Would be indeed nice to move this elsewhere for reusability, readability and maintainability.

I think you know best where to open this issue and where it's applicable, so feel free to do so.

Schamper:
> I think you know best where to open this issue and where it's applicable, so feel free to do so.

I think we currently have some CPIO stuff in `dissect.util`, and `vmtar` in `dissect.hypervisor`... and we have a project called `dissect.archive`. Might be nice to put all of those in one place, and dissect.archive sounds like it might be the most suitable?

yunzheng:
> > I think you know best where to open this issue and where it's applicable, so feel free to do so.
> 
> I think we currently have some CPIO stuff in `dissect.util`, and `vmtar` in `dissect.hypervisor`... and we have a project called `dissect.archive`. Might be nice to put all of those in one place, and dissect.archive sounds like it might be the most suitable?

Sounds reasonable to me.

PR #1715, opened 2026-04-10
Add lcd, lpwd and shell commands to target-shell

### LINKED ISSUE #1606: Add command to change local working directory in `target-shell`
Currently there is no way to change the local working directory when you are in `target-shell`, but can be useful if you want to export something using `save` to the current working directory or maybe some other commands.

A good way to implement this is to introduce two new commands (`sftp` also does this for example):

`lcd` -> local change directory
`lpwd` -> local print working directory

Or we could also think of adding `do_shell` which would [automatically](https://docs.python.org/3/search.html?q=do_shell) map `!` to this:

`!cd <directory>` -> shell out and change directory using `cd`
`!pwd` -> shell out and run `pwd`
`!ls -la` -> shell out and run `ls -la`.
`!<omg code exec>` -> shell out and run any arbitrary code..

I think adding `do_shell` can introduce some security risks (eg: clipboard paste oopsies) so maybe better not. But just adding it here for discussion.

yunzheng:
# Add lcd, lpwd and shell commands to target-shell

This PR adds the following new commands to `target-shell`:

 * `lcd <path>` -> change local working directory, also handles `-` for previous local directory or '~' for home directory.
 * `lpwd` -> prints the current local working directory.
 * `shell <cmd>` -> execute a local command, eg: `shell mkdir /tmp/test` to create a local directory.
 * `!<cmd>` -> shorthand for `shell` (automatically handled by Python's Cmd)

resolves #1606

## Proposed Changes

The changes a trivial and by adding the `do_shell` method, we automatically get `!<cmd>` for free.

However, we also want to catch `!cd /tmp` to not spawn a new process but handle this locally for the current process. This is implemented in `do_shell`.

It also tracks previous working directory so `lcd -` can be used.
It also expands "~" and environment variables (eg: $HOME), but there is no test for this.

You could say that `lcd` and `lpwd` are not necessary because we can do `!cd <path>` and `!pwd` respectively, but think it's still nice to have them available.

## Checklist

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -494,7 +494,9 @@ class ExtendedCmd(cmd.Cmd):
             line = os.path.expandvars(line)
             os.chdir(line)
             print("Local directory changed to", Path.cwd())
-            self._local_prev_dir = str(prev)  # only update after successful chdir
+            # only update after successful chdir and only if it's a different directory
+            if prev != Path.cwd():
+                self._local_prev_dir = str(prev)  
         except FileNotFoundError:
             print(f"cd: no such file or directory: {line}")
         except PermissionError:
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -9,7 +9,6 @@ import io
 import itertools
 import logging
 import os
-import pathlib
 import platform
 import pydoc
 import random
@@ -206,7 +205,7 @@ class ExtendedCmd(cmd.Cmd):
 
         return object.__getattribute__(self, attr)
 
-    def _load_targetrc(self, path: pathlib.Path) -> None:
+    def _load_targetrc(self, path: Path) -> None:
         """Load and execute commands from the run commands file."""
         try:
             with path.open() as fh:
@@ -219,9 +218,9 @@ class ExtendedCmd(cmd.Cmd):
         except Exception as e:
             log.debug("Error processing .targetrc file: %s", e)
 
-    def _get_targetrc_path(self) -> pathlib.Path | None:
+    def _get_targetrc_path(self) -> Path | None:
         """Get the path to the run commands file. Can return ``None`` if ``DEFAULT_RUNCOMMANDS_FILE`` is not set."""
-        return pathlib.Path(self.DEFAULT_RUNCOMMANDS_FILE).expanduser() if self.DEFAULT_RUNCOMMANDS_FILE else None
+        return Path(self.DEFAULT_RUNCOMMANDS_FILE).expanduser() if self.DEFAULT_RUNCOMMANDS_FILE else None
 
     def preloop(self) -> None:
         super().preloop()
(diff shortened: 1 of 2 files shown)


dissect/target/tools/shell.py
```diff
@@ -466,6 +468,45 @@ def do_debug(self, line: str) -> bool:
...
+            subprocess.run(line, shell=True, check=False)
+        return False
+
+    def do_lcd(self, line: str) -> bool:
+        """Change the local working directory. Usage: lcd <path>."""
+        if line == "-":
+            if self._local_prev_dir is None:
+                print("cd: no previous directory")
+                return False
+            line = self._local_prev_dir
```
Schamper, line 488: Is this `lftp` behavior? TIL. Wouldn't you go in a loop though? Wouldn't a `popd`/`pushd` behavior make more sense?

dissect/target/tools/shell.py
```diff
@@ -466,6 +468,45 @@ def do_debug(self, line: str) -> bool:
 
         return False
 
+    def do_shell(self, line: str) -> bool:
+        """Execute a local shell command. Usage: !<command>."""
+        parts = line.strip().split(maxsplit=1)
+
+        # Handle `cd` as a special case, as it needs to change the state of our current process.
+        if parts and parts[0] == "cd":
+            target = parts[1].strip() if len(parts) > 1 else pathlib.Path.home()
```
Schamper: I notice you use `pathlib.Path` and `Path` interchangeably, and I just noticed we have both imported. Not sure exactly why, but could you make it consistent (at least within this PR, but preferably in the entire file).

Schamper (changes requested):

dissect/target/tools/shell.py
yunzheng: `cd -` is change to previous directory. It's common and works in bash or whatever shell as well.

it has no loop, it only keeps track of the previous one so no need to keep a stack. If you `cd` to the same directory twice it probably would, I could make an exception for that.

dissect/target/tools/shell.py
yunzheng: done

dissect/target/tools/shell.py
yunzheng: I made an exception if you `lcd <samepath>` twice.

Schamper: > cd - is change to previous directory. It's common and works in bash or whatever shell as well

TIL

yunzheng:
I was planning to add local path completion to `lcd` but then I ran into issues, which resulted into me deep diving this a bit more and came to the conclusion of #1720. So I will probably add local path completion in a future PR.

Schamper:
@yunzheng do you want to merge https://github.com/fox-it/dissect.target/pull/1726 before continuing this one?

yunzheng:
> @yunzheng do you want to merge #1726 before continuing this one?

I consider this PR in a finished state. So you can review and merge this one first and then I can add the path completion via #1726 easily.

Schamper (approved):

yunzheng:
i'm not sure where this CI failure is coming from suddenly:

```
FAILED tests/plugins/os/windows/log/test_evt.py::test_evt_scraping - assert 0 == 5
```

I'm checking if it's caused by this PR or something else.

UPDATE: I cannot seem to reproduce this failure using a ubuntu docker image :(

Schamper:
It was because of https://github.com/fox-it/dissect.eventlog/pull/32. I restarted the jobs.

PR #1719, opened 2026-04-13
Add macOS install history plugin

MrJayTechie:
Add a macOS plugin that parses `/Library/Receipts/InstallHistory.plist`, the system-level install log written by `installd` / `softwareupdated` whenever an install, OS update, or configuration-data delivery (XProtect, MRT, Gatekeeper) completes.

The plugin emits one `InstallHistoryRecord` per install event with fields for timestamp, display name, version, process, content type, and package identifiers.

Implementation notes:
- Uses stdlib `plistlib` to parse the plist; no new helpers added.
- Follows the `macos/user.py` pattern: bare `Plugin` base, `UnsupportedPluginError` on missing path, `TargetRecordDescriptor` with the `macos/install_history` namespace.
- Test fixture is a real `InstallHistory.plist` trimmed to 6 Apple-system-only entries (`macOS 15.5`, `XProtectPlistConfigData`, `MRTConfigData`, `Gatekeeper Compatibility Data`, `Command Line Tools for Xcode 26.0`, `Safari 26.0.1`) to exclude any third-party software.

### DIFF
--- a/dissect/target/plugins/os/unix/bsd/darwin/macos/installhistory.py
+++ b/dissect/target/plugins/os/unix/bsd/darwin/macos/installhistory.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+import plistlib
+from typing import TYPE_CHECKING
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import TargetRecordDescriptor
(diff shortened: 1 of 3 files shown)


Schamper (changes requested):
Can you share the prompts used to generate this code?

PR #1726, opened 2026-04-20
Add prompt_toolkit support to TargetCmd

### LINKED ISSUE #1720: Switch to prompt_toolkit for target-shell
I've come to the conclusion that if we switch to [prompt_toolkit](https://python-prompt-toolkit.readthedocs.io/) for our Cmd prompt in `target-shell` that it would solve some of our outstanding issues and annoyances out of the box and make the overall UX more pleasant as a bonus.

For instance, switching would already fix:

* #1587  (would be fixed immediately)
* #1621 (not directly, but easier to fix)

It would also remove some unnecessary hacks like our `readline_escape()` for printing ANSI chars that somehow were required.

We already install `prompt_toolkit` via our (optional) ipython dependency (`full` extras). If we decide to switch to `prompt_toolkit` it needs to be added to our main dependencies. Currently the third-party dependencies for `dissect.target` are `defusedxml` and `structlog`.

Also, `prompt_toolkit` is Pure Python; no C extensions, the bloodline is pure.

Implementation wise it's relatively easy to hook into our current Cmd loop, just figuring out some compatibility issues with our completer functions. I will try to draft an initial PR.

yunzheng:
# Add prompt_toolkit support to TargetCmd

This PR adds support for `prompt_toolkit` as an alternative to the current `readline` shell in `target-shell`. 

It fixes some outstanding issues, in particular it's now possible to autocomplete `TargetPath` and local `Path` objects really smooth now, and supports spaces and quotes in path names. I've ensured that most edge cases work for autocompletion and is also reflected in the unit tests.

closes #1720

## Proposed Changes

As discussed with @Schamper offline, i've made the `prompt_toolkit` still optional and not a mandatory dependency. But it's likely already available when using it via the `dissect` meta package. 

If `prompt_toolkit` is not installed you still get the old readline shell (and with all the current issues).

For reviewing, the changes can be divided into the following pieces:

* `helpers/completer.py` is added, which contains a custom `QuotedPathCompleter` prompt_toolkit completer class and functions for parsing quoted strings.
* `tests/helpers/test_completer.py` tests several edge cases for autocompleting paths regarding quoting, spaces, etc.
* `shell.py` changes:
  * A `cmdloop_prompt_toolkit()` function is added to `ExtendedCmd`, this is swapped in when initializing the class. An environment variable called `NO_PROMPT_TOOLKIT` can be used to not load it, of course if `prompt_toolkit` is not available it will not be loaded either. The environment variable was mainly added to keep some existing tests working as is.
  * `@arg` options for path objects are now annotated with a `TargetPathArgument` or `LocalPathArgument` argparse action, This is mainly used for type checking in `TargetCmdCompleter` class to parse the arguments of a function to determine the context the user is in for autocompleting.
* a `ReadlineCompatHistory` is added so we can keep using the original readline history file. If readline is not available it will fall back to prompt_toolkit’s history format. 
* The `TargetCmdCompleter` is the main completer, but delegates to other completers depending on the context.

How to test, I usually test with:

```shell
$ uv run -p 3.13 target-shell -q tests/_data/tools/shell/unicode.tar

# Test autocompleting of the positional arguments (TargetPath)
unicode.tar:/$ tar <TAB><TAB><TAB>

# And try the same for local Path
unicode.tar:/$ tar unicode/charsets/hello/ -cvf /tm<TAB>
```

## Checklist

- [X] PR Title is descriptive of the changes
- [X] The description is descriptive of the changes
- [X] Closes `related issue number`


yunzheng:
Will look into the failing Windows tests later on a Windows machine.

yunzheng:
Windows tests now pass, however on 3.13 it fails on:

```
=========================== short test summary info ============================
FAILED tests/filesystems/test_vmtar.py::test_filesystems_vmtar - AttributeError: 'VisorTarInfo' object has no attribute 'is_visor'
FAILED tests/filesystems/test_vmtar.py::test_filesystems_vmtar_zstd - AttributeError: 'VisorTarInfo' object has no attribute 'is_visor'
```

But this is most likely due to changes somewhere else and not related to this PR.

JSCU-CNI:
@yunzheng This is due to changes in CPython 3.13.13, I think https://github.com/fox-it/dissect.hypervisor/pull/84 should fix this.

dissect/target/helpers/completer.py
```diff
@@ -0,0 +1,299 @@
...
+        only_directories: bool = False,
+        file_filter: Callable[[str], bool] | None = None,
+        expanduser: bool = True,
+    ) -> None:
+        self.cli = cli
+        self.target = target
+        self.only_directories = only_directories
+        self.file_filter = file_filter or (lambda _: True)
+        self.expanduser = expanduser
+        self.path_func = self.target.fs.path if self.target else Path
```
Schamper:
```suggestion
        self._path = self.target.fs.path if self.target else Path
```
Or `_create_path` or something. Nitpicky.

dissect/target/tools/shell.py
```diff
@@ -27,6 +27,40 @@
...
+    from prompt_toolkit.formatted_text import ANSI
+    from prompt_toolkit.history import FileHistory
+    from prompt_toolkit.key_binding import KeyBindings
+    from prompt_toolkit.keys import Keys
+    from prompt_toolkit.shortcuts import CompleteStyle
+
+    if TYPE_CHECKING:
+        from collections.abc import Iterable
+
+        from prompt_toolkit.completion import CompleteEvent, Completion
```
Schamper: I think these can be moved to the existing `TYPE_CHECKING` block as import errors are always ignored for type checking.

dissect/target/tools/shell.py
```diff
@@ -501,8 +550,52 @@ def __init__(self, target: Target):
         elif getattr(target._config, "NO_COLOR", None) or os.getenv("NO_COLOR"):
             self.prompt_ps1 = "{base}:{cwd}$ "
 
+        # Enable if prompt_toolkit is available and not disabled via environment variable
+        if HAS_PROMPT_TOOLKIT and os.getenv("NO_PROMPT_TOOLKIT", "0") not in {"1", "true", "True", "TRUE"}:
```
Schamper:
```suggestion
        if HAS_PROMPT_TOOLKIT and os.getenv("NO_PROMPT_TOOLKIT", "0").lower() not in {"1", "true"}:
```

dissect/target/tools/shell.py
```diff
@@ -501,8 +550,52 @@ def __init__(self, target: Target):
...
+
+        @key_bindings.add(Keys.Enter, filter=completion_is_selected)
+        def enter_confirms_completion(event: CompleteEvent) -> None:
+            """When a completion is selected, Enter inserts it, but doesn't submit."""
+            event.current_buffer.complete_state = None  # close the dropdown
+
+        self.preloop()
+        completer = TargetCmdCompleter(self)
+        session = PromptSession(
+            history=FileHistory(f"{self.histfile}_pt"),
```
Schamper: Are the history formats incompatible? Would be nice if we could somehow preserve it.

Schamper (changes requested):

dissect/target/tools/shell.py
yunzheng: incompatible yes. It still preserves it, but just in two different files :)

* ~/.target_history
* ~/.target_history_pt

Schamper: Ah with preserve I meant, it's cross compatible. But I assume that'll be quite a lot of work?

dissect/target/tools/shell.py
yunzheng: done

dissect/target/helpers/completer.py
yunzheng: changed to a _make_path() function.

dissect/target/tools/shell.py
yunzheng: done

dissect/target/tools/shell.py
yunzheng: Aight, made it compatible by adding a custom `ReadlineCompatHistory` class, this will read and write using readline library under the hood. If readline is not available it will fallback to `FileHistory`.

dissect/target/tools/utils/completer.py
Schamper, line 1: For nitpicking sake, would you say this belongs more here or at `dissect/target/tools/utils`? Since this feels like it's just a helper for tools, not anything within the core framework itself.

Same would go for the test file, if you agree.

Schamper (changes requested):

dissect/target/tools/utils/completer.py
yunzheng: I saw it as a helper code for autocompletion. So placed in the `helpers` directory. I didn't find it to be a tool itself, so I skipped the `tools` directory.

`tools/utils` is a bit ambiguous, but i'm happy to put it here. I have no strong opinion or objection.

Give a thumbs up and I will move it.

Schamper: Yeah `tools/utils` could admittedly be named better, but the purpose of it is "helpers specific to tools". And I think this completer code will only ever be used as a helper for "tools". The plain `helpers` directory is (aims to be) more for helpers that relate to interacting with targets or data for parsing.

Having some separation between the two would be nice, I think. But any thoughts you have could be insightful.

yunzheng: Looking at the folder structure again, and I agree the most logical place would be to put it under the `tools` folder as it mainly affects `tools/shell.py`.

There is a `tools/autocompletion` directory, but that is not python related. A `tools/completer/` would make more sense if the code was more reusable but that is not the case. So to avoid creating even more directories let's just put it in `tools/utils`.

yunzheng: i like to move it, move it

Schamper: Not for this PR but for future consideration, would renaming `tools/utils` to `tools/helpers` make more sense?

Schamper (approved):

dissect/target/tools/utils/completer.py
yunzheng: Renaming it to `helpers` would just be another similar name. Maybe `tools/lib`? 

But honestly I think it's fine, my main problem was that I was looking at it from a single parent directory way, when you look at it from a namespace/nested way then I think makes sense to have sub directory under `tools` that is still related to tools but it's there not to make the parent directory cluttered.

PR #1717, opened 2026-04-12
Add 61 macOS forensic plugins (195 functions)

MrJayTechie:
## Summary

I've been working on macOS forensic plugins for dissect and built 61 plugins covering 195 functions across all major macOS artifacts. Right now dissect.target only has 3 macOS plugins (user, network, _os) — this would bring macOS coverage much closer to what's available for Windows and Linux.

## What's included

| Category | Plugins | Functions |
|----------|---------|-----------|
| Browsers | safari, chromium, firefox, cookies | 21 |
| User Activity | knowledgec, biome, screentime, interactions, spotlight, launchpad | 54 |
| Communications | imessage, callhistory, facetime, addressbook, notifications, notes | 15 |
| Persistence | autostart, kext, execpolicy, applications | 16 |
| Security | tcc, firewall, keychain, profiles | 15 |
| Logs | logs, crashreporter, powerlogs | 16 |
| File System | fsevents, dsstore, docrevisions, trash, quicklook | 8 |
| Network | wifiintelligence, ssh, ard, msrdc, screensharing | 9 |
| System/Other | preferences, osinfo, users, accounts, icloud, wallet, and more | 41 |

## Key points

- macOS 13, 14, 15 compatible (including Sequoia path changes)
- All SQLite plugins handle WAL/SHM sidecars
- Handles Velociraptor `%2E` filename encoding
- Ruff clean, type hints and docstrings on all exports
- Tested with 0 errors across multiple collections

## Testing

Tested against real macOS collections — 135/195 functions return data on a typical Mac. Remaining are legitimate empties (app not installed, feature disabled). Happy to add pytest tests based on your framework.

PR #1727, opened 2026-04-21
Add macOS plugins

EmperorValkorion:
Added various macOS plugins.

- Parsers for install, system, wifi, fsck_apfs, asl logs and fs events

- Locale parsers for timezone, language, install date & location services 

- Plist parsers for: LaunchAgents, LaunchDaemons, Login Items, Airport Preferences, Code Signature Coderesources, Contents Info, Contents Version, Global User Preferences, Groups, Installation History, Keyboard Layout, Login Window, Resources Info Strings, User Password Hashes (Shadow), Software Update Preferences, System Preferences, Time Machine, Safari Per Site Zoom Preferences, Safari Downloads, Safari Recently Closed Tabs & Safari User Notification Permissions

- SQLite parsers for: Authorization Rules, Call History, Directory Services Local Nodes, Duet Activity Scheduler, Duet KnowledgeC, Duet InteractionC, Gatekeeper Opaque, Identity Services, Keychain, Notes, TCC, Text Replacements, User Accounts, Safari Favicons & Safari Per Site Preferences

- Parsers for macOS cron tabs, at jobs & periodic functions

- Added parser for Resources Localizable Strings, but it still has some issues (new records do not always begin on a new line). Also have not written tests for this artifact yet. Not sure how valuable the artifact is considering it appears to just contain a list of translations.

- Added build_records.py helper file to build SQLite and Plist records and match them to the provided record descriptors

This PR was initially based on #1637

ASL & fs events plugins are based on #1717 

This PR includes #1753


dissect/target/plugins/os/unix/bsd/darwin/macos/logs/fsck_apfs_log.py
```diff
@@ -0,0 +1,68 @@
...
+    def check_compatible(self) -> None:
+        if not self.target.fs.exists(self.FSCK_APFS_LOG_PATH):
+            raise UnsupportedPluginError("No fsck_apfs.log file found.")
+
+    @export(record=FsckAPFSLogRecord)
+    def fsck_apfs_log(self) -> Iterator[FsckAPFSLogRecord]:
+        """Return all fsck_apfs log messages."""
+        logfile = self.target.fs.path(self.FSCK_APFS_LOG_PATH).open(mode="rt")
+
+        for line in logfile.readlines():
```
Schamper: General tip for reading from text files:

```python
with self.target.fs.path(self.FSCK_APFS_LOG_PATH).open(mode="rt") as fh:
    for line in fh:
        ...
```
The context manager is not _strictly_ necessary, but we do try to be a little bit more conscious of our dangling open files these days. The second, `for line in fh` causes "lazy" line iteration over the opened file. Using `.readlines()` eagerly reads all lines into memory at once. Iterating over `fh` reads them one by one. This reduces memory usage when reading large files.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/fsck_apfs_log.py
```diff
@@ -0,0 +1,68 @@
...
+    from collections.abc import Iterator
+
+    from dissect.target.target import Target
+
+FsckAPFSLogRecord = TargetRecordDescriptor(
+    "macos/fsck_apfs_log",
+    [
+        ("string", "disk_path"),
+        ("string", "message"),
+        ("datetime", "ts"),
```
Schamper: We generally like to put timestamp fields first, so that sorting output naturally sorts records into chronological order.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/install_log.py
```diff
@@ -0,0 +1,111 @@
...
+    """Return information related software installations and updates on macOS.
+
+    References:
+        - https://sansorg.egnyte.com/dl/m9ftGF7heI
+    """
+
+    INSTALL_LOG_PATH = "/var/log/install.log"
+
+    def __init__(self, target: Target):
+        super().__init__(target)
```
Schamper: If you're not doing anything special in `__init__`, you can omit it.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/install_log.py
```diff
@@ -0,0 +1,111 @@
...
+                if current_ts:
+                    asdf = current_buf[len(current_ts.group()) + 1 :]
+
+                    parts = asdf.split(" ", 2)
+
+                    if len(parts) == 3:
+                        hostname, component, message = parts
+                    elif len(parts) == 2:
+                        hostname, message = parts
+                        component = None
```
Schamper: When you're doing some "non-standard" handling of lines as you do here, it can be nice to either shortly describe in a code comment (`#`) with maybe a 2-3 line example of what the pattern is that you're dealing with here. Saves someone from having to look up the test file.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/install_log.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        logfile = self.target.fs.path(self.INSTALL_LOG_PATH).open(mode="rt")
+
+        current_ts: re.Match[str] | None = None
+        current_buf = ""
+
+        for line in logfile.readlines():
+            if ts_match := RE_TIMESTAMP_PATTERN.match(line):
+                if current_ts:
+                    asdf = current_buf[len(current_ts.group()) + 1 :]
```
Schamper: nice.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/system_log.py
```diff
@@ -0,0 +1,65 @@
...
+from dissect.target.plugins.os.unix.log.helpers import RE_TS
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from dissect.target.target import Target
+
+SystemLogRecord = TargetRecordDescriptor(
+    "macos/system_log",
+    [("datetime", "ts"), ("string", "host"), ("string", "component"), ("string", "message"), ("path", "source")],
```
Schamper: General style comment, if you add a trailing comma like so
```suggestion
    [("datetime", "ts"), ("string", "host"), ("string", "component"), ("string", "message"), ("path", "source"),],
```
The linter will automatically "pretty format" it.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/system_log.py
```diff
@@ -0,0 +1,65 @@
...
+
+class SystemLogPlugin(Plugin):
+    """Return system logs on macOS."""
+
+    SYSTEM_LOG_GLOB = "/var/log/system.log*"
+
+    def __init__(self, target: Target):
+        super().__init__(target)
+        self.log_files = set()
+        self._resolve_files()
```
Schamper:
```suggestion
        self.log_files = self._resolve_files()
```
It's better to use a structure like this. The `_resolve_files` method is now stateful for no real reason, while it could just as easily just return a set directly.

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/system_log.py
```diff
@@ -0,0 +1,65 @@
...
+    @export(record=SystemLogRecord)
+    def system_log(self) -> Iterator[SystemLogRecord]:
+        """Return all macOS system log messages."""
+        for file in self.log_files:
+            filepath = self.target.fs.path(file)
+
+            current_buf = ""
+
+            for ts, line in year_rollover_helper(filepath, RE_TS, "%b %d %H:%M:%S", timezone.utc):
+                current_buf = line + "\n\t" + current_buf
```
Schamper: Small nitpick and Python speed hack, instead of using `+` to build up a string, it's actually faster to append to a list, and then in the end join them together like so `"\n\t".join(lines)`.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
Schamper, line 1: Just a project style thing, but we don't necessarily categorize things under terms like "persistence". You could move them one directory up.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
```diff
@@ -0,0 +1,104 @@
...
+            queue = name[0]
+            seq = int(name[1:6], 16)
+            time_hex = name[6:]
+
+            execution_time = None
+            try:
+                minutes = int(time_hex, 16)
+                execution_time = minutes * 60
+            except ValueError:
+                pass
```
Schamper, line 99: Similarly here, a small comment would be nice with an example file name.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
```diff
@@ -0,0 +1,104 @@
...
+    def _find_files(self) -> None:
+        for pattern in self.PATHS:
+            for path in self.target.fs.glob(pattern):
+                self.at_jobs_files.add(path)
+
+    @export(record=AtJobsRecord)
+    def at_jobs(self) -> Iterator[AtJobsRecord]:
+        """Yield macOS at jobs."""
+        for file in self.at_jobs_files:
+            name = Path(file).name
```
Schamper:
```suggestion
            name = file.name
```
`file` is already a `Path`-like object, so you should be able to just do `file.name`.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
```diff
@@ -0,0 +1,104 @@
...
+                    else:
+                        line = line.strip()
+
+                        if not line or line.startswith(("#", "export")):
+                            continue
+
+                        line = line.split("#", 1)[0].strip()
+
+                        if "export OLDPWD" in line:
+                            command_line = True
```
Schamper, line 116: Add some comment here about the assumption that you make here.

dissect/target/plugins/os/unix/bsd/darwin/macos/persistence/at_jobs.py
```diff
@@ -0,0 +1,104 @@
...
+        ("string", "command"),
+        ("path", "source"),
+    ],
+)
+
+FIELD_MAPPINGS = {
+    "backgroundAppRefreshLoadCount": "background_app_refresh_load_count",
+    "launchServicesItemsImported": "launch_services_items_imported",
+    "serviceManagementLoginItemsMigrated": "service_management_login_items_migrated",
+}
```
Schamper: Unused?

dissect/target/plugins/os/unix/bsd/darwin/macos/airport_preferences.py
```diff
@@ -0,0 +1,63 @@
...
+        preferred_order = plist.get("PreferredOrder")
+
+        yield AirportPreferencesRecord(
+            counter=counter,
+            device_uuid=device_uuid,
+            version=version,
+            preferred_order=preferred_order,
+            source=self.file,
+            _target=self.target,
+        )
```
Schamper, line 69:
```suggestion
        with self.file.open() as fh:
            plist = plistlib.load(fh)

        yield AirportPreferencesRecord(
            counter=plist.get("Counter"),
            device_uuid=plist.get("DeviceUUID"),
            version=plist.get("Version"),
            preferred_order=plist.get("PreferredOrder"),
            source=self.file,
            _target=self.target,
        )
```
Could be shortened like so.

dissect/target/plugins/os/unix/bsd/darwin/macos/code_signature_coderesources.py
```diff
@@ -0,0 +1,89 @@
...
+    NestedRecord,
+    CDHashRecord,
+)
+
+
+class CodeSignatureCodeResourcesPlugin(Plugin):
+    """macOS Code signature CodeResources plugin."""
+
+    PATHS = (
+        "/Applications/Utilities/*.app/Contents/_CodeSignature/CodeResources",
```
Schamper: Isn't this something that could in in any `.app`? So also anything in e.g. `/Applications/*.app`?

dissect/target/plugins/os/unix/bsd/darwin/macos/contents_info.py
```diff
@@ -0,0 +1,55 @@
...
+from dissect.target.helpers.record import DynamicDescriptor
+from dissect.target.plugin import Plugin, export
+from dissect.target.plugins.os.unix.bsd.darwin.macos.helpers.build_records import build_plist_records
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from dissect.target.target import Target
+
+re_illegal_characters = re.compile(r"[\(\): \.\-#\/\>\<]")
```
Schamper: Unused?

dissect/target/plugins/os/unix/bsd/darwin/macos/contents_info.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class ContentsInfoPlugin(Plugin):
+    """macOS contents info plugin."""
+
+    PATHS = (
+        "/Applications/*/*.app/Contents/Info.plist",
```
Schamper: What about apps in `/Applications/*.app`?

Might be nice to add a utility for iterating applications and kexts. It could yield paths for every `.app` and `.kext` it finds (within known locations), as well as perform some basic nested plugin/bundle checks for each one. That way, you don't end up with an evergrowing list of glob patterns scattered over different plugins. Plugins that act on these packages would then just look into `path / "Contents/Info.plist"` for every path the helper returns.

I.e. search within certain paths for `*.app`, `.kext`, `.framework` (`/Applications`, `/System/Library/CoreServices`, `/System//Library/Extensions` etc), then for each of those paths, recursively check for `Contents/PlugIns` and `Contents/Resources` and additional apps, kexts, frameworks, plugins and bundles in here.

A `applications` plugin could utilize this information too (as we also have them for Linux and Windows).

Some suggestions:
- I forget the exact nomenclature, but I believe Apple generally calls these "bundles", so you could make a `bundles.py`
- Within there, you could add some exports (or `@internal`, which are plugin functions that are solely meant to be used by other plugins, not tools like `target-query`) for some of the individual types of bundles, and something to iterate over them all
- This plugin and e.g. the code signature one would simple iterate over the results of that plugin

dissect/target/plugins/os/unix/bsd/darwin/macos/directory_services_local_nodes.py
```diff
@@ -0,0 +1,111 @@
...
+                "smb_rid",
+                "groupmembers",
+                "users",
+                "nestedgroups",
+            }
+
+            r_rows: list[tuple[str, object]] = []
+            n_tables = set()
+
+            for table in database.tables():
```
Schamper, line 90: A little bit of documentation on what this database looks like would help.

Schamper (changes requested):

dissect/target/plugins/os/unix/bsd/darwin/macos/logs/install_log.py
EmperorValkorion: I have added some comments for install_log. Still need to go through all the other plugins and helpers to see where else this applies.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
EmperorValkorion: file is a string, so doing file.name without Path() causes an AttributeError

dissect/target/plugins/os/unix/bsd/darwin/macos/contents_info.py
EmperorValkorion: I have added a find_bundle_files helper function in macos/helpers/build_paths.py, but I'm not quite satisfied with the performance yet.

dissect/target/plugins/os/unix/bsd/darwin/macos/at_jobs.py
EmperorValkorion: Replaced Path() with self.target.fs.path()

PR #1732, opened 2026-04-22
Shorten test names and prevent unintentional duplicate test names

JSCU-CNI:
This PR shortens long test names and their parameters using `pytest.param`. We also introduce a pytest `hookimpl` to catch any unintentional duplicate test names. 

This should make reading and debugging pytest test output easier in dissect.


Schamper:
The `pytest.param` I get, but can you give a more concrete example of what the renaming of some duplicate test names accomplishes? I just figured it's a matter of style, but I considered each test file to be its own "unit", and the test function name makes sense within the context of that unit. I.e. I found it superfluous to name something `test_xxxx_something` if it's already in a file called `test_xxxx.py`, especially if with a `-v` it would show as `test_xxxx.py::test_xxxx_something`.

In the end it's just test names so it's whatever, I won't die on this hill, but I'm just curious what the problem/workflow you're tackling with this change.

JSCU-CNI:
>  I considered each test file to be its own "unit", and the test function name makes sense within the context of that unit.

I agree, that's also why we added `allowlist` so we can ignore certain common test name patterns across test modules.

> [...] can you give a more concrete example of what the renaming of some duplicate test names accomplishes?

This mainly addresses ~lazy~ developer behavior (i.e. copying tests from `test_foo.py` to `test_bar.py` without renaming the test functions. This can seem innocent at the surface but can quickly become confusing when reading test output. During PR review this is often overlooked and duplicates might not seem apparent in the diff of a PR. See for example the renamed tests in `tests/filesystems/test_ffs.py`, `tests/filesystems/test_lima.py` and `tests/plugins/os/uinix/linux/test_network.py`.

tests/loaders/test_uac.py
```diff
@@ -109,7 +109,7 @@ def test_compressed_zip() -> None:
     assert test_file.open().readline() == b"root:x:0:0:root:/root:/bin/bash\n"
 
 
-def test_dir(mock_uac_dir: Path) -> None:
+def test_dir_uac(mock_uac_dir: Path) -> None:
```
Schamper:
```suggestion
def test_uac_dir(mock_uac_dir: Path) -> None:
```

tests/loaders/test_kape.py
```diff
@@ -51,7 +51,7 @@ def test_target_open(opener: Callable[[str | Path], Target], mock_kape_dir: Path
 
 
 @patch("dissect.target.filesystems.dir.DirectoryFilesystem.ntfs", None, create=True)
-def test_dir(mock_kape_dir: Path) -> None:
+def test_dir_kape(mock_kape_dir: Path) -> None:
```
Schamper:
```suggestion
def test_kape_dir(mock_kape_dir: Path) -> None:
```

tests/helpers/test_magic.py
```diff
@@ -109,12 +109,19 @@
...
         ("from_buffer", b"Received:", True, "message/rfc822"),
         ("from_buffer", b"!BDN", True, "application/vnd.ms-outlook-pst"),
-        ("from_buffer", (b"\x00" * 2112) + b"Microsoft Word document data", True, "application/msword"),
-        (
+        pytest.param(
+            "from_buffer",
+            (b"\x00" * 2112) + b"Microsoft Word document data",
+            True,
+            "application/msword",
+            id="msword",
```
Schamper, line 127: Do none of the others need this? I was thinking maybe a more general `ids` on the parameters that just takes the expected mime output would work better, but then you'd run into issues where you're trying to detect the same mime output. Maybe there's some clever way around that, but right now it feels strange to only have an `id` for a handful of these?

tests/loaders/test_vmsupport.py
```diff
@@ -88,7 +88,7 @@ def test_compressed_tar(data_path: str) -> None:
     assert b"/resourceGroups/version" in test_file.open().read()
 
 
-def test_dir(mock_vmsupport_dir: Path) -> None:
+def test_dir_vmsupport(mock_vmsupport_dir: Path) -> None:
```
Schamper:
```suggestion
def test_vmsupport_dir(mock_vmsupport_dir: Path) -> None:
```

tests/loaders/test_vmsupport.py
```diff
@@ -112,7 +112,7 @@ def test_dir(mock_vmsupport_dir: Path) -> None:
     ],
 )
 @pytest.mark.benchmark
-def test_benchmark(benchmark: BenchmarkFixture, archive: str, loader: type[Loader]) -> None:
+def test_benchmark_vmsupport(benchmark: BenchmarkFixture, archive: str, loader: type[Loader]) -> None:
```
Schamper, line 115: Wouldn't `test_vmsupport_benchmark` be more consistent? Also goes for the other benchmarks.

tests/conftest.py
```diff
@@ -80,6 +81,29 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
+def pytest_pycollect_makemodule(module_path: str, parent: str) -> None:
+    if duplicates := find_duplicate_functions(module_path):
+        pytest.exit(f"duplicate functions found in module: {module_path}: {duplicates}")
+
+
+def find_duplicate_functions(source_path: str) -> set[str]:
+    source_code = pathlib.Path(source_path).read_text()
+    tree = ast.parse(source_code, source_path)
+    duplicates = set()
+    allowlist = ("test_other", "test_all", "test_target_open", "test_loader")
```
Schamper: Why not go all out and rename those too?

tests/conftest.py
```diff
@@ -80,6 +81,29 @@ def pytest_sessionstart(session: pytest.Session) -> None:
                     )
             break
 
+function_names = {}
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_pycollect_makemodule(module_path: str, parent: str) -> None:
+    if duplicates := find_duplicate_functions(module_path):
+        pytest.exit(f"duplicate functions found in module: {module_path}: {duplicates}")
```
Schamper:
```suggestion
        pytest.exit(f"duplicate test function names found in module: {module_path}: {duplicates}")
```

tests/conftest.py
```diff
@@ -80,6 +81,29 @@ def pytest_sessionstart(session: pytest.Session) -> None:
                     )
             break
 
+function_names = {}
```
Schamper:
```suggestion
_function_names = {}
```

Schamper (changes requested):

tests/loaders/test_vmsupport.py
JSCU-CNI: I was under the impression `test_benchmark_*` was the standard prefix for benchmark tests.

Schamper: Me too 😄 idk man. Keep these as-is then. Maybe worth adding a check that `@pytest.mark.benchmark` marked tests start with `test_benchmark_`.

tests/helpers/test_magic.py
```diff
@@ -24,160 +24,195 @@
...
+        pytest.param("from_buffer", b"070701", True, "application/x-cpio", id="cpio-1"),
+        pytest.param("from_buffer", b"070702", True, "application/x-cpio", id="cpio-2"),
+        pytest.param("from_buffer", b"070707", True, "application/x-cpio", id="cpio-3"),
+        pytest.param("from_buffer", bytes.fromhex("504b0304"), True, "application/zip", id="zip-1"),
+        pytest.param("from_buffer", bytes.fromhex("504b0506"), True, "application/zip", id="zip-2"),
+        pytest.param("from_buffer", bytes.fromhex("504b0708"), True, "application/zip", id="zip-3"),
+        pytest.param("from_buffer", bytes.fromhex("526172211a07"), True, "application/vnd.rar", id="rar-1"),
+        pytest.param("from_buffer", bytes.fromhex("526172211a070100"), True, "application/vnd.rar", id="rar-2"),
+        pytest.param("from_buffer", bytes.fromhex("0E031301"), True, "application/x-hdf", id="x-hdf-1"),
+        pytest.param("from_buffer", bytes.fromhex("894844460D0A1A0A"), True, "application/x-hdf", id="x-hdf-2"),
```
Schamper, line 43: Not to be nitpicky, but some you start with `x-` and others you don't? What's the logic there?

tests/loaders/test_ab.py
```diff
@@ -24,7 +24,7 @@
         pytest.param(lambda x: next(Target.open_all([x])), id="target-open-all"),
     ],
 )
-def test_target_open(opener: Callable[[str | Path], Target]) -> None:
+def test_target_open_ab(opener: Callable[[str | Path], Target]) -> None:
```
Schamper, line 27: Since we're here to discuss nitpicks anyway, `test_ab_target_open` or `test_target_open_ab`?

Schamper (changes requested):

JSCU-CNI:
It seems that `pytest_pycollect_makemodule` does not actually collect all test modules recursively but only the eight `test_*.py` files next to `conftest.py`. I will take a look tomorrow if we can use a different pytest hook.

Issue #1571, opened 2026-02-23
Add RST to ANSI converter for function docstrings


JSCU-CNI:
Currently docstring output of `target-query -f function -h` is raw rst. It would be nice if this can be formatted to terminal-friendly ANSI.

Perhaps `docutils.parsers.rst` or (a fork of) the `rst2ansi` project can be adapted (https://github.com/timcera/python-rst2ansi) or used as inspiration.

PR #1733, opened 2026-04-22
Unify target architecture return values

### LINKED ISSUE #931: Unify unix and windows architecture return values
Currently UNIX and Windows' `architecture` property implementations are not unified.

It would be beneficial for the dissect project to unify the return values of this property across all different supported targets. Perhaps we can further look into returning somewhat standardized target triples. For inspiration we can look at GCC, LLVM and Rust.

- https://mcyoung.xyz/2025/04/14/target-triples/
- https://wiki.osdev.org/Target_Triplet
- https://llvm.org/doxygen/classllvm_1_1Triple.html
- https://doc.rust-lang.org/rustc/platform-support.html

As discussed in https://github.com/fox-it/dissect.target/pull/921#discussion_r1824702918.

JSCU-CNI:
Implements and fixes https://github.com/fox-it/dissect.target/issues/931.

### DIFF
--- a/dissect/target/helpers/arch.py
+++ b/dissect/target/helpers/arch.py
@@ -37,13 +37,26 @@ MACHINE_MAP_LINUX = {
 }
 
 
-# Translate a Mach-O instruction CPU sub type to a standardized machine name.
-# https://en.wikipedia.org/wiki/Mach-O
+# Translate a Mach-O instruction CPU type to a standardized machine name.
+# Cherry-picked to only architectures widely distributed by Apple.
+# https://en.wikipedia.org/wiki/Mach-O, /usr/include/mach-o/loader.h & /usr/include/mach-o/machine.h
 MACHINE_MAP_DARWIN = {
-    0x0C000001: "aarch64",  # big endian, x64
-    0x0100000C: "aarch64",  # little endian, x64
-    0x0C000000: "aarch32",  # big endian, x32
-    0x0000000C: "aarch32",  # little endian, x32
+    0x00000006: "mc68k",  # CPU_TYPE_MC680x0
+    0x06000000: "mc68k",
+    0x00000007: "x86",  # CPU_TYPE_X86
+    0x07000000: "x86",
+    0x01000007: "x86_64",  # (CPU_TYPE_X86 | CPU_ARCH_ABI64)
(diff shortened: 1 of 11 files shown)


dissect/target/helpers/arch.py
```diff
@@ -0,0 +1,116 @@
...
+    0x0C000001: "aarch64",
+    0x0200000C: "aarch64",  # CPU_TYPE_ARM64_32 = (CPU_TYPE_ARM | CPU_ARCH_ABI64_32)
+    0x0C000002: "aarch64",
+}
+
+
+# Translate the operating system to it's corresponding vendor name.
+# Linux derivatives generally translate to 'unknown', except Android, which translates to 'linux'.
+OS_TO_VENDOR = {
+    OperatingSystem.WINDOWS.value: "pc",
```
Schamper, line 66: We really ought to start using the enums instead of the string values at some point 😄 not a comment for this PR, just a comment in general.

dissect/target/plugin.py
```diff
@@ -1288,11 +1288,9 @@ def os_tree(self) -> list[str]:
 
     @export(property=True)
     def architecture(self) -> str | None:
-        """Return a slug of the target's OS architecture.
+        """Return a target triple slug of the target's instruction set architecture, vendor and OS.
 
-        Returns:
-            A slug of the OS architecture, e.g. 'x86_32-unix', 'MIPS-linux' or
-            'AMD64-win32', or 'unknown' if the architecture is unknown.
+        See ``target_triple`` in ``dissect.target.helpers.arch`` for implementation details.
```
Schamper:
```suggestion
        See :func:`dissect.target.helpers.arch.target_triple` for implementation details.
```
Maybe consider this? It'll at least render out nicely in the docs. Maybe nicely with https://github.com/fox-it/dissect.target/issues/1571 too?

Schamper (changes requested):

dissect/target/helpers/arch.py
JSCU-CNI: Agreed :smile:

Schamper: https://github.com/fox-it/dissect.target/issues/1735

Schamper (approved):

PR #1736, opened 2026-04-23
Replace `OperatingSystem.*.value` with `OperatingSystem.*`

### LINKED ISSUE #1735: Start using the `OperatingSystem` like an enum instead of a string.
We're using the string value everywhere mostly because of legacy reasons, which causes us to have to do `.value` in a lot of places. We should just start using the actual enum value more.  It's actually a subclass of `StrEnum`, so comparisons should keep working. It's primarily usage as dictionary keys where it's most awkward.

The only place where this might be a bit awkward is in `target-query` or `target-shell` output, open to suggestions on that front. We could override the `__str__`?

JSCU-CNI:
Renames `OperatingSystem.*.value` invocations and s `OperatingSystem` instances instead of raw strings where possible.

Fixes #1735.

### DIFF
--- a/dissect/target/plugins/apps/vpn/openvpn.py
+++ b/dissect/target/plugins/apps/vpn/openvpn.py
@@ -129,8 +129,8 @@ class OpenVPNPlugin(Plugin):
     )
 
     user_config_paths: Final[dict[str, list[str]]] = {
-        OperatingSystem.WINDOWS.value: ["OpenVPN/config/"],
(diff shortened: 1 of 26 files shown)


dissect/target/plugins/apps/vpn/openvpn.py
```diff
@@ -129,8 +129,8 @@ class OpenVPNPlugin(Plugin):
     )
 
     user_config_paths: Final[dict[str, list[str]]] = {
```
Schamper: The typing would need updating too.

Schamper (changes requested):

PR #1739, opened 2026-04-30
Update cstruct.util colours

Miauwkeru:


twiggler (approved):

dissect/target/tools/qfind.py
```diff
@@ -92,7 +92,7 @@ def main() -> int:
                 header = f"[{hit.offset:#08x} @ {hit.needle} ({hit.codec})]"
 
                 if not NO_COLOR:
-                    header = utils.COLOR_WHITE + header + utils.COLOR_NORMAL
+                    header = utils.COLOR_WHITE + header + utils.COLOR_CLEAR_BOLD
```
Schamper, line 95:
```suggestion
                    header = utils.COLOR_WHITE + header + utils.COLOR_CLEAR
```
Maybe just use this one, as it's not personally clear to me why the `_BOLD` variant should be used. I actually remove the BOLD variant in this PR: https://github.com/fox-it/dissect.cstruct/pull/156

Schamper (changes requested):

Schamper:
https://github.com/fox-it/dissect.target/pull/1740 also changes the other colors so maybe let that one be leading?

dissect/target/tools/qfind.py
Miauwkeru: I used the `_BOLD` variant as it had the same value as `COLOR_NORMAL`, so it just maintained what was there before. I have no issue changing it to `COLOR_CLEAR`, shall I just change everything immediately so we don't have a similar issue when that pr gets merged?

Miauwkeru:
> #1740 also changes the other colors so maybe let that one be leading?

Sure, then I'll close this one. I'll request them to update the pyprojects.toml file

PR #1740, opened 2026-04-30
Update cstruct color imports

JSCU-CNI:


dissect/target/tools/qfind.py
```diff
@@ -92,7 +92,7 @@ def main() -> int:
                 header = f"[{hit.offset:#08x} @ {hit.needle} ({hit.codec})]"
 
                 if not NO_COLOR:
-                    header = utils.COLOR_WHITE + header + utils.COLOR_NORMAL
+                    header = utils.COLOR_WHITE_BOLD + header + utils.COLOR_CLEAR_BOLD
```
Schamper:
```suggestion
                    header = utils.COLOR_WHITE_BOLD + header + utils.COLOR_CLEAR
```
Basically https://github.com/fox-it/dissect.target/pull/1739#discussion_r3166663285 unless you can say why `CLEAR_BOLD` is superior 😄.

Schamper (changes requested):

dissect/target/tools/qfind.py
JSCU-CNI: Mainly to preserve previous behavior, unsure who decided on keeping bold active after a FG reset :)

Miauwkeru:
could you also update the `pyproject.toml`?

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 0a69d474..cd2ea0a6 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "defusedxml",
-    "dissect.cstruct>=4,<5",
+    "dissect.cstruct>=4.8.dev7,<5",
     "dissect.database>=1.1,<2",
     "dissect.eventlog>=3,<4",
     "dissect.evidence>=3.13,<4",
@@ -83,7 +83,7 @@ dev = [
     "dissect.cim[dev]>=3.0.dev,<4.0.dev",
     "dissect.clfs[dev]>=1.0.dev,<2.0.dev",
     "dissect.cramfs[dev]>=1.0.dev,<2.0.dev",
-    "dissect.cstruct>=4.0.dev,<5.0.dev",
+    "dissect.cstruct>=4.8.dev7,<5.0.dev",
     "dissect.database[dev]>=1.1,<2.0.dev",
     "dissect.etl[dev]>=3.0.dev,<4.0.dev",
     "dissect.eventlog[dev]>=3.0.dev,<4.0.dev",

```

dissect/target/tools/qfind.py
Schamper: According to Claude:

> These two codes both reset terminal formatting, but they work differently:
>
> \033[0m is the standard "reset all attributes" code. It clears everything — color, bold, underline, italic, etc. — returning the terminal to its default state.
>
> \033[1;0m is a sequence with two parameters: 1 (enable bold) followed by 0 (reset all). However, because ANSI parameters are processed left to right, the 0 reset happens after the 1, so the net effect is also a full reset. The leading 1 is effectively a no-op.
> 
> [...]
> 
> - Use \033[0m — it's the correct, idiomatic reset. It's universally supported, intention is clear, and it's shorter.
> - Avoid \033[1;0m — there's no good reason to use it. If you found this in existing code, it was probably intended to be \033[0;1m (reset then bold, which actually does leave text bold), or just \033[0m.

JSCU-CNI: Reached the same conclusion using https://notes.burke.libbey.me/ansi-escape-codes/ and https://jakob-bagterp.github.io/colorist-for-python/ansi-escape-codes/standard-16-colors/#normal-colors :smile:

Schamper (approved):

PR #1742, opened 2026-04-30
Add filesystem allocation support

JSCU-CNI:
This pull requests adds support for distinguishing which parts of a filesystem are in use and which ones are not. `Filesystem` classes can implement a few functions that yield unallocated or allocated streams, or a tuple containing one stream for each.

The PR includes implementations of these functions for NTFS and EXTFS. They depend on functionality added in https://github.com/fox-it/dissect.ntfs/pull/58 and https://github.com/fox-it/dissect.extfs/pull/52 which themselves depend on https://github.com/fox-it/dissect.util/pull/127 .

We have added one small extension to `target-mount` to expose the different streams when filesystems support it. 

In the future we hope more scraping functionality can use the distinction between the allocated/unallocated space. In particular we were thinking about https://github.com/fox-it/dissect.target/pull/1329. It would be nice to have one function that can assemble streams for scraping while taking some preferences into account (in this case, allocated only, unallocated only, or don't care). `create_streams` seems to be a good candidate.

### DIFF
--- a/dissect/target/filesystem.py
+++ b/dissect/target/filesystem.py
@@ -7,7 +7,10 @@ import pathlib
 import stat
 from collections import defaultdict
 from functools import cache, cached_property
+from itertools import zip_longest
 from typing import TYPE_CHECKING, Any, BinaryIO, Final
+
+from dissect.util.stream import MappingStream
 
 from dissect.target.exceptions import (
     FileNotFoundError,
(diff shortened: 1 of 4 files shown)


JSCU-CNI:
@Schamper and other reviewers, any chance you can fast-track the review of this PR, as well as https://github.com/fox-it/dissect.extfs/pull/52 and https://github.com/fox-it/dissect.ntfs/pull/58? 

The dependency-chain gives us quite a headache and prevents us from building on top of this. If you think it will be a while that's also okay but then we know we'll have to work around it for now.

PR #1747, opened 2026-05-07
Fix stored ZIP random access

### LINKED ISSUE #1746: Stored ZIP entries can return incorrect data during interleaved reads
## Problem

Stored ZIP entries can return incorrect data when multiple entries are
opened and seek/read access is interleaved through the same `ZipFile`
instance.

This can surface with Velociraptor collection ZIPs during registry and
EVTX parsing, because those parsers may keep multiple files open and
seek within them.

## Expected behavior

Opening multiple files inside a ZIP archive should provide independent
read state per entry, matching normal filesystem behavior.

## Reproduction

A `ZIP_STORED` archive with multiple entries and interleaved seek/read
calls can return bytes from the wrong entry. Extracting the archive first
avoids the issue because normal files have independent file handles.

julianghill:
Uncompressed files inside a ZIP could be read incorrectly when multiple
files from the same ZIP were open at the same time. This happened with
Velociraptor collection ZIPs, where registry and EVTX parsing can jump
around inside several files.

This PR gives those uncompressed ZIP files their own ZIP handle when
possible. That keeps reads from one file from affecting reads from
another file. Compressed ZIP files keep the existing behavior.

Proposed Changes

- Open uncompressed files inside ZIP archives with their own ZIP handle.
- Keep the existing behavior for compressed ZIP files.
- Support ZIP files opened from disk and ZIP files opened from memory.
- Add tests that open multiple files inside a ZIP and seek/read from them.

Tested with:

```bash
ruff check dissect/target/filesystems/zip.py tests/filesystems/test_zip.py
ruff format --check dissect/target/filesystems/zip.py tests/filesystems/test_zip.py
pytest tests/filesystems/test_zip.py -q
pytest tests/loaders/test_velociraptor.py -q
```

Also verified against a Velociraptor collection ZIP

## Checklist

- [x] Closes related issue number

Closes #1746

Let me know if changes are needed.

### DIFF
--- a/dissect/target/filesystems/zip.py
+++ b/dissect/target/filesystems/zip.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
+import io
+import os
 import stat
 import zipfile
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import TYPE_CHECKING, BinaryIO
 
 from dissect.util.stream import BufferedStream
@@ -98,6 +101,8 @@ class ZipFilesystemEntry(VirtualDirectory):
             return self._resolve().open()
 
         try:
+            if self.entry.compress_type == zipfile.ZIP_STORED and (zip_file := _reopen_zip_file(self.fs.zip)):
+                return _ZipFilesystemEntryStream(*zip_file, self.entry, self.fs.zip.pwd)
             return BufferedStream(self.fs.zip.open(self.entry), size=self.entry.file_size)
         except Exception:
             raise FileNotFoundError(self.path)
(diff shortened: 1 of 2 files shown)


Schamper:
Your description sounds almost more like a bug in CPython. Are you able to reproduce with just a small bare standard library script, no Dissect code?

Also, if I remove your changes in `zip.py`, the tests still pass. Could you make a test that highlights the problem better or otherwise share more information on how to reproduce?

Edit: I do see the test doesn't pass on 3.13, but could you still try to narrow the bug down further? I'd like to know what exactly causes this problem. Maybe we're seeing a symptom of a larger problem, and rather than just fixing the symptom, I'd like to understand the full scope of the problem.

julianghill:
> Your description sounds almost more like a bug in CPython. Are you able to reproduce with just a small bare standard library script, no Dissect code?
> 
> Also, if I remove your changes in `zip.py`, the tests still pass. Could you make a test that highlights the problem better or otherwise share more information on how to reproduce?
> 
> Edit: I do see the test doesn't pass on 3.13, but could you still try to narrow the bug down further? I'd like to know what exactly causes this problem. Maybe we're seeing a symptom of a larger problem, and rather than just fixing the symptom, I'd like to understand the full scope of the problem.

You are correct, this can be reproduced with CPython alone. I have found what appears to be the root cause, but the fix will have to wait until I am back from vacation. Thanks for the feedback!

julianghill:
Problem is in CPython, not Dissect.

PR #1753, opened 2026-05-14
Return None instead of skipping lines without timestamp in year_rollo…

EmperorValkorion:
# Yield ts as None instead of skipping lines in year_rollover_helper that have no timestamp

This way the helper is also compatible with multiline logs. Also updated test_messages.py, since the new logic caused the number of results for the year_rollover tests to change from 2 to 3. Maybe worth looking into if messages plugin should be updated with the new logic. @JSCU-CNI was there a reason that the original implemention skipped lines without a valid timestamp fully instead of just yielding ts as None?

### DIFF
--- a/dissect/target/helpers/utils.py
+++ b/dissect/target/helpers/utils.py
@@ -110,7 +110,7 @@ STRIP_RE = re.compile(r"^[\s\x00]*|[\s\x00]*$")
 
 def year_rollover_helper(
     path: Path, re_ts: str | re.Pattern, ts_format: str, tzinfo: tzinfo = timezone.utc
-) -> Iterator[tuple[datetime, str]]:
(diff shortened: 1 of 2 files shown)


JSCU-CNI:
> @JSCU-CNI was there a reason that the original implemention skipped lines without a valid timestamp fully instead of just yielding ts as None?

Yes, this method was not created with multi-line log messages in mind, but for structured log messages. If a structured log message is malformed we chose to discard it as the forensic value of the record can not be determined without manual review (e.g. we do not know what part of the record was malformed). 

Yielding the message without a timestamp seems fine to me, I do not have a strong opinion on that. The previous behavior was mainly intended to catch the attention of the user - hopefully causing them to investigate and either 1) fix the parser or 2) conclude the log is partially corrupted.

PR #1755, opened 2026-05-19
Fix breaking shell when loading histfile fails

JSCU-CNI:
It seems #1726 does not handle some malformed history files correctly. This try/catch fixes a broken shell when such a file is encountered.


yunzheng (changes requested):
Oops, missed this. Wonder how a histfile can get corrupted though. Can you also add a check to only load the history when the file exists, similar to the old code? Else it would still log an error when the histfile doesn't exist.

JSCU-CNI:
Corrupted is just an assumption I made, this seems to occur on a regular `~/.dissect_history` file with 10k lines.

```
/$ 2026-05-19T13:27:17.401808Z [error    ] Loading history failed [prompt_toolkit.buffer]
╭─────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────────────────────╮
│ /tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/buffer.py:405 in load_history_done                                              │
│                                                                                                                                                                                 │
│    402 │   │   │   │   when an exception was raised.                                                                                                                            │
│    403 │   │   │   │   """                                                                                                                                                      │
│    404 │   │   │   │   try:                                                                                                                                                     │
│ ❱  405 │   │   │   │   │   f.result()                                                                                                                                           │
│    406 │   │   │   │   except asyncio.CancelledError:                                                                                                                           │
│    407 │   │   │   │   │   # Ignore cancellation. But handle it, so that we don't get                                                                                           │
│    408 │   │   │   │   │   # this traceback.                                                                                                                                    │
│                                                                                                                                                                                 │
│ ╭────────────────────────────────────────────────────────────────────────────────── locals ───────────────────────────────────────────────────────────────────────────────────╮ │
│ │ f = <Task finished name='Task-28' coro=<Buffer.load_history_if_not_yet_loaded.<locals>.load_history() done, defined at                                                      │ │
│ │     /tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/buffer.py:392> exception=OSError(22, 'Invalid argument')>               │ │
│ ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                                                                                 │
│ /tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/buffer.py:393 in load_history                                                   │
│                                                                                                                                                                                 │
│    390 │   │   if self._load_history_task is None:                                              ╭────────────────────────────── locals ──────────────────────────────╮          │
│    391 │   │   │                                                                                │ self = <Buffer(name='DEFAULT_BUFFER', text='') at 131316770579056> │          │
│    392 │   │   │   async def load_history() -> None:                                            ╰────────────────────────────────────────────────────────────────────╯          │
│ ❱  393 │   │   │   │   async for item in self.history.load():                                                                                                                   │
│    394 │   │   │   │   │   self._working_lines.appendleft(item)                                                                                                                 │
│    395 │   │   │   │   │   self.__working_index += 1                                                                                                                            │
│    396                                                                                                                                                                          │
│                                                                                                                                                                                 │
│ /tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/history.py:60 in load                                                           │
│                                                                                                                                                                                 │
│    57 │   │   method is called.                                                                                                                                                 │
│    58 │   │   """                                                                                                                                                               │
│    59 │   │   if not self._loaded:                                                                                                                                              │
│ ❱  60 │   │   │   self._loaded_strings = list(self.load_history_strings())                                                                                                      │
│    61 │   │   │   self._loaded = True                                                                                                                                           │
│    62 │   │                                                                                                                                                                     │
│    63 │   │   for item in self._loaded_strings:                                                                                                                                 │
│                                                                                                                                                                                 │
│ ╭────────────────────────────────────── locals ──────────────────────────────────────╮                                                                                          │
│ │ self = <dissect.target.tools.shell.ReadlineCompatHistory object at 0x776e91769940> │                                                                                          │
│ ╰────────────────────────────────────────────────────────────────────────────────────╯                                                                                          │
│                                                                                                                                                                                 │
│ /tmp/dissect.target/dissect/target/tools/shell.py:1966 in load_history_strings                                                                        │
│                                                                                                                                                                                 │
│   1963 │   │                                                                                                                                                                    │
│   1964 │   │   def load_history_strings(self) -> Iterable[str]:                                                                                                                 │
│   1965 │   │   │   """Yield most recent history strings from the readline history file."""                                                                                      │
│ ❱ 1966 │   │   │   readline.read_history_file(self.histfile)                                                                                                                    │
│   1967 │   │   │   for i in range(readline.get_current_history_length(), 0, -1):                                                                                                │
│   1968 │   │   │   │   yield readline.get_history_item(i)                                                                                                                       │
│   1969                                                                                                                                                                          │
│                                                                                                                                                                                 │
│ ╭────────────────────────────────────── locals ──────────────────────────────────────╮                                                                                          │
│ │ self = <dissect.target.tools.shell.ReadlineCompatHistory object at 0x776e91769940> │                                                                                          │
│ ╰────────────────────────────────────────────────────────────────────────────────────╯                                                                                          │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
OSError: [Errno 22] Invalid argument


Unhandled exception in event loop:
  File "/tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/buffer.py", line 393, in load_history
    async for item in self.history.load():
        self._working_lines.appendleft(item)
        self.__working_index += 1
  File "/tmp/dissect.target/.venv/lib/python3.13/site-packages/prompt_toolkit/history.py", line 60, in load
    self._loaded_strings = list(self.load_history_strings())
                           ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/dissect.target/dissect/target/tools/shell.py", line 1966, in load_history_strings
    readline.read_history_file(self.histfile)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^

Exception [Errno 22] Invalid argument
Press ENTER to continue...
```

Schamper:
> a regular `~/.dissect_history` file with 10k lines

he dissec, he protec, he did not keep his histfile in chec

yunzheng:
The error seems to occur in `readline.read_history_file()`. But i'm not sure how it would raise a `Errno 22 (EINVAL)`, can you check if the path exists and is a file?

JSCU-CNI:
> The error seems to occur in readline.read_history_file(). But i'm not sure how it would raise a Errno 22 (EINVAL), can you check if the path exists and is a file?

File exists and is a file, was able to reproduce this on two separate Linux machines running Ubuntu 24.04 (amd64) using cpython 3.13.13.

yunzheng:
> > The error seems to occur in readline.read_history_file(). But i'm not sure how it would raise a Errno 22 (EINVAL), can you check if the path exists and is a file?
> 
> File exists and is a file, was able to reproduce this on two separate Linux machines running Ubuntu 24.04 (amd64) using cpython 3.13.13.

Thanks, I can also reproduce in docker, it seems to happen when the history file does not have the magic header `_HiStOrY_V2_`. 

```shell
$ docker run -it --rm ubuntu:24.04 bash

ubuntu$ # install uv
ubuntu$ seq 100 > ~/.dissect_history
ubuntu$ uv run -p 3.13.13 --with "dissect.target[full]" --pre --force-reinstall target-shell /
```

This behavior might vary between Python versions I think. These issues might be related:

* https://github.com/python/cpython/issues/121160
* https://github.com/python/cpython/issues/120766

yunzheng:
@Schamper any thoughts on how to move forward with this? to recap: this seems be caused due different libraries (libedit/libreadline) being used/linked between Python versions. Also Python is not going to fix this, it only added a [note](https://docs.python.org/3/library/readline.html) in the docs:

```
Also note that different libraries may use different history file formats.
When switching the underlying library, existing history files may become unusable.
```

I think there are several options:

1. when it fails to load the history file, try to fix it by prepending the `_HiStOrY_V2_` string to the file and load it again.
2. catch the exception, don't read the history and start over. (this loses history)
3. use `~/.dissect_history_pt` for prompt_toolkit history and use it's own History format (this one has timestamps), when it doesn't exist yet try to import the old history file from `~/.dissect_history`.
4. other options?

@JSCU-CNI  feel free to chime in.

Schamper:
I like the sound of 3. 1 sounds like a bit of a hack, and 2 doesn't sound very nice to lose history like that (never forget).

JSCU-CNI:
Agree with @Schamper, approach 3 seems the most sane.

JSCU-CNI:
Could this solution be merged for the meantime? The shell also breaks if no `.dissect_history` file exists (e.g. on a fresh dissect install).

yunzheng (approved):

yunzheng:
> Agree with @Schamper, approach 3 seems the most sane.

i'll make a new issue for the improved prompt history suggestions.

meanwhile, I reviewed the current changes and they look good for merging.

PR #1757, opened 2026-05-19
Add Windows OEM device information to target-info

JSCU-CNI:
This PR adds the `device` property to the Windows implementation of `OSPlugin`. It is populated by OEM manufacturer and model names stored in the Windows registry.

PR #1759, opened 2026-05-19
Add iOS OEM device information to target-info

JSCU-CNI:
Depends on #1757 for output to `target-info`.


dissect/target/plugins/os/unix/bsd/darwin/util/model.py
```diff
@@ -0,0 +1,268 @@
+from __future__ import annotations
+
+# Reference: https://theapplewiki.com/wiki/Models
+# Does not include Mac models.
```
Schamper, line 4: Why not? Too many?

JSCU-CNI: Yes. Mac also seems to store this somewhere else, so out of scope for this PR.

Schamper (changes requested):
Can you fix the linter?

Schamper (approved):

PR #1762, opened 2026-05-20
Fix cstruct hexdump invocation

Schamper:
Regression from https://github.com/fox-it/dissect.cstruct/pull/158


yunzheng:
Was this the only one?

yunzheng (approved):

Schamper:
> Was this the only one?

Yes, all other invocations already use kwargs.

PR #1763, opened 2026-05-20
Add Android persistent property plugin

JSCU-CNI:
Improves the Android `OSPlugin` implementation using persistent property (protobuf and file-based) parsing for functions `version`, `device`, `timezone` and `language`. Added missing `architecture`, `activity` and `install_date` while I was at it.


dissect/target/plugins/os/unix/linux/android/_os.py
```diff
@@ -21,34 +24,22 @@ class AndroidPlugin(LinuxPlugin):
...
-                )
-            except Exception as e:  # noqa: PERF203
-                self.target.log.warning("Unable to parse Android build.prop file %s: %s", build_prop, e)
+        # Populate props with Android build.prop files.
+        self.build_prop_paths = list(find_build_props(self.target.fs))
+        self.props.update(parse_build_props(self.build_prop_paths))
+
+        # Add persistent properties (``persist.*``) to props.
+        if (dir := self.target.fs.path("/data/property")).is_dir():
+            self.props.update(read_persistent_props(dir))
```
Schamper: It's a bit of a stylistic choice perhaps, but could it be nicer to move the property stuff from being a "utility", to it being its own plugin, similar to the registry for Windows?

Schamper (changes requested):

Schamper (approved):

PR #1761, opened 2026-05-19
Add Android Binary XML (ABX) parser

JSCU-CNI:
This PR adds an ABX (Android Binary XML) parser. The `AndroidPlugin` is extended to use ABX features (`hostname` and `users`). `target-shell` gained the `abx` command to read ABX files on the fly.

PR #1765, opened 2026-05-20
Add Android applications plugin

JSCU-CNI:
This PR adds an Android `applications` plugin and an Android AXML parser (not to be confused with ABX or regular XML). Depends on #1761.

### DIFF
--- a/dissect/target/plugins/os/unix/linux/android/applications.py
+++ b/dissect/target/plugins/os/unix/linux/android/applications.py
@@ -1,0 +1,220 @@
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING
+
+from asn1crypto import cms, pem
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.filesystems.zip import ZipFilesystem
+from dissect.target.helpers.certificate import COMMON_CERTIFICATE_FIELDS, parse_x509
+from dissect.target.helpers.record import (
+    COMMON_APPLICATION_FIELDS,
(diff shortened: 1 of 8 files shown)


dissect/target/plugins/os/unix/linux/android/applications.py
```diff
@@ -0,0 +1,220 @@
...
+            AndroidApplicationExportsRecord,
+        ]
+    )
+    def applications(
+        self,
+    ) -> Iterator[AndroidApplicationRecord | AndroidApplicationCertificateRecord | AndroidApplicationExportsRecord]:
+        """Yield installed Android apps."""
+        for package_file in self.target.fs.path("/").glob("*/system/packages.xml"):
+            try:
+                root = read_android_xml(package_file.open("rb"))
```
Schamper, line 95: Can you use a context manager?

dissect/target/plugins/os/unix/linux/android/applications.py
```diff
@@ -0,0 +1,220 @@
...
+        # Sometimes, an apk referenced in packages.xml is a directory, containing the actual apk.
+        if path.is_dir():
+            apks = list(path.glob("*.apk"))
+            if len(apks) == 1:
+                path = apks[0]
+            else:
+                return None
+
+        fs = ZipFilesystem(path.open())
+        et = read_android_xml(fs.get("AndroidManifest.xml").open())
```
Schamper, line 166: Can you use context managers?

Schamper (changes requested):

PR #1772, opened 2026-05-21
Add zip and tar filesystem benchmarks

Schamper:
Adds a simple benchmark for the zip and tar filesystem, as well as add tests for absolute pathed zips and tars.

Ominous. To test against a future change.

Also, remove the `exists` method on `DirFilesystemEntry`, no idea where that came from, it's not a method on the base class.

### DIFF
--- a/dissect/target/filesystems/tar.py
+++ b/dissect/target/filesystems/tar.py
@@ -60,11 +60,11 @@ class TarFilesystem(Filesystem):
         self._fs = VirtualFilesystem(alt_separator=self.alt_separator, case_sensitive=self.case_sensitive)
 
         for member in self.tar.getmembers():
-            mname = member.name.removeprefix("./").strip("/")
+            mname = member.name.removeprefix("./")
             if not mname.startswith(self.base) or mname == ".":
                 continue
 
-            rel_name = fsutil.normpath(mname[len(self.base) :], alt_separator=self.alt_separator)
+            rel_name = fsutil.normpath(mname.removeprefix(self.base), alt_separator=self.alt_separator).strip("/")
 
             entry_cls = TarFilesystemDirectoryEntry if member.isdir() else TarFilesystemEntry
             file_entry = entry_cls(self, rel_name, member)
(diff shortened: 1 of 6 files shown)


dissect/target/filesystems/zip.py
```diff
@@ -60,11 +60,10 @@ def __init__(
         self._fs = VirtualFilesystem(alt_separator=self.alt_separator, case_sensitive=self.case_sensitive)
 
         for member in self.zip.infolist():
-            mname = member.filename.strip("/")
-            if not mname.startswith(self.base) or mname == ".":
+            if not member.filename.startswith(self.base) or member.filename in (".", "./"):
                 continue
 
-            rel_name = self._resolve_path(mname)
+            rel_name = self._resolve_path(member.filename).strip("/")
```
twiggler, line 66: idem

dissect/target/filesystems/tar.py
```diff
@@ -60,11 +60,11 @@ def __init__(
         self._fs = VirtualFilesystem(alt_separator=self.alt_separator, case_sensitive=self.case_sensitive)
 
         for member in self.tar.getmembers():
-            mname = member.name.removeprefix("./").strip("/")
+            mname = member.name.removeprefix("./")
```
twiggler, line 63: Is this change safe?

What if the tar has members like `///joehoe` or `/./joehoe`?

Also, can we get collisions because we strip after normalization?

twiggler (changes requested):

dissect/target/filesystems/tar.py
Schamper: All tests pass so I consider this a safe change for now. This change was so that the absolute path tests actually work in this implementation. I can also remove that test (because this implementation is not compatible with it) and just re-add that test in PR that follows this one.

Keep in mind that this PR just adds the benchmark and a few more tests, and the PR that follows this one will provide a complete re-implementation of both the tar and zip filesystem implementations. Any such edge cases should be dealt with in that new implementation, IMO.

If you have more ideas for edge cases, please do make a note or suggest them for added test cases.

twiggler: > Any such edge cases should be dealt with in that new implementation, IMO.

That is acceptable

> If you have more ideas for edge cases, please do make a note or suggest them for added test cases.

Will think.

twiggler (approved):

PR #1774, opened 2026-05-27
Build windows timezone map statically instead of dynamically

JSCU-CNI:
Should slightly improve dissect initialization speed.


Schamper (dismissed):

tests/helpers/locale/test_util.py
Schamper, line 1: Actually, can you move this (and rename stuff accordingly) to match the new file location? So `tests/helpers/locale/test_util.py`?

Schamper (changes requested):

Schamper (approved):

PR #1776, opened 2026-05-28
Add support for VeraCrypt system volumes

JSCU-CNI:
This PR adds support for VeraCrypt system volumes. Depends on https://github.com/fox-it/dissect.fve/pull/27.


dissect/target/volumes/veracrypt.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self, fh: BinaryIO | Volume, *args, **kwargs):
+        super().__init__(fh, *args, **kwargs)
+        self.veracrypt = VeraCrypt(fh.disk if isinstance(fh, Volume) else fh, is_system=True)
+
+    @staticmethod
+    def _detect(fh: BinaryIO | Volume) -> bool:
+        """Detect by checking for certain VeraCrypt volume characteristics and if any keychain value(s) are given."""
+        # NOTE: For VeraCrypt encrypted system volumes, the Windows EFI partition will include VeraCrypt files.
+        # However, from this context we do not have access to that filesystem, so we have to be a little more creative.
```
Schamper, line 49: Ah, interesting. I'm trying to think if there's any way we can possibly improve this in the future. 

I first thought maybe we can optionally pass in a `Target | None` to these methods, but then I think you have a chicken and egg problem (not all volumes may have been added to the `Target` object yet, and we're ahead of filesystem loading at this stage). 

Another possible solution within current mechanisms could be a custom OS plugin, one that just detects VeraCrypt files on the Windows EFI partition, adds it and re-triggers filesystem parsing and OS loading. You'd be duplicating a lot of logic though.

Final idea I quickly had would be to introduce a new concept of "postprocessing" loaders. Where regular loaders are used to go from raw data to a `Target`, this "loader" type would perform final post processing steps on a pre-OS `Target` to fix up small things like this.

I'd be interested to hear your thoughts on this and if you think either of these could be interesting enough to explore more in the future. In which case, I can make an issue for that.

dissect/target/volumes/veracrypt.py
```diff
@@ -0,0 +1,119 @@
...
+            volume_details = {
+                "number": 1,
+                "offset": 0,
+                "vtype": None,
+                "name": None,
+            }
+
+        try:
+            stream = self.unlock_volume()
+        except VeraCryptVolumeSystemError:
```
Schamper: This error is silently ignored now. It's also slightly different how we do this on e.g. BDE, could you comment on that? Is the error handling of the code that calls this code not good enough to handle these failures?

Schamper (changes requested):

Schamper:
Could you also add a small unit test, if possible.

dissect/target/volumes/veracrypt.py
JSCU-CNI: This is definitely something that we should tackle in the near future. We also (kind of) run into this chicken-egg problem with non-system FVE volumes (Bitlocker, LUKS) that can be mounted using keys present on the system volume (e.g. BitLocker autounlock, LUKS crypttab).

dissect/target/volumes/veracrypt.py
JSCU-CNI: Haven't really put a thought into this, I'll remove the try/catch to match the BDE implementation.

JSCU-CNI:
Added a small test in [cb276b9](https://github.com/fox-it/dissect.target/pull/1776/commits/cb276b938dbd76bd967aa837129d2737cbc954ef). Also implemented support for non-system volumes to make the test work.

dissect/target/volumes/veracrypt.py
```diff
@@ -0,0 +1,162 @@
...
+    likely the system volume.
+
+    For performance reasons we do not attempt to load the uninitialized filesystems of these partitions.
+    A future version could check if the last partition of the disk is the WinRE partition, as that seems
+    to be a requirement for recent Windows versions.
+    """
+    if not isinstance(fh, Volume):
+        return False
+
+    index = fh.vs.volumes.index(fh)
```
Schamper, line 148: I get an `AttributeError` here when I run `target-shell` against the test file.

Schamper (changes requested):

dissect/target/volumes/veracrypt.py
JSCU-CNI: Fixed in [95b60e5](https://github.com/fox-it/dissect.target/pull/1776/commits/95b60e58510c3b1959d1993b866033d3c5afc1e7)

Schamper (approved):

dissect/target/volumes/veracrypt.py
Schamper: I created https://github.com/fox-it/dissect.target/issues/1791. 

I believe the other examples you give could be resolved within e.g. `WindowsPlugin` and `LinuxPlugin` though, right? You could argue that BitLocker autounlock and LUKS crypttab are just parts of the mounting process that we have not yet implemented.

Schamper (changes requested):
Oh actually, I forgot. Could you bump the version in `pyproject.toml`?

Schamper (approved):

dissect/target/volumes/veracrypt.py
JSCU-CNI: It is a bit more complicated unfortunately, as we would need to access the user registry hives for the bitlocker autounlocking, which is not available yet during the initialization of the Windows OS plugin. 

Perhaps a `__post__init__` method in `OSPlugin` or something like that, called just after `self._os = self.add_plugin(os_plugin)` in `target.py` would work.

Schamper: Hm I see. Perhaps in `Target.apply` we can add something like a call to `self._os.post()`? To keep naming similar to `create`.

PR #1781, opened 2026-06-06
Fix ValueError in MRU plugin when running mru.recentdocs

JazzCore:
Got this exception when running `mru.recentdocs` on 2 separate machines:

```
  File "/home/user/venv/bin/target-query", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/tools/utils/cli.py", line 475, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/tools/query.py", line 260, in main
    for record in record_generator:
                  ^^^^^^^^^^^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 164, in recentdocs
    yield from parse_mru_ex_key(self.target, key, RecentDocsRecord)
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 371, in parse_mru_ex_key
    entry_index = mrulist_ex.index(int(value.name))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: tuple.index(x): x not in tuple
```

This happens when value is present in registry path (`RecentDocs` for `mru.recentdocs`), but its `name` value is missing in `MRUListEx`.

There are 2 ways to fix this:

 - Just skip those entries like this:
    ```
    entry_name = int(value.name)
    if mrulist_ex is None or entry_name not in mrulist_ex:
      continue
    entry_index = mrulist_ex.index(entry_name)
    ```

    Other MRU parsers I checked (plaso etc.) does it

 - Set some fallback index value for those records and return them. I went with this fix, since in my case such entries on both machnies were correct and I see no drawbacks in returning them (but I should mention that on both machines they were duplicate of entry with index 0)

Not sure which fix is better here, I can go with either of them.


dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -364,7 +364,13 @@ def parse_mru_ex_key(target: Target, key: RegistryKey, record: TargetRecordDescr
         if value.name == "MRUListEx":
             continue
 
-        entry_index = mrulist_ex.index(int(value.name))
+        # value can be missing in MRUListEx
+        # instead of skipping, set high entry_index for it to separate from correct entries    
+        if int(value.name) not in mrulist_ex:
+            entry_index = 9999
```
Schamper: I believe all value types in records are null/none-able, so perhaps we can just set `None` here. Seems more appropriate.


```suggestion
            entry_index = None
```

tests/plugins/os/windows/test_mru.py
```diff
@@ -186,7 +189,8 @@ def test_mru_plugin(target_win_mru: Target) -> None:
     msoffice = list(target_win_mru.mru.msoffice())
 
     assert len(run) == 2
-    assert len(recentdocs) == 1
+    assert len(recentdocs) == 2
+    assert any(r.index == 9999 for r in recentdocs)
```
Schamper:
```suggestion
    assert any(r.index is None for r in recentdocs)
```

Schamper (changes requested):

dissect/target/plugins/os/windows/regf/mru.py
JazzCore: Agreed, did not thought of it for some reason, thanks :)

Linter also wants it changed to ternary operator

JazzCore:
@Schamper Should be good now

Schamper:
Can you resolve the conflict?

JazzCore:
Done

tests/plugins/os/windows/test_mru.py
```diff
@@ -41,6 +41,9 @@ def target_win_mru(target_win_users: Target) -> Target:
     )
     recentdocs_key.add_value("0", VirtualValue(user_hive, "0", recentdocs_value))
 
+    # value name is not present in MRUListEx, but should be still found with index 9999
```
Schamper:
```suggestion
    # value name is not present in MRUListEx, but should be still found with index None
```

Schamper (changes requested):
Small change for consistency.

Schamper (approved):

PR #1782, opened 2026-06-06
Fix ValueError in MRU plugin when running mru.msoffice

JazzCore:
Got this exception when running `mru.msoffice`:

```
  File "/home/user/venv/bin/target-query", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/tools/utils/cli.py", line 475, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/tools/query.py", line 260, in main
    for record in record_generator:
                  ^^^^^^^^^^^^^^^^
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 307, in msoffice
    yield from parse_office_mru(self.target, subkey, MSOfficeMRURecord)
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 435, in parse_office_mru
    yield from parse_office_mru(target, subkey, record)
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 429, in parse_office_mru
    yield from parse_office_mru_key(target, key.subkey("Place MRU"), record)
  File "/home/user/venv/lib/python3.12/site-packages/dissect/target/plugins/os/windows/regf/mru.py", line 447, in parse_office_mru_key
    entry_index = int(value.name.split(" ")[1])
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'Metadata'
```

In my case in addition to regular registry office MRU values like `Item 1, Item 2` in File/Place MRU there were extra values with names like `Item Metadata 1` and data like this: `<Metadata><AppSpecific><OneNoteNotebookUrl>C:\Path</OneNoteNotebookUrl></AppSpecific></Metadata>`

### DIFF
--- a/dissect/target/plugins/os/windows/regf/mru.py
+++ b/dissect/target/plugins/os/windows/regf/mru.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 import struct
 from typing import TYPE_CHECKING
 
@@ -113,8 +114,12 @@ MSOfficeMRURecord = UserRegistryRecordDescriptor(
         ("varint", "index"),
         ("string", "value"),
         ("string", "key"),
+        ("string", "metadata"),
     ],
 )
+
+RE_OFFICEMRU_VALUE_NAME = re.compile(r"^Item (\d+)$")
+RE_OFFICEMRU_METADATA_VALUE_NAME = re.compile(r"^Item Metadata (\d+)$")
 
 
 class MRUPlugin(Plugin):
@@ -434,20 +439,25 @@ def parse_office_mru(target: Target, key: RegistryKey, record: TargetRecordDescr
 def parse_office_mru_key(target: Target, key: RegistryKey, record: TargetRecordDescriptor) -> Iterator[Record]:
     user = target.registry.get_user(key)
 
+    mru_metadata = {}
+    mru_data = []
+
     for value in key.values():
-        if not value.name.startswith("Item"):
-            continue
+        if match := RE_OFFICEMRU_METADATA_VALUE_NAME.fullmatch(value.name):
+            mru_metadata[int(match.group(1))] = value.value
+        elif match := RE_OFFICEMRU_VALUE_NAME.fullmatch(value.name):
+            entry_index = int(match.group(1))
+            info_str, path = value.value.split("*", 1)
+            info = {part[0]: int(part[1:], 16) for part in info_str.strip("[]").split("][")}
+            mru_data.append((entry_index, info, path))
 
-        entry_index = int(value.name.split(" ")[1])
-        info_str, path = value.value.split("*", 1)
-
-        info = {part[0]: int(part[1:], 16) for part in info_str.strip("[]").split("][")}
-
+    for entry_index, info, path in mru_data:
         yield record(
             ts=wintimestamp(info["T"]),
             regf_mtime=key.ts,
             index=entry_index,
             value=path,
+            metadata=mru_metadata.get(entry_index),
             _target=target,
             _user=user,
             _key=key,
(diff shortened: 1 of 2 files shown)


Schamper:
Is this the only variant you encountered? Maybe it's worth yielding a record for this? Can be the same record type, just with the "raw value" as value, and no `ts` value, since we can't derive one.

JazzCore:
Yeah, this was only one I seen. Also while checking seen another example on SO - https://stackoverflow.com/questions/71692905/batch-script-to-change-a-value-in-a-registry-key-for-templates-of-microsoft-offi.

For yielding record you mean for this `Item Metadata` field? I dont see the need to add new record just for it, since it's a part of `msoffice` record anyway. We can probably extend `MSOfficeMRURecord` record to add `metadata` field for it and set it to raw key data, if exists.  Not sure about backwards compatibility of extending already existing record though. If this is a concern - then yea, probably better to just yield this field separately. Another question if this data is even interesting and merits a yield...

Schamper:
> We can probably extend `MSOfficeMRURecord` record to add `metadata` field for it and set it to raw key data, if exists. Not sure about backwards compatibility of extending already existing record though. If this is a concern - then yea, probably better to just yield this field separately.

Oh yeah, hadn't considered that. Adding a `metadata` field seems like a good idea. Backwards compatibility is not really a concern in that regard.

> Another question if this data is even interesting and merits a yield...

🤷. I do agree it's probably not of enough value to add a whole new record for it, but adding it as a `metadata` field doesn't seem overly complex and does slightly contribute towards that synthetic warm fuzzy feeling for making a parser slightly more complete.

JazzCore:
Added `metadata` field to `MSOfficeMRURecord` record

Schamper:
Can you resolve the conflicts :)

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -113,9 +114,13 @@
         ("varint", "index"),
         ("string", "value"),
         ("string", "key"),
+        ("string", "metadata"),
     ],
 )
 
+OFFICEMRU_VALUE_NAME_RE = re.compile(r"^Item \d+$")
+OFFICEMRU_METADATA_VALUE_NAME_RE = re.compile(r"^Item Metadata \d+$")
```
Schamper:
```suggestion
RE_OFFICEMRU_VALUE_NAME = re.compile(r"^Item (\d+)$")
RE_OFFICEMRU_METADATA_VALUE_NAME = re.compile(r"^Item Metadata (\d+)$")
```

dissect/target/plugins/os/windows/regf/mru.py
```diff
@@ -431,20 +436,25 @@ def parse_office_mru(target: Target, key: RegistryKey, record: TargetRecordDescr
...
-
-        info = {part[0]: int(part[1:], 16) for part in info_str.strip("[]").split("][")}
+    mru_metadata = {}
+    mru_data = []
 
+    for value in key.values():
+        if OFFICEMRU_METADATA_VALUE_NAME_RE.search(value.name):
+            mru_metadata[int(value.name.split(" ")[2])] = value.value
+        elif OFFICEMRU_VALUE_NAME_RE.search(value.name):
+            entry_index = int(value.name.split(" ")[1])
```
Schamper:
```suggestion
        if match := RE_OFFICEMRU_METADATA_VALUE_NAME.fullmatch(value.name):
            mru_metadata[int(match.group(1))] = value.value
        elif RE_OFFICEMRU_VALUE_NAME.fullmatch(value.name):
            entry_index = int(match.group(1))
```

Schamper (changes requested):

JazzCore:
> Can you resolve the conflicts :)

Done :)

Schamper (approved):

Schamper:
Huge thanks for all the fixes with detailed explanations and tests, all the effort that went into it is highly appreciated!

PR #1787, opened 2026-06-10
Add support for new EWF implementation

Schamper:
Ensures everything works fine with https://github.com/fox-it/dissect.evidence/pull/42. Not really any big changes, mostly just adding support for ADCRYPT (copy pasted from AD1), and cleaned up the opening.


JSCU-CNI:
LGTM.

dissect/target/containers/ewf.py
```diff
@@ -47,3 +59,4 @@ def tell(self) -> int:
 
     def close(self) -> None:
         self._stream.close()
```
JSCU-CNI: ~I think `_stream` does not exist anymore in `EwfContainer`.~ The behavior I observed has to do with dissect.target `main` not being compatible with `main` on dissect.evidence. (The `__init__` crashes, causing `close()` to be called and attempting to access `EwfContainer._stream` when that attribute does not exist).

Schamper: We set it on line 40.

JSCU-CNI: See my updated comment. If `EWF(fh)` fails, `close()` is called.

JSCU-CNI: I'd suggest adding a guard (e.g. `if hasattr(self, "ewf"):`) to the contents of `close()` to prevent confusing stacktraces when `EWF(fh)` fails (e.g. when the `fh` is not an EWF file).

Schamper: Yes, Just did that.

twiggler (approved):

PR #1792, opened 2026-06-16
Fix APFS implementation due to VFS changes

Schamper:
https://github.com/fox-it/dissect.target/pull/1301 caused an previously unnoticed bug to the APFS implementation in `dissect.target`.


Miauwkeru (approved):
LGTM

Issue #1835, opened 2026-06-18
Target architecture does not return expected value for macOS targets


Schamper:
Looks like https://github.com/fox-it/dissect.target/pull/1733 does not function as intended on macOS targets, or at least disk images of ARM virtual machines:

```
❯ uv run target-info ~/Downloads/ED213FEE-7B2F-410F-A7C3-C5976EBF7903.img
[...]
Architecture   : 2-apple-macos
```

@JSCU-CNI could you provide some insight on how you implemented/tested this?

JSCU-CNI:
See https://github.com/fox-it/dissect.target/pull/1870 for a test example. My guess is that your image contains a multi-arch mach-o binary (https://en.wikipedia.org/wiki/Mach-O#Multi-architecture_binaries). Could you share the image?

Schamper:
> Could you share the image?

It's just a standard macOS virtual machine created through UTM on a Apple Silicon MacBook. I'll see if I can share any Mach-O files later, or find any other leads.

PR #1870, opened 2026-06-22
Improve macOS tests

JSCU-CNI:
To aid with #1835 this PR improves existing macOS functionality by adding test data from https://cfreds.nist.gov/all/Hexordia/2026MVSCTFMac. The `MacOSPlugin.architecture` attribute is now properly tested.


dissect/target/plugins/os/unix/history.py
```diff
@@ -63,9 +64,19 @@ def _find_history_files(self) -> list[tuple[str, TargetPath, UnixUserRecord]]:
...
-                    history_files.append((shell, history_path, user_details.user))
+                if "*" in history_relative_path:
+                    base, _, glob = history_relative_path.partition("*")
+                    history_dir = user_details.home_path.joinpath(base)
+                    if history_dir.is_dir():
+                        history_files.extend(
+                            (shell, history_path, user_details.user)
+                            for history_path in history_dir.glob(f"*{glob}")
+                            if history_path.is_file()
+                        )
```
Schamper: Wouldn't a `user_details.home_path.glob(history_relative_path)` already work? In theory, the logic you're trying to implement here is already implemented in the pathlib globbing logic, and this variant currently only works if the `*` is right after a path separator.

Schamper (changes requested):

dissect/target/plugins/os/unix/history.py
JSCU-CNI: Good point, simplified that part in [6b34c9c](https://github.com/fox-it/dissect.target/pull/1870/commits/6b34c9c581f6abc3eb69c9878eb2503f4168991c)

Schamper (changes requested):
Can you fix the unit tests?

Schamper (approved):

PR #1873, opened 2026-07-01
Ignore boot partitions when scanning mounts for Trash folders

JSCU-CNI:
This patch makes sure the `GnomeTrashPlugin` does not end up scanning FAT partitions. This is a workaround for a recursion bug in traversing FAT filesystems which hopefully is fixed with https://github.com/fox-it/dissect.fat/pull/37.

PR #1874, opened 2026-07-06
Fix abspath for special drive names

Schamper:
Needed to properly handle the "special drive names" in abspath to correctly deal with situations such as encountered in https://github.com/fox-it/dissect.target/pull/1873.


Miauwkeru (approved):
LGTM

Did notice that we might need to adjust the dirloader later, due to the changes in targetpath it will use the `efi/` for the efi disk, however the dirloader won't show it

Schamper:
> Did notice that we might need to adjust the dirloader later, due to the changes in targetpath it will use the `efi/` for the efi disk, however the dirloader won't show it

Can you make a new issue exactly describing what you mean?

PR #1509, opened 2026-01-20
Improve ESXiPlugin support for recent version

william-billaud:
Enhance ESXiOs plugin, especially for non disk collection. 

- [x] On ESXi7 and later : users are also stored in the configstore : Merge entries from Configstore and /etc/passwd when possible.
- [x] On Esxi8 and later : hostname/ips are also stored in the configstore
- [x] Add vm support sample from esxi 7 and 9 (in addition to already present 6 and 8), and a uac collection sample from the same ESXi9
- [x] Properly symlink log folders (related to #1385) : Symlink to /var/run/log from log_dir (defaulting to /scratch/log) as this is what is observed on live esxi hosts. I Was not able to test with disk where a locker partition is present. (ESXi9 sample has an unsual log dir) 
- [x] Ensure compatibility with raw disk (already broken before this PR).

## Other changes

* Change the configstore get plugins signature to ease key retrieval.

## Note

* Maybe we could have a public plugins to yields all entries from configstore?
* Remove the symlink between /var/log and log_dir, as what is observed on ESXi host is a bit more complex that a simple folder symlink (Only some files are symlinked). Furthermore /var/log may already contains some file (depending on how ESXi is collected).
* I still have no idea where the /etc/passwd file is stored on disk....

## Test on raw disk

hostname,os,domain,version,ips
```
❯ uv run --python 3.12 --refresh --extra dev --extra full target-query -f hostname,os,domain,version,ips --cmdb -d ";"   ~/Documents/VM/esxi_*
2026-01-21T14:14:30.343866Z [warning  ] Failed to detect <class 'dissect.target.volumes.bde.BitlockerVolumeSystem'> volume system [dissect.target.volume]
2026-01-21T14:14:30.344076Z [warning  ] Failed to detect <class 'dissect.target.volumes.bde.BitlockerVolumeSystem'> volume system [dissect.target.volume]
2026-01-21T14:14:30.344231Z [warning  ] Failed to detect <class 'dissect.target.volumes.bde.BitlockerVolumeSystem'> volume system [dissect.target.volume]
2026-01-21T14:14:30.344377Z [warning  ] Failed to detect <class 'dissect.target.volumes.bde.BitlockerVolumeSystem'> volume system [dissect.target.volume]
2026-01-21T14:14:30.509153Z [warning  ] Failed to detect fat filesystem [dissect.target.filesystem]
2026-01-21T14:14:30.509428Z [warning  ] Failed to detect fat filesystem [dissect.target.filesystem]
[...]

PR #1878, opened 2026-07-07
Add vmware vCenter support bundle and ESXi plugins

### LINKED ISSUE #1877: Add support for vmware vCenter support packages

Matthijsy:
Resolves #1877 

# Add vmware vCenter support bundle

This PR adds a child plugin to identify vmware vCenter support bundles. These support bundles are zip files consisting of tgz files which are basically forensic collects of ESXi hypervisors. 

Currently the name of the hypervisor is not extracted in the child plugin, as that would requiring opening all the tgz files, which is not nice from performance perspective.

Furthermore this PR also fixes the authentication log plugin to support the ESXi path and format. Lastly it introduces a new plugin to process the history file of an ESXi host. If we want these last two things in a seperate PR please let me know.

## Checklist

### DIFF
--- a/dissect/target/plugins/child/vmware_vcenter.py
+++ b/dissect/target/plugins/child/vmware_vcenter.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.helpers.record import ChildTargetRecord
+from dissect.target.plugin import ChildTargetPlugin
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from dissect.target.target import Target
+
+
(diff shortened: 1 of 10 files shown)


william-billaud:
Fyi there is two PR related to ESXi

#1509 Improve ESXiPlugin support for recent version -> TLDR try to properly mount log folder for different loader (vmsupport/raw disk) 
#1385  (Add support for ESXI auth, hostd and shell logs)

There is some overlap regarding `auth.log`, I think that the main difference is that in 1385 there is a common record for different ESXi log, in this PR this is a common record for auth format. IMO both are fine and it was more to highlight the point (E.g for IIS webserver logs there is two different record type for the same data).

Matthijsy:
Thank you for pointing it out! I did a check if there was a PR for the vCenter part, but forgot to do so for the ESXi plugins. I am more than happy to remove that part again if any of the other PRs are preferred.

PR #1880, opened 2026-07-08
Simplify GenericTarSubLoader

### LINKED ISSUE #1876: Tar files without direntries cannot be iterated
With the current state of `GenericTarSubLoader` the following test fails:

```python
def test_windows_tar_map(tmp_path: Path) -> None:
    """Test if we can parse a simple Windows tar tar image without explicit directory entries."""
    path = tmp_path / "test.tar"
    tf = tarfile.TarFile(path, mode="w")
    _mkfile(tf, "Users/John/Documents/hello.txt", b"Hello world!")
    _mkfile(tf, "Windows/System32/drivers/etc/hosts", b"127.0.0.1\tlocalhost\t#example")
    tf.close()
    loader = loader_open(path)
    target = Target()
    loader.map(target)

    assert list(map(str, target.fs.path("/").iterdir())) == ["Users", "Windows"]
    assert target.fs.path("Users").is_dir()
    assert target.fs.path("Windows").is_dir()
    assert target.fs.path("Users/John/Documents/hello.txt").read_text() == "Hello world!"
```
In #1777 you mentioned something about a tar rewrite @Schamper, is that something you have lying around somewhere? 😄

JSCU-CNI:
Moves all parsing logic in the GenericTarSubLoader to the TarFilesystem implementation. Fixes #1876.


dissect/target/loaders/tar.py
```diff
@@ -92,47 +87,11 @@ def detect(path: Path, tarfile: tf.TarFile) -> bool:
...
-
-        for vol_name, vol in volumes.items():
-            loaderutil.add_virtual_ntfs_filesystem(
-                target,
-                vol,
-                usnjrnl_path=[
-                    "$Extend/$Usnjrnl:$J",
-                    "$Extend/$Usnjrnl:J",  # Old versions of acquire used $Usnjrnl:J
-                ],
-            )
```
Schamper, line 133: Is this functionality not lost? Or is this just old legacy acquire stuff that we can probably safely remove now?

Maybe strange that I'm the one asking this question 😄 but I'm hoping you can refresh my memory on this.

Schamper (changes requested):

dissect/target/loaders/tar.py
JSCU-CNI: If I understand correctly `loaderutil.add_virtual_ntfs_filesystem` is a left over from when Acquire loader logic was still inside the tar loader.

PR #1883, opened 2026-07-20
Make target-shell commands glob paths

JSCU-CNI:
Makes sure `stat`, `file`, `cat`, `zcat` and all `hash` functions can handle glob path arguments just like the real commands do.


yunzheng:
Currently these commands only accepts only one `path` argument. It would be nice if they accept multiple arguments as well using `nargs=“+”`.

PR #1886, opened 2026-07-29
Compatibility with the new dissect.cstruct

Miauwkeru:
Requires https://github.com/fox-it/dissect.cstruct/pull/170 to function

### DIFF
--- a/dissect/target/plugins/os/windows/everything/parser.py
+++ b/dissect/target/plugins/os/windows/everything/parser.py
@@ -17,6 +17,8 @@ if TYPE_CHECKING:
     from collections.abc import Iterator
     from datetime import datetime
 
+    from dissect.cstruct import Endianess
+
 BZIP_HEADER = b"BZh9"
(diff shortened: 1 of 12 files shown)


JSCU-CNI:
Since cstruct `main` is now on version 5, it would be nice if target `main` got compatible with `main` of cstruct too.

dissect/target/plugins/os/windows/everything/parser.py
```diff
@@ -17,6 +17,8 @@
     from collections.abc import Iterator
     from datetime import datetime
 
+    from dissect.cstruct import Endianess
```
JSCU-CNI: I think it is spelled `Endianness`. What is the difference between `Endianess` and `AllowedEndianess`?

Schamper: Can you make a PR in cstruct to fix that? 🙈

PR #812, opened 2024-08-08
Refactor target-shell

JSCU-CNI:
This PR contains a `target-shell` refactor as discussed in https://github.com/fox-it/dissect.target/pull/664.

We've tried to break the refactor into separate commits:

- **4d7e0c7 Moves some functions outside of the `TargetCli` class so they can be used elsewhere.** 
  Such as in target-fs and hopefully target-diff at a later point in time. Functionally it changes two things: `target-fs ls` now prints with colors when outputting to a tty, and `target-shell ls` now supports human-readable file sizes. This closes https://github.com/fox-it/dissect.target/issues/624. 
- **7a45275 Introduces the `ExtendedCmd` class.**
  Moves some shell-interactivity functionality outside of the `TargetCli` class so it can be re-used (as we intend to do in https://github.com/fox-it/dissect.target/pull/664).
- **4cacdc7 Add trailing slash to directories when auto completing.**
  A small quality of life improvement where autocomplete in target-shell now adds a trailing slash to directories. This way, when pressing tab again, you'll get suggestions for that folder.
- **8153fee Add stat function to target-fs**
  Is a quick win, implementing the stat command from target-shell. There is potential to transfer even more functionality from target-shell to a shared utils file so it can be reused in target-fs (The `walk` command comes to mind).
- **2ed01b9 Fix opening tar and zip files inside target-shell using `enter`**
  Previously this worked but has been broken since the velociraptor plugin tries to load zip files using `os`.
- **b7f6765 UI/UX improvements to target-shell**
  We added `@alias` decorators for `cmd.Cmd` functions and generally improved most of the target-shell functions. We also added `PS1` support as can be seen in the screenshots below.
  
  Overview of all commands and aliases:
  image
  Default PS1 prompt colors:
  image
  Or if you want you could customize it in a `.targetcfg.py` file:
  image
  <sub>for those interested in the uwu config: `PS1="\x1b[1;35m{base}\x1b[1;36m{cwd}\x1b[0m (ﾉ^ヮ^)ﾉ✨ "`</sub>

PR #1891, opened 2026-07-30
Shows dissect.target version in --help.

### LINKED ISSUE #1665: Add version (--version, -V) information to target-*
Add a `--version` or `-v` switch to the target\* tooling to show the current running version.

Version information should also be added to the -h help output

william-billaud:
# Shows dissect.target version in --help.

* Shows the dissect.target version in each target-* --help.

## Proposed Changes

* Add a short description for each target-* in `--help` (from the readme)
* Add --version flag to `target-inspect`
* Use argparse built-in version [action](https://docs.python.org/3/library/argparse.html#action) instead of manually displaying version. I assume that when the feature was introduced in #812, author was unware of this feature, but I will revert him if this was here for a good reason (allowing to run something like `target-query -f <plugin> --version` that displays plugin version?).

Note : 
## Checklist

- [x] Closes #1665


dissect/target/tools/utils/cli.py
```diff
@@ -67,6 +67,13 @@ def __call__(self, parser: argparse.ArgumentParser, namespace: argparse.Namespac
         setattr(namespace, self.dest, self.const)
 
 
+def get_dissect_target_version() -> str:
+    try:
+        return version("dissect.target")
+    except PackageNotFoundError:
+        return "unable to determine version"
```
yunzheng:
```suggestion
        return "0+unknown"
```

`dissect.target version unable to determine version` looks weird. It seems `0+unknown` is a good fallback, see also: https://github.com/python-versioneer/python-versioneer/blob/master/details.md

dissect/target/tools/dd.py
```diff
@@ -26,7 +27,7 @@
 def main() -> int:
     help_formatter = argparse.ArgumentDefaultsHelpFormatter
     parser = argparse.ArgumentParser(
-        description="dissect.target",
+        description=f"target-dd {get_dissect_target_version()} : export (a part of) a target to a file or to stdout.",
```
yunzheng:
```suggestion
        description=f"target-dd {get_dissect_target_version()} : export (a part of) a raw target disk to a file or stdout, similar to the `dd` command on Unix systems.",
```

dissect/target/tools/inspect.py
```diff
@@ -186,10 +187,11 @@ def _add_cmds(parser: argparse.ArgumentParser, scope: str) -> None:
 def main() -> int:
     help_formatter = argparse.ArgumentDefaultsHelpFormatter
     parser = argparse.ArgumentParser(
-        description="dissect.target",
+        description=f"target-inspect {get_dissect_target_version()} : debugging utility.",
         fromfile_prefix_chars="@",
         formatter_class=help_formatter,
     )
+    parser.add_argument("--version", action="version", version=get_dissect_target_version())
```
yunzheng, line 194: Any reason why this was added to `inspect.py` and not to other commands? Also, this might not be necessary as it calls `configure_generic_arguments()` later which adds this automatically, correct?

yunzheng (changes requested):

dissect/target/tools/utils/cli.py
william-billaud: I do not really have an opinion of this topic, I just kept the previous value but let's go for "0+unknown"

dissect/target/tools/inspect.py
william-billaud: `configure_generic_arguments` is used on the base argument, which is used for subparser (`disk`, `volume`, `filesystem` sub-commands) but not on the "root" command help

### Without this line

```
❯ uv run --python 3.12 --refresh --extra dev --extra full target-inspect --help
usage: target-inspect [-h] {disk,volume,filesystem} ...

target-inspect 3.25.2.dev80+g383d6eded.d20260905 : debugging utility.

positional arguments:
  {disk,volume,filesystem}
                        scope to inspect
    disk                inspect disks
    volume              inspect volumes
    filesystem          inspect filesystems

options:
  -h, --help            show this help message and exit
```

### With this line

```
❯ uv run --python 3.12 --refresh --extra dev --extra full target-inspect --help
usage: target-inspect [-h] [--version] {disk,volume,filesystem} ...

target-inspect 3.25.2.dev80+g383d6eded.d20260905 : debugging utility.

positional arguments:
  {disk,volume,filesystem}
                        scope to inspect
    disk                inspect disks
    volume              inspect volumes
    filesystem          inspect filesystems

options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
```


### Version properly present in subparsers (via configure_generic_arguments)

```
❯ uv run --python 3.12 --refresh --extra dev --extra full target-inspect disk --help
usage: target-inspect disk [-h] [-K KEYCHAIN_FILE] [-Kv KEYCHAIN_VALUE] [-L LOADER] [--child CHILD] [--children] [--list-children] [--recursive] [-v] [--version] [-q] [--plugin-path PLUGIN_PATH [PLUGIN_PATH ...]]
                           TARGET {hexdump,info,list,shell} ...

positional arguments:
  TARGET                Target to load
  {hexdump,info,list,shell}
                        available commands
    hexdump             Hexdump some bytes.
    info                Print information.
    list                List available items.
    shell               Open (I)Python shell.

options:
  -h, --help            show this help message and exit
  -K KEYCHAIN_FILE, --keychain-file KEYCHAIN_FILE
                        keychain file in CSV format
  -Kv KEYCHAIN_VALUE, --keychain-value KEYCHAIN_VALUE
                        passphrase, recovery key or key file path value
  -L LOADER, --loader LOADER
                        select a specific loader (i.e. vmx, raw)
  --child CHILD         load child of target by path of index (see --list-children)
  --children            include children (depth=1 or use --recursive)
  --list-children       list children indices and paths (optionally use --recursive), then exit
  --recursive           make --(list-)children behave recursively
  -v, --verbose         increase output verbosity
  --version             show program's version number and exit
  -q, --quiet           do not output logging information
  --plugin-path PLUGIN_PATH [PLUGIN_PATH ...]
                        a file or directory containing plugins and extensions
```

yunzheng: I see, thanks for clarifying

yunzheng (changes requested):
I noticed there are two other target-* tools not touched by this PR, namely:

 * target-build-magic
 * target-build-pluginlist

Can you update them similar as well?

There is also a `target-build-winzones` but I think this is meant to be a internal tool.

william-billaud:
> * target-build-magic
> 
> * target-build-pluginlist

magic is also means to be used internally-> #1387, but it's listed as a script entry point so let do it, this does not cost a lot of energy.

Also added a quick description to `target-build-winzones`, so everything is clean and documented.

Just fyi, I will not use a computer until October. Feel free to apply modification if needed, or wait until I'm back, this is not an highly important issue, just a small improvement.

yunzheng (approved):

PR #1894, opened 2026-07-31
Fix Regression in the Dirloader loader following change on TargetPath

### LINKED ISSUE #1893: Regression is the Dirloader following change on TargetPath
#1390 seems to have introduced a regression on the DirLoader. which does not work anymore with windows Target (on a linux Workstation).

I have added an ugly `print(f"path : {path}")` in the DirecotryFilesystem.get function https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/dissect/target/filesystems/dir.py#L53

and I have the following result on main
```
uv run --python 3.13 --refresh --extra dev --extra full target-shell ~/Documents/tools/target/c/
c:/$ ls c:/windows/system32/config
path : /windows
path : windows\system32
ls: cannot access c:/windows/system32/config: No such file or directory
```

On 3.25.1 : 
```
uv run --python 3.13 --refresh --extra dev --extra full target-shell ~/Documents/tools/target/c/
<HOSTNAME>.local:/$ ls c:/windows/system32/config
path : /windows
path : windows/system32
path : windows/system32/config
# able to list files as expected.
```
1. Target hostname is not identified on main
2. The path requested does not have the same separator on each version.

From my understanding, it's probably related to https://github.com/fox-it/dissect.target/pull/1390/changes#diff-76da6ca6c4ec27ff5d714410247f4d7fe0befe4485f3047ca49a147e670a4468 and others change related to separator.

I will provides a P.R with some tests that succeeds on 3.25.1, but fail on main and try do dig deeper on this part of the code.

william-billaud:
# Fix Regression in the Dirloader loader following change on TargetPath

As specified on #1893, TargetPath introduce regression on the DirLoader (which no longer work). This P.R fix this regression and add tests.

## Proposed Changes

* Fix issue by normalizing path before lookup (as suggested by @twiggler ).
* Add tests DirLoader to test multiple opener and some scenario with more folder depth (issue was non caught as only one level of file/folder were created in tests).
* Add simple tests for the ZipLoader to ensure proper path resolution (no test were present) (basically the same as the one for the DirLoader).

## Checklist

- [ ] PR Title is descriptive of the changes
- [ ] The description is descriptive of the changes
- [ ] Tests pass

Close #1893


twiggler:
@william-billaud 

The internal convention seems to be to store paths as POSIX (/)

~~Perhaps change `DirectoryFilesystemEntry::get` to join with `/` instead of `self.fs.sep`?~~

Change `DirectoryFilesystem:_resolve_path` to normalize to POSIX before splitting?

```diff
diff --git a/dissect/target/filesystems/dir.py b/dissect/target/filesystems/dir.py
index 5cce43ad..f56edbc2 100644
--- a/dissect/target/filesystems/dir.py
+++ b/dissect/target/filesystems/dir.py
@@ -36,7 +36,7 @@ class DirectoryFilesystem(Filesystem):
         if not self.case_sensitive:
             searchpath = self.base_path
 
-            for p in path.split("/"):
+            for p in fsutil.normalize(path, sep=self.sep).strip("/").split("/"):
                 match = [d for d in searchpath.iterdir() if d.name.lower() == p.lower()]
 
                 if not match or len(match) > 1:
```

william-billaud:
Thanks, look like a good direction, I need to do more test to ensure proposed fix works

william-billaud:
@twiggler thanks for the help, I have added the change to the `get` function, and not `resolve` as get is the entry point for path access/resolution, and this prevent issues if get is called with `path == '\''`

Can I let you review/assign this issue?

twiggler (changes requested):
As the bug was in `DirectoryFileSystem`, I think we should add tests in `tests/filesystems/test_dir.py`, which tests that particular layer.

Applying the fix to `get`is the better solution.

tests/loaders/test_zip.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.mark.parametrize(
+    ("opener"),
+    [
+        pytest.param(Target.open, id="target-open"),
+        pytest.param(lambda x: next(Target.open_all([x])), id="target-open-all"),
+    ],
+)
+def test_unix_zip(opener: Callable[[str | Path], Target]) -> None:
+    """Test the ``ZipLoader`` for Linux directories."""
+    path = absolute_path("_data/loaders/zip/linux.zip")
```
twiggler, line 26: [def _create_zip(prefix: str = "", insert_dir: bool = True) -> io.BytesIO:](https://github.com/william-billaud/dissect.target/blob/84f0164840bc0b6220d63001f4c225dd01777da4/tests/filesystems/test_zip.py#L46)

writes in memory, perhaps cleaner than using fixtures?

Issue #1892, opened 2026-07-31
Normalize "source"/"path" record field, containing path of the file (or registry key) used to generate a specific record.


william-billaud:
Currently source/path record fields does not have the same meaning depending on the plugin.

Some plugins define a "source" fields, some other a "path" field and some other define a custom field name, such as : https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/dissect/target/plugins/os/windows/defender/mplog.py#L11 
Sometime path also refer to a field referenced in a log entry (e.g path to a malicious file in av logs) :

https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/dissect/target/plugins/apps/av/sophos.py#L131

Depending on plugin, "source' field can be typed as a string or a path, resolved or not, using unix of Windows style (some time no consistent inside the same plugin) :  

* https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/tests/plugins/os/windows/log/test_evtx.py#L143
* https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/tests/plugins/os/windows/test_search.py#L64
* https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/tests/plugins/apps/remoteaccess/test_rustdesk.py#L48
* https://github.com/fox-it/dissect.target/blob/bbd7fae8ce1830a252e33842dc03583da22c112e/tests/plugins/apps/remoteaccess/test_rustdesk.py#L29

Furthermore, some time, such a for the unix network plugin, source can refer to a path (such as `"/var/lib/NetworkManager/internal-d6b936ad-d73f-4898-a826-edbb61d6155a-eth0.lease"` or a string indicating that this was found in syslog files (but not the path of the syslog path)

It would be easier if the field was normalized, as it's valuable during analysis to know from which file / part of byte a record was generated (without looking inside plugin code):

My suggestion is the following

* `source` (or renamed to `source_path`): resolved path of the (main) file used to generated the record  : 
  * For windows systems in format "c:\xx\xx", for unix in format '/path/to/file.txt' 
  * For registry key : path to the registry hive.
* `source_registry_key`: abbreviated name of the registry key - if available - e.g. HKLM\\SOFTWARE\\Foo\\Bar.
[...]

Issue #1006, opened 2025-01-27
Refactor Windows Prefetch plugin


Horofic:
Issue #994 revealed that not all Windows Prefetch versions where included in the `prefetch` plugin. This was implemented in #1005, but revealed that the `prefetch` plugin is in need of a refactor.

PR #1895, opened 2026-07-31
Add source field to record of 9 plugins

william-billaud:
# Add source field to record of 7 plugin

Some plugins records does not define the source field, and does not allows to quickly identified from which file an record was generated. This PR add this missing field to 7 plugins

## Proposed Changes

Add "source" field to records of 9 plugins + associated tests: 
* apps/av/mcafee
* apps/av/sophos
* os/unix/log/atop
* os/windows/defender
* os/windows/log/agentexecutor
* os/windows/log/intunemanagementextension
* os/windows/sru
* os/windows/ad/ntds
* os/windows/prefetch

Add a test to ensure every plugins records define the `source` field. As there is already some backlogs (81 without path or source, not an issue for all), the test fail if the number of records type without 'source' or  `path` defined change.

Add a sample to test the prefetch plugin with a real file. Plugin seems to have an issue and does not properly identify run counts, but I let it for #1006.

## Checklist

- [ ] Related to #1892 , but does not fully complete the issue. If there is some value to continue the work, it will be done in new P.R.


dissect/target/plugins/os/windows/defender/mplog.py
```diff
@@ -8,7 +8,7 @@
     "windows/defender/mplog/processimage",
     [
         ("datetime", "ts"),
-        ("path", "source_log"),
+        ("path", "source"),
```
Schamper: Can you move the fields here to the bottom?

Schamper (changes requested):

dissect/target/plugins/os/windows/defender/mplog.py
william-billaud: Done

Schamper (approved):

PR #1896, opened 2026-08-01
Add docker paths for MacOS

Matthijsy:
This PR adds the the default docker datapath for MacOS

### DIFF
--- a/dissect/target/plugins/apps/container/docker.py
+++ b/dissect/target/plugins/apps/container/docker.py
@@ -8,7 +8,7 @@ from dissect.cstruct import cstruct
 from dissect.util import ts
 
 from dissect.target import container, filesystem, volume
-from dissect.target.exceptions import UnsupportedPluginError
+from dissect.target.exceptions import ContainerError, UnsupportedPluginError
 from dissect.target.helpers.fsutil import open_decompress
 from dissect.target.helpers.logging import get_logger
 from dissect.target.helpers.protobuf import ProtobufVarint
@@ -103,22 +103,6 @@ class DockerPlugin(ContainerPlugin):
         super().__init__(target)
         self.installs = set(find_installs(target))
 
-        for install in self.installs:
-            # macOS installs are within Docker.raw
-            if (docker_path := install.joinpath("Docker.raw")).exists():
-                vs = volume.open(container.open(docker_path))
-                if len(vs.volumes) > 1:
-                    self.target.log.warning(
-                        "Docker container %s has multiple volumes, only using first volume", install.name
-                    )
-
-                fs = filesystem.open(vs.volumes[0])
-
-                if not fs.get("/docker"):
-                    self.target.log.warning("Docker.raw misses docker directory")
-                else:
-                    self.target.fs.mount(str(install), fs.get("/docker"))
-
     def check_compatible(self) -> None:
         if not self.installs:
             raise UnsupportedPluginError("No Docker install(s) found")
@@ -127,7 +111,8 @@ class DockerPlugin(ContainerPlugin):
     def images(self) -> Iterator[DockerImageRecord]:
         """Returns any pulled docker images on the target system."""
         for data_root in self.installs:
-            images_path = data_root.joinpath("image/overlay2/repositories.json")
+            data_relative_root = open_docker_raw(data_root)
+            images_path = data_relative_root.joinpath("image/overlay2/repositories.json")
 
             if not images_path.exists():
                 self.target.log.debug("No docker images found, file %s does not exist", images_path)
@@ -139,6 +124,10 @@ class DockerPlugin(ContainerPlugin):
                 self.target.log.warning("Unable to parse JSON in: %s", images_path)
                 self.target.log.debug("", exc_info=e)
                 continue
+
+            full_images_path = images_path
+            if data_relative_root != data_root:
+                full_images_path = str(data_root) + "/" + str(images_path)
 
             for name, tags in repositories.items():
                 for tag, hash in tags.items():
@@ -161,7 +150,7 @@ class DockerPlugin(ContainerPlugin):
                         image_id=hash_to_image_id(hash),
                         created=created,
                         hash=hash,
-                        source=images_path,
+                        source=full_images_path,
                         _target=self.target,
                     )
 
@@ -169,7 +158,8 @@ class DockerPlugin(ContainerPlugin):
     def containers(self) -> Iterator[DockerContainerRecord]:
         """Returns any docker containers present on the target system."""
         for data_root in self.installs:
-            for config_path in data_root.joinpath("containers").glob("**/config.v2.json"):
+            data_relative_root = open_docker_raw(data_root)
+            for config_path in data_relative_root.joinpath("containers").glob("**/config.v2.json"):
                 try:
                     config = json.loads(config_path.read_text())
                 except (json.JSONDecodeError, UnicodeDecodeError) as e:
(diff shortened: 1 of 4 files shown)


JSCU-CNI:
Could you share the folder structure you observe locally inside `$HOME/Library/Containers/com.docker.docker/Data/vms/0/data`? As I understand Docker on macOS uses a virtual machine to run containers and the `data` directory would contain e.g. `Docker.raw` or `Docker.qcow2`. I think you would have to parse that virtual disk to actually read `/var/lib/docker/*`.

Matthijsy:
You're right, I missed that while implementing. I have a .raw on my system, not sure if it can be qcow on other systems. While implementing I found that moving that directory is not reflected in daemon.json but in `Library/Group Containers/group.com.docker/settings-store.json`.

The raw file looks like this:
```
cni
containerd
containerd-stargz-grpc
desktop-containerd
docker
dpkg
kubeadm
kubelet-plugins
lost+found
machine-id
mutagen
nfs
swap
wasm
```

The `docker` directory is comparable to `/var/lib/docker`. So after finding the installs it now mounts the docker directory such that we can use that in further processing. On my systems the repositories.json is missing, so potentially this is stored differently than on linux, but that can be investigated and fixed in a later PR.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -102,6 +103,22 @@ def __init__(self, target: Target):
...
+            if (docker_path := install.joinpath("Docker.raw")).exists():
+                vs = volume.open(container.open(docker_path))
+                if len(vs.volumes) > 1:
+                    self.target.log.warning(
+                        "Docker container %s has multiple volumes, only using first volume", install.name
+                    )
+
+                fs = filesystem.open(vs.volumes[0])
+
+                if not fs.get("/docker"):
```
Schamper: I don't think this method can return `None`? I think it raises.

In any case, preferably already use `.path()` API here. And add a unit test for the negative case.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -102,6 +103,22 @@ def __init__(self, target: Target):
...
+                    self.target.log.warning("Docker.raw misses docker directory")
+                else:
+                    self.target.fs.mount(str(install), fs.get("/docker"))
```
Schamper: I'm not a great fan of the modifications of the mounts (that would only happen when running this plugin). The source paths also won't be completely truthful.

I'm not sure if it'll actually be nicer, but a random idea I had:
- Don't mount it, but do keep a reference of the filesystem
- When yielding paths, also yield paths from the stored filesystem
- When setting source paths, somehow construct a path that resolves to `path/to/Docker.raw/path/to/container`

This works around both points, and makes it more explicit that the path comes from the `Docker.raw` file.

Let me know your thoughts!

Schamper (changes requested):

Matthijsy:
Thank you for the review @Schamper I've updated the logic to your proposed implementation. I have added tests based on two `Docker.raw` images which I generated by hand. If there is a better approach please let me know

dissect/target/plugins/apps/container/docker.py
```diff
@@ -110,7 +111,8 @@ def check_compatible(self) -> None:
     def images(self) -> Iterator[DockerImageRecord]:
         """Returns any pulled docker images on the target system."""
         for data_root in self.installs:
-            images_path = data_root.joinpath("image/overlay2/repositories.json")
+            data_relative_root = open_docker_raw(data_root)
+            images_path = data_relative_root.joinpath("image/overlay2/repositories.json")
```
JSCU-CNI: I am not a big fan of fixing the (relative) data root paths in each exported function. This causes duplication and does not feel like the right approach. I propose to change the `find_installs` method instead.

dissect/target/plugins/apps/container/docker.py
```diff
@@ -308,7 +324,25 @@ def get_data_path(path: Path) -> str | None:
         log.debug(exc_info=e)
         return None
 
-    return config.get("data-root")
+    return config.get("data-root") or config.get("DataFolder")
```
JSCU-CNI, line 312: Could you add a comment where `DataFolder` originates from (macOS-specific)?

dissect/target/plugins/apps/container/docker.py
```diff
@@ -357,7 +403,10 @@ def find_installs(target: Target) -> Iterator[Path]:
                     target.log.info("Unable to get data-root from docker daemon file %s", config_file)
                     continue
                 if (data_root_path := target.fs.path(data_path)).exists():
-                    yield data_root_path
+                    try:
+                        yield expand_datapath_macos(data_root_path)
+                    except Exception as e:
+                        print(e)
```
JSCU-CNI, line 392:
```suggestion
                        target.log.warning("<a meaningful warning message here>: %s", e)
                        target.log.debug("", exc_info=e)
```

dissect/target/plugins/apps/container/docker.py
```diff
@@ -325,6 +359,11 @@ def find_installs(target: Target) -> Iterator[Path]:
         "sysvol/ProgramData/docker",
     ]
 
+    user_data_paths = [
+        # MacOS
+        "Library/Containers/com.docker.docker/Data/vms/0/data",
```
JSCU-CNI: Technically more vms could exist if I understand the earlier referenced resources correctly. I'd opt to iterate the `vms` directory instead of assuming only `0` exists.

JSCU-CNI (changes requested):

dissect/target/plugins/apps/container/docker.py
Matthijsy: Would you agree than that `find_installs` can yield either a Path or a Filesystem? Than I can make it already yield the filesystem of the `Docker.raw` and just do the joinpath the same way as for a Path here.

JSCU-CNI: I don't think `find_installs` needs to return anything other than `Path`s, since we can yield a `Path` that originates from inside the `Docker.raw` container. Perhaps make `find_installs` return `tuple[Path, Path | None]` where the second `Path | None` indicates where the first `Path` originates from. That way you can reconstruct the full original path in the `source` of your records.

Matthijsy: But where would you open the container in that case? Would we still do that in every exported function? Since the full Path within the `Docker.raw` is not accessible on the target itself.

JSCU-CNI: Inside `find_installs`, something akin to:

```python
    ...
    for path in user_data_paths:
        for user_details in target.user_details.all_with_home():
            if (user_path := user_details.home_path.joinpath(path)).exists():
                if (docker_raw := user_path.joinpath("Docker.raw").is_file():
                    try:
                        vs = volume.open(container.open(docker_raw))
                        fs = filesystem.open(vs.volumes[0])
                        yield fs.path("/docker"), docker_raw
                    except ContainerError as e:
                        target.log.warning("...")
                        target.log.debug("", exc_info=e)
                else:
                    yield user_path, None
```

dissect/target/plugins/apps/container/docker.py
Matthijsy: I cannot find any official documentation that specifies how this file should look like. I can reference the Docker Desktop settings if that adds any value? The DataFolder key is not documented there unfortunately. 
https://docs.docker.com/desktop/settings-and-maintenance/settings/

PR #1898, opened 2026-08-04
Add Apple Disk Image (DMG) container

joost-j:
Counterpart of https://github.com/fox-it/dissect.hypervisor/pull/88. Enables users to use the dissect tools directly on (encrypted) `.dmg` files. Password can be provided with the `-Kv` parameter.


joost-j:
Validated the `dissect.target` and `dissect.hypervisor` implementation with an actual Fuji `.dmg` output file using `target-shell` and `target-query` and it seems to work fine.

PR #1905, opened 2026-08-27
Fix wrong replacement of subfolders containing fs/

joost-j:
# Fix wrong replacement of subfolders containing fs/

The recently merged changes regarding the clobber in https://github.com/fox-it/dissect.target/pull/1301 suddenly raised the new `KeyError` in some test acquire tar files of Linux distributions. It turns out that when subfolders actually contain the string `fs/`, they are then wrongly replaced because the loader assumes it's part of the `fs/` prefix. This applies to folders like `java.prefs/`, `java.zipfs` and common folders on linux like `/etc/brltty/Input/fs/pacmate.ktb` which contain `fs/`.

## Proposed Changes

This PR replaces the old logic with a oneliner that catches both the old and new Acquire tar format. `removeprefix` will return the original string if the prefix is not found. Test cases are added to verify this.

### DIFF
--- a/dissect/target/loaders/acquire.py
+++ b/dissect/target/loaders/acquire.py
@@ -82,7 +82,11 @@ class AcquireTarSubLoader(TarSubLoader):
 
             entry_cls = TarFilesystemDirectoryEntry if member.isdir() else TarFilesystemEntry
             entry = entry_cls(volume, fsutil.normpath(mname), member)
-            volume.map_file_entry(entry.path, entry)
+
+            try:
+                volume.map_file_entry(entry.path, entry)
+            except KeyError as e:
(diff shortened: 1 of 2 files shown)


joost-j:
I noticed that the new clobber implementation also raises issues when ran on legacy Acquire binaries, such as the one listed on the Dissect tools page (https://files.dissect.tools/images/MSEDGEWIN10_20220708124036.tar). This is because in those versions, e.g. the `NTUSER.DAT` file was collected both for the `c:` volume and `sysvol` volume, and when merging this, this obviously results in duplicate files and thus, a `KeyError`. Added a fix & test for that as well, copied the try/except from the `GenericTarSubLoader`.

Schamper (approved):

PR #1907, opened 2026-08-27
Add containerd support

### LINKED ISSUE #1897: Add docker support for containerd snapshotter
Currently the docker implementation only implements overlayfs2, new installations seem to use containerd snapshotter. Would be nice to have support for this. 

Docker info on a fresh trixie machine:
```
 Storage Driver: overlayfs
  driver-type: io.containerd.snapshotter.v1
```

If anyone has a few pointers if anything related to containerd snapshotter might already be in any of the dissect projects that would be great! If I find some time I might have a look into how it exactly works.

JSCU-CNI:
This PR:
* Fixes #1897
* Depends on https://github.com/fox-it/dissect.util/pull/133
* Depends on https://github.com/fox-it/dissect.database/pull/63
* Adds a containerd `OverlayFsFilesystem`
* Adds a containerd `OverlayFsLoader`
* Adds `overlayfs` driver support to the `DockerPlugin`
* Adds a new `ContainerdPlugin` supporting several containerd-based Kubernetes platforms
* Adds child detection for `containerd.containers`


JSCU-CNI:
I have added tests in [12cdc7e](https://github.com/fox-it/dissect.target/pull/1907/commits/12cdc7e9e97a83c17646df5883ccc173debd4656) and [a5986d2](https://github.com/fox-it/dissect.target/pull/1907/commits/a5986d2213ef95cd185a4be9474e07818c1bf8ad). This is now ready for review.

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+                    self.target.log.warning("Unable to parse containerd meta.db file %s: %s", meta_path, e)
+                    continue
+
+                try:
+                    ns = infer_bbolt_namespace(meta_db)
+                except ValueError as e:
+                    self.target.log.warning("Unable to determine containerd meta.db namespace %s: %s", meta_path, e)
+                    continue
+
+                yield dir, meta_db, ns
```
Miauwkeru, line 115:
```suggestion
            dir = self.target.fs.path(path)
            if not dir.is_dir():
                continue
            
            if not (meta_path := dir.joinpath("io.containerd.metadata.v1.bolt/meta.db")).is_file():
                self.target.log.warning("Containerd install %s does not contain meta.db", dir)
                continue
            try:
                 meta_db = Bbolt(meta_path)
            except Exception as e:
                self.target.log.warning("Unable to parse containerd meta.db file %s: %s", meta_path, e)
                continue

            try:
                ns = infer_bbolt_namespace(meta_db)
            except ValueError as e:
                self.target.log.warning("Unable to determine containerd meta.db namespace %s: %s", meta_path, e)
                continue

            yield dir, meta_db, ns
```
Maybe a preference thing, but i don't like too much indentation

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+            for container_id in containers:
+                spec = {}
+                meta = {}
+
+                if raw_spec := meta_db.get(f"v1 {ns} containers {container_id} spec", decode=False):
+                    try:
+                        path, spec = unmarshal_any_json(raw_spec)
+                    except ValueError as e:
+                        self.target.log.warning(
+                            "Failed to decode typeurl structure %s for container %s: %s", path, container_id, e
```
Miauwkeru, line 169: if you get a `ValueError,` `path` doesn't get set, this either crashes during the first iteration, or it uses a previous value of `path` during the loop

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+    Returns: inferred namespace name.
+    """
+    if not (root_keys := db.keys("v1")):
+        raise ValueError("Unsupported bbolt database version")
+
+    if "moby" in root_keys:
+        return "moby"
+
+    if "k8s.io" in root_keys:
+        return "k8s.io"
```
Miauwkeru, line 256: Can there be a situation it contains both? a kubernetes host and one that uses moby?

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+                        )
+
+                if raw_meta := meta_db.get(
+                    f"v1 {ns} containers {container_id} extensions io.cri-containerd.sandbox.metadata", decode=False
+                ):
+                    try:
+                        path, meta = unmarshal_any_json(raw_meta)
+                    except ValueError as e:
+                        self.target.log.warning(
+                            "Failed to decode typeurl structure %s for container %s: %s", path, container_id, e
```
Miauwkeru, line 179: same thing with `path` here

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+                        )
+
+                volumes = [
+                    f"{mount_point.get('source')}:{mount_point.get('destination')}"
+                    for mount_point in spec.get("mounts", [])
+                    if mount_point.get("type") == "bind"
+                ]
+
+                created_at = meta_db.get(f"v1 {ns} containers {container_id} createdat", decode=False)
```
Miauwkeru, line 188: is this always populated with value? otherwise converting the timestamp will create an error.

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,258 @@
...
+                    elif type == "F" and entry["message"] == "":
+                        entry["message"] = buf
+                        buf = ""
+
+                    yield ContainerdLogRecord(
+                        container=log_path.stem,
+                        **entry,
+                        source=log_path,
+                        _target=self.target,
+                    )
```
Miauwkeru, line 238: As this is basically the same for `containerd` and `podman`, is it an idea to make it generic and make both of the functions use this?

tests/helpers/test_typeurl.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from dissect.target.helpers.typeurl import unmarshal_any_json
+from tests._utils import absolute_path
+
+
+def test_typeurl_unmarshal_json() -> None:
```
Miauwkeru, line 7: this is the happy flow, can you also add a test for the unhappy flow?

Miauwkeru (changes requested):

dissect/target/plugins/apps/container/containerd.py
JSCU-CNI: This seems to always be populated (https://github.com/containerd/containerd/blob/main/core/metadata/boltutil/helpers.go) but I can add a sanity check.

dissect/target/plugins/apps/container/containerd.py
JSCU-CNI: Seems unlikely but I think it is technically possible. Not sure how that would work in practice as I think the bbolt files are not usable in such a way (e.g. read-write-many, file locks, etc). I think for now it is safe to assume only one container implementation is used per target. The refactoring required to make this work for multiple namespaces does in my opinion not outweigh the likeliness of a target having multiple ctr implementations.

dissect/target/plugins/apps/container/containerd.py
```diff
@@ -0,0 +1,277 @@
...
+                created_at: bytes = meta_db.get(f"v1 {ns} containers {container_id} createdat", decode=False)  # type: ignore
+
+                # NOTE: The following fields cannot be populated with the information we have here:
+                # image_id, command, running, pid, started, finished, config_path, image_path
+                yield ContainerdContainerRecord(
+                    container_id=container_id,
+                    image=meta_db.get(f"v1 {ns} containers {container_id} image"),
+                    created=golangtimestamp(created_at) if created_at else None,
+                    ports=[
+                        f"0.0.0.0:{port.get('container_port')}->{port.get('container_port')}/tcp"
```
Miauwkeru: if it is a mapping from container to host, i don't think they are both `container_port`

dissect/target/plugins/apps/container/containerd.py
Miauwkeru: thanks!

Miauwkeru (commented):
Doesn't the pyproject.toml also need to be updated for dissect.database?

pyproject.toml
Miauwkeru:
```suggestion
    "dissect.util>=3.25.dev9,<4",  # TODO: Update on release!
```

Miauwkeru:
```suggestion
    "dissect.util>=3.25.dev9,<4.0.dev",  # TODO: Update on release!
```

Miauwkeru (changes requested):
I forgot about the dissect.util requirements

JSCU-CNI:
> I forgot about the dissect.util requirements

Same

Miauwkeru:
I don't know why, but it seems the exception is different on pypy, can you fix those tests?

PR #1909, opened 2026-08-27
Add --absolute-names to tar command in target-shell

JSCU-CNI:
This PR adds the ability to save absolute member names (`--absolute-names` / `-P`) to the `target-shell` `tar` command.

Without `-P`:
```
localhost:/var$ tar -cvf log.tar ./log
a log
a log/alternatives.log
a log/apt
a log/apt/eipp.log.xz
a log/apt/history.log
a log/bootstrap.log
a log/btmp
a log/dpkg.log
a log/lastlog
a log/wtmp
```

With `-P`:
```
localhost:/var$ tar -cvPf log.tar ./log
a var/log
a var/log/alternatives.log
a var/log/apt
a var/log/apt/eipp.log.xz
a var/log/apt/history.log
a var/log/bootstrap.log
a var/log/btmp
a var/log/dpkg.log
a var/log/lastlog
a var/log/wtmp
```

### DIFF
--- a/dissect/target/tools/shell.py
+++ b/dissect/target/tools/shell.py
@@ -1139,7 +1139,7 @@ class TargetCli(TargetCmd):
 
         return False
 
-    @arg("path", nargs="+", type=TargetPathArgument)
+    @arg("path", nargs="+", metavar="PATH", type=TargetPathArgument)
     @arg(
         "-f",
         "--file",
@@ -1162,6 +1162,12 @@ class TargetCli(TargetCmd):
         "--dereference",
         action="store_true",
         help="follow symlinks and archive the files they point to instead of the symlinks themselves",
+    )
+    @arg(
+        "-P",
+        "--absolute-names",
+        action="store_true",
+        help="store the absolute path of a file",
     )
     @arg("-v", "--verbose", action="store_true")
     def cmd_tar(self, args: argparse.Namespace, stdout: TextIO) -> bool:
@@ -1247,8 +1253,11 @@ class TargetCli(TargetCmd):
                 print(f"tar: {PurePosixPath(path)}: unsupported file type, skipping", file=sys.stderr)
                 return
             if info.isreg():
-                with path.open("rb") as f:
-                    tar.addfile(info, fileobj=f)
+                with path.open("rb") as fh:
+                    try:
+                        tar.addfile(info, fileobj=fh)
+                    except Exception as e:
+                        print(f"tar: {PurePosixPath(path)}: failed to read file: {e}", file=sys.stderr)
             elif info.isdir():
                 tar.addfile(info)
                 for child in path.iterdir():
@@ -1265,11 +1274,11 @@ class TargetCli(TargetCmd):
         fobj = stdout.buffer if args.file == "-" else Path(args.file).open("wb")  # noqa: SIM115
         mode = "w|"
         if args.gzip:
-            mode = "w|gz"
+            mode += "gz"
         elif args.bzip2:
-            mode = "w|bz2"
+            mode += "bz2"
         elif args.xz:
-            mode = "w|xz"
+            mode += "xz"
 
         try:
             inodes_cache = {}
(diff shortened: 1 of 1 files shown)


Miauwkeru:
Can you add something about what this attempts to solve or add?

JSCU-CNI:
@Miauwkeru See the updated description.

Miauwkeru:
@JSCU-CNI thank you, i will take a look :)

dissect/target/tools/shell.py
```diff
@@ -1286,11 +1295,18 @@ def add_to_tar(
...
                     for path in paths:
-                        arcname = str(PurePosixPath(path).relative_to("/"))
-                        if not Path(arg_path).is_absolute():
-                            pure_path = PurePosixPath(path).relative_to(PurePosixPath(self.cwd))
-                            arcname = "/".join(p for p in pure_path.parts if p not in (".", ".."))
-                            arcname = arcname or "."
+                        if args.absolute_names:
+                            arcname = str(PurePosixPath(path))
+                            if arcname.startswith("/"):
+                                arcname = arcname.lstrip("/")
```
Miauwkeru, line 1301:
```suggestion
                            # use the absolute version of path
                            arcname = str(PurePosixPath(path)).lstrip("/")
```

As `lstrip` doesn't change the output if there is no leading `/` i think it's fine to just always use it and get rid of the if condition

dissect/target/tools/shell.py
```diff
@@ -1286,11 +1295,18 @@ def add_to_tar(
...
+                        else:
+                            if Path(arg_path).is_absolute():
+                                arcname = str(PurePosixPath(path).relative_to(path.drive or "/"))
+                            else:
+                                pure_path = PurePosixPath(path).relative_to(PurePosixPath(self.cwd))
+                                arcname = "/".join(p for p in pure_path.parts if p not in (".", ".."))
+                                arcname = arcname or "."
+
```
Miauwkeru, line 1309:
```suggestion
                        elif Path(arg_path).is_absolute():
                            arcname = str(PurePosixPath(path).relative_to(path.drive or "/"))
                        else:
                            pure_path = PurePosixPath(path).relative_to(PurePosixPath(self.cwd))
                            arcname = "/".join(p for p in pure_path.parts if p not in (".", ".."))
                            arcname = arcname or "."
```
You can reduce the indentation of the if statements so it looks a bit better

Miauwkeru (changes requested):
Can you add a test for this functionality?

yunzheng:
I've been looking at this PR and the real `tar` and I want to start a short design question:

What do we want the `target-shell tar` options to be, a faithful mirror of real `tar`, or our own thing that borrows the familiar names? 

I ask because `-P` here doesn't do what `tar -P` does. In GNU tar, `-P` only disables stripping of leading `/` (and `../`) from member names. It never resolves relative paths, and with an absolute argument it keeps the slash. So `tar -cvPf log.tar ./log` stores `./log/...`, not `var/log/...`, and `tar -cvPf log.tar /var/log` stores `/var/log/....` This PR does the opposite for both: it resolves to absolute and strips the slash.

So far I've kept the dissect shell's `tar` options as close to real `tar` as possible, mostly for muscle memory and interoperability.

Two things that might change the picture:

1. If the goal is "full path from root, no leading slash", that already works by passing an absolute path (from the top of my head):

```
localhost:/var$ tar -cvf log.tar /var/log
a var/log
a var/log/alternatives.log
a var/log/apt
a var/log/apt/eipp.log.xz
a var/log/apt/history.log
a var/log/bootstrap.log
a var/log/btmp
a var/log/dpkg.log
a var/log/lastlog
a var/log/wtmp
```

2. If the goal is to make that work for relative paths too (resolve against cwd), which is a reasonable feature to have. I suggest to introduce a custom long option, e.g. `--resolve-paths` or `--realpath`. Short flag would be difficult as most of them are already taken by `tar`.

Note that I'm also open to deviating from real `tar` as we will never implement all of its options and short flags in the dissect shell anyway. I'm just raising this as a design decision.

PR #1910, opened 2026-09-02
Fix deprecation warning for argparse.FileType in fortifw

Miauwkeru:
# Fix argparse.Filetype deprecation

argparse.Filetype is deprecated since python 3.14 and prefers just
opening the file yourself

https://docs.python.org/3/library/argparse.html#filetype-objects


yunzheng (approved):
LGTM

PR #1912, opened 2026-09-02
Add target-extract command

### LINKED ISSUE #1911: Add command to extract files from targets
Multiple plugins currently support `get_paths()`. It would be great to have a commandlet which allows us to extract the raw files from a target based on a plugin. Currently we can already do `target-fs cp`, but that requires us to know the exact paths.

Matthijsy:
# Add target-extract command

This PR adds a new commandlet called `target-extract`, which allows us to extract raw evidence out of a target. 

For example `target-extract -t <target> -f iis -o tmp` will output all IIS logs into the tmp directory.

## Checklist

- [x] Closes #1911


Schamper:
Maybe this could be an addition to `target-fs` instead of a whole new command?

Matthijsy:
> Maybe this could be an addition to `target-fs` instead of a whole new command?

Good point. I moved it to `target-fs`. Only thing is that `target-fs` always expects an path to operate on. For `target-fs extract` this is not needed. For that reason I added a default, and before doing the glob we already call the handle.

yunzheng:
@Schamper This reminds me of our discussion about adding `-e` / `--execute` to `target-shell`, so you could run one or more commands on a Target non-interactively (e.g. `cp`, `tar`, or `extract` here) and exit. Handy for scripting. Longer term that could let us deprecate `target-fs` and cut some duplicated code.

Not a blocker for this PR just curious if you think it’s worth a follow-up issue.

PR #1934, opened 2026-09-16
Add carving of FortiGate firmware decryption keys

### LINKED ISSUE #1935: Add support for decrypting latest FortiGate firmware (v7.6.5+ and v8.0.0+)
The "known" FortiGate keys in `dissect.target` have not been updated for a while now and does not work with newer FortiGate images unless you know the key. Also since v7.6.5 and v8.0.0 it uses a modified RC4 which is not supported yet.

It would be nice to add transparent support for retrieving the decryption keys instead of working with a known keys map.

yunzheng:
This PR adds automatic carving of FortiGate firmware decryption keys so `rootfs.gz` can be decrypted when no key is known or supplied.

The `fortios` plugin now includes a `fwkey` helper that structurally scans the kernel (`flatkc`) for the raw key material and RSA blob, then uses those to decrypt `rootfs.gz`.

This covers the layouts used through FortiGate v8.0.0:

 - Carve static ChaCha20 key+iv + RSA blob -> ChaCha20 decrypt rootfs.gz
 - Carve ChaCha20Seed + RSA blob -> KDF -> ChaCha20 decrypt rootfs.gz
 - Carve ChaCha20Seed + RSA blob -> KDF -> ChaCha20 decrypt RSA blob -> RSA decrypt footer -> AES key+iv -> AES-CTR decrypt rootfs.gz
 - Carve XOR key + RSA blob -> unmask RSA blob -> RSA decrypt footer -> RC4 key -> (modified) RC4 decrypt rootfs.gz

`fwkey` lives under the `fortios` plugin but is self-contained and can be copied out for standalone use.

Carving is done by a structural scan of the kernel (`flatkc`) that locates the raw key and RSA blob by layout/heuristics rather than by emulating any code. This differs from the public writeups linked below. This makes it fast and architecture-independent (no per-CPU emulator needed). Because it's layout-based, a future firmware layout may not match and would need an added pattern. In that case `fwkey` carves nothing and `rootfs.gz` is left encrypted, so existing behavior is unaffected.

Spot tested against: 8.0.1, 8.0.0, 7.6.5, 7.6.6, 7.6.7, 7.4.12, 7.4.1

Related public writeups:
- https://blog.randorisec.fr/fortigate-rootfs-decryption/
- https://github.com/hacefresko/forticrack_v8
- https://n0p.me/2026/08/2026-08-21-fortitool-fortios-decryption/
- https://web.archive.org/web/20250617060330/https://www.noways.io/blogs/tech/fortigate-firmware-analysis
- https://bishopfox.com/blog/further-adventures-in-fortinet-decryption

closes #1935

## Proposed Changes

* Add a `fwkey` helper under the `fortios` plugin to carve FortiGate firmware decryption material (raw key + RSA blob) by structurally scanning the kernel (`flatkc`).
* Use the carved material to decrypt `rootfs.gz` without a pre-supplied key, including the v8.0.0 layout (XOR-masked RSA blob and modified RC4). If the carving fails it will not be able to unpack `rootfs.gz` and leaves it encrypted.
* Keep `fwkey` self-contained inside the plugin so it can be copied out and used standalone. No separate CLI entrypoint is included in this PR.
* Refactor `fortios._os` to call `fwkey` when no known key or keychain value is specified.

## Checklist

- [X] PR Title is descriptive of the changes
- [X] The description is descriptive of the changes
- [x] Closes `#1935`


dissect/target/plugins/os/unix/linux/fortios/fwkey/ciphers.py
```diff
@@ -0,0 +1,175 @@
...
+        Custom CTR increment.
+    """
+    increment = 0
+    for i in range(16):
+        increment ^= (iv[i] & 15) ^ ((iv[i] >> 4) & 0xFF)
+    return max(increment, 1)
+
+
+def aes_decrypt(buffer: bytes, key: bytes, iv: bytes) -> bytes:
+    """Decrypt buffer using a custom AES CTR increment with given AesKey.
```
Schamper:
```suggestion
    """Decrypt buffer using a custom AES CTR increment with given key.
```

dissect/target/plugins/os/unix/linux/fortios/fwkey/ciphers.py
```diff
@@ -0,0 +1,175 @@
...
+        data: data to encrypt/decrypt.
+        key: RC4 key.
+        i_bits: number of bits for cross-mixed indices.
+        reset_j: whether to reset j.
+        mix_const: mix constant.
+
+    Returns:
+        Encrypted/decrypted bytes.
+    """
+    assert 1 <= i_bits <= 7
```
Schamper, line 114: Make this a raise?

dissect/target/plugins/os/unix/linux/fortios/fwkey/validate.py
```diff
@@ -0,0 +1,163 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from .ciphers import aes_decrypt, chacha20_crypt, rc4_crypt, rsa_public_decrypt, xor
+from .kdf import kdf_7_4_x
+from .keycarver import iter_raw_key_candidates
+from .models import AesKey, ChaCha20Key, ChaCha20Seed, RC4Key
+from .tools.extract_vmlinux import unpack_kernel
```
Schamper, line 10: Make absolute imports?

Schamper (changes requested):
I didn't follow through the code thoroughly enough to see if you log somewhere (to a target-bound logger) if a valid key was found?

dissect/target/plugins/os/unix/linux/fortios/fwkey/validate.py
yunzheng: I made everything in `fwkey` submodule standalone and nothing is tied to dissect code. This way you could just copy out the `fwkey` module and use the APIs on your own firmware files.

Would there be a problem by keeping them relative?

dissect/target/plugins/os/unix/linux/fortios/fwkey/ciphers.py
yunzheng: It would already raise with `AssertionError`

yunzheng:
> I didn't follow through the code thoroughly enough to see if you log somewhere (to a target-bound logger) if a valid key was found?


It does so in `_os.py:162`:

```python
                # Try to decrypt with all available keys
                rootfs_data = rootfs.read_bytes()
                for key in keys:
                    target.log.warning("decrypt_rootfs() check with %r", key)
                    try:
                        decrypt_rootfs(rootfs_data[:4], key)  # small decrypt for quick fail check
                        target.log.info("  SUCCESS with key %r", key)
                        target.log.warning("Found the correct key. Decrypting full rootfs.gz; this may take a while...")
                        start_time = time.time()
                        rootfs_data = decrypt_rootfs(rootfs_data, key)
                        target.log.info("Decryption took %s seconds", time.time() - start_time)
                        break
                    except ValueError:
                        target.log.info("  FAILED with key %r", key)
                        continue
                else:
                    raise ValueError("All decryption attempts failed :(")  # noqa: TRY301
```

the "SUCCESS" and "FAILED" messages are logged as `INFO`, so you would not see them if you don't specify `-v` (verbose).

However, I made `Found the correct key. Decrypting full rootfs.gz; this may take a while...` message log as `WARNING`, so this always shows.

dissect/target/plugins/os/unix/linux/fortios/fwkey/ciphers.py
Schamper: I know, but we don't do asserts anywhere else in the Dissect codebase, and you do a similar check just a few lines down that does have an explicit raise.

dissect/target/plugins/os/unix/linux/fortios/fwkey/validate.py
Schamper: Ah okay, maybe add a comment in the code near the imports explicitly stating that.

It's more a style thing, and a comment would protect it from an overzealous contributor and/or a future linting rule.

yunzheng: Added the comment in my latest commit

dissect/target/plugins/os/unix/linux/fortios/fwkey/ciphers.py
yunzheng: aight, changed the assert to a check that raises ValueError

Schamper (approved):

